PR #38, opened 2018-08-13
Update the Antora project configuration

settermjd:
This makes a minor changes the directory, inside the deployment directory, where the project files are deployed. Previously, it was set to "owncloud-docs", but for some reason it's been changed to "server". This breaks all of the static file links within the documentation. To avoid a lot of work correcting the links, the directory name's being changed in antora.yml instead.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -1,4 +1,4 @@
-name: server
+name: owncloud-docs
 title: ownCloud Documentation
 version: master
 


tboerger:
I have changed it to server because it reflects the current docs and because it's the server documentation. So it had been a change from my side on purpose.

settermjd:
Bummer. Ok, going to do some find and replace in the docs.

PR #61, opened 2018-09-03
Indexless

voroyam:


voroyam:
The appliance is different than the standard instance. These are the steps to enable indexless URLs in the appliance.

tboerger:
So you suggest changes to an ephemeral container to loose changes if the container gets recreated for some reason... All this should be already part of the docker image...

PR #62, opened 2018-09-03
 Info about upgrade from Community Version to Enterprise Version

voroyam:
Info about upgrade from Community Version to Enterprise Version

### DIFF
--- a/modules/ROOT/pages/faq/index.adoc
+++ b/modules/ROOT/pages/faq/index.adoc
@@ -1,6 +1,14 @@
 FAQ
 ===
 
+[[I want to upgrade from Community Version to Enterprise Version. What are the changes?]]
+== I want to upgrade from Community Version to Enterprise Version. What are the changes?
+
+In ownCloud Enterprise you will get access to new apps and features, mainly targeted towards enterprises.  Apps and features that ensure security for example.
+
+Your existing *files*, *shares* and *users* remain as they are. You can upgrade to the Enterprise version without concerns.
+
+
 [[how-do-i-transfer-files-from-one-user-to-another]]
 == How do I transfer files from one user to another?
 
--- a/modules/ROOT/pages/faq/index.adoc
+++ b/modules/ROOT/pages/faq/index.adoc
@@ -1,12 +1,12 @@
 FAQ
 ===
 
-[[I want to upgrade from Community Version to Enterprise Version. What are the changes?]]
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/faq/index.adoc
```diff
@@ -1,6 +1,14 @@
 FAQ
 ===
 
+[[I want to upgrade from Community Version to Enterprise Version. What are the changes?]]
+== I want to upgrade from Community Version to Enterprise Version. What are the changes?
+
+In ownCloud Enterprise you will get access to new apps and features, mainly targeted towards enterprises.  Apps and features that ensure security for example.
```
mmattel: Double blank between:
``enterprises.  Apps``

modules/ROOT/pages/faq/index.adoc
```diff
@@ -1,6 +1,14 @@
...
+
+Your existing *files*, *shares* and *users* remain as they are. You can upgrade to the Enterprise version without concerns.
```
settermjd: I feel that it reads better when the two sentences are swapped and slightly reworked, as follows:

You can upgrade to the Enterprise version without concern, as your existing *files*, *shares* and *users* remain as they are.

modules/ROOT/pages/faq/index.adoc
```diff
@@ -1,6 +1,14 @@
 FAQ
 ===
 
+[[I want to upgrade from Community Version to Enterprise Version. What are the changes?]]
```
settermjd: The anchor's quite long, but if you happy with it, then to follow the convention, you need to add hyphens between each word, as in the example below.

modules/ROOT/pages/faq/index.adoc
voroyam: Don't you think that the sentence as it is puts the emphasis on the files shares and users, as they are in the first part of the sentence. In you rewrite it reads better, but the files shares and users are in the last part of the sentence, making the user read.

modules/ROOT/pages/faq/index.adoc
voroyam: good eye

modules/ROOT/pages/faq/index.adoc
voroyam: Do I need this anchor? I just copy pasted from the things that were already there

PR #70, opened 2018-09-07
limitations of encryption

voroyam:
known limitations of user key encryption
https://github.com/owncloud/documentation/issues/4179


modules/ROOT/pages/configuration/files/encryption_configuration.adoc
```diff
@@ -229,6 +230,16 @@ You may only disable encryption by using the `occ Encryption Commands`_. Make su
...
+= Here are the limitations of this type of encryption:
+
+* User added to group cannot decrypt files on existing shares
+* OnlyOffice will not work
+* Impersonate will not work
+* oauth2 does will not work
+* elastic search will not work
+* user getting access to external storage already containing existing encrypted files cannot get access to said files for similar reasons like the group case above
+* When having data shared with a group and group membership changes after the share is established, subsequently added users will not be able to open the shared data unless the owner will share it again
+
```
mmattel: I would move this paragraph as last part of ``Before Enabling Encryption`` and reword the headline:
``Here are the limitations of this type of encryption`` -->
``Limitations of User-Key Based Encryption``

Please also see the header size which imho does not look good, limitations should be a bit smaller than the header before
image

voroyam: I adjusted the header size.

On the first point though, I must disagree.

If we put this info at the top, far away from the commands that enable the user key encryption I fear that users will just skip that section, and scroll down to the commands.

When we, on the other hand, put the warning directly above the commands, the chance of the users to read this is higher imho.

mmattel: My point is, that there is no mentioning of limitations in ``Before Enabling Encryption`` which is the must read part before you go for what ever type of encryption.
But we can bring our points together :-)
Move it up and refrence to it from below like like ``Before you enable, please see the limitations described above``.

@settermjd can we ask you for a review of our thoughts ?

modules/ROOT/pages/configuration/files/encryption_configuration.adoc
```diff
@@ -229,6 +230,16 @@ You may only disable encryption by using the `occ Encryption Commands`_. Make su
 
 == Enabling User-Key Based Encryption From the Command-line
 
+=== Here are the limitations of this type of encryption:
+
+* User added to group cannot decrypt files on existing shares
+* OnlyOffice will not work
+* Impersonate will not work
+* oauth2 does will not work
```
settermjd: s/oauth2/OAuth2/g

modules/ROOT/pages/configuration/files/encryption_configuration.adoc
```diff
@@ -229,6 +230,16 @@ You may only disable encryption by using the `occ Encryption Commands`_. Make su
 
 == Enabling User-Key Based Encryption From the Command-line
 
+=== Here are the limitations of this type of encryption:
+
+* User added to group cannot decrypt files on existing shares
```
settermjd: s/User added to group/Users added to groups/g

modules/ROOT/pages/configuration/files/encryption_configuration.adoc
```diff
@@ -229,6 +230,16 @@ You may only disable encryption by using the `occ Encryption Commands`_. Make su
...
+* OnlyOffice will not work
+* Impersonate will not work
+* oauth2 does will not work
+* elastic search will not work
```
settermjd: s/elastic search/Elasticsearch/g

modules/ROOT/pages/configuration/files/encryption_configuration.adoc
```diff
@@ -229,6 +230,16 @@ You may only disable encryption by using the `occ Encryption Commands`_. Make su
...
+* user getting access to external storage already containing existing encrypted files cannot get access to said files for similar reasons like the group case above
```
settermjd: s/user/User/g

modules/ROOT/pages/configuration/files/encryption_configuration.adoc
settermjd: After having a look at the existing documentation and its structure, I'd keep the order as it is, as it seems to be correct because warnings for each type of encryption are provided before (yet alongside) the commands to enable it. However, I'd make several changes:

- Add a new, third-level, header: "Master Key Encryption", after the existing header "Encryption Types".
- Reflow all the existing headers down to, and including "Disabling Encryption" to be in sync with that header level.
- Add a new, third-level, header: "User-Key Encryption", before "Enabling User-Key Based Encryption From the Command-line".
- Reword "Here are the limitations of this type of encryption:" as "Encryption Limitations".
- Make "To enable User-Key based encryption:" a fourth-level header.
- Reflow all the existing headers down to, and including "Disabling Encryption" to be in sync with that header level.

voroyam: > * Add a new, third-level, header: "_Master Key Encryption_", after the existing header "_Encryption Types_".

right after "Encryption Types" comes:
"ownCloud provides two encryption types"

Putting a "Master Key Encryption Header" there makes no sense to me. Could you specify?

settermjd: The intent is to make it clear where Master Key encryption starts, as there are two encryption types.

voroyam:
@settermjd do you need something from me to finish this?

settermjd:
@voroyam, doesn't look like it.

PR #73, opened 2018-09-07
Minimum client versions

voroyam:
A user in central has brought to my attention that nowhere in our docs we have mentioned what client version still works with the current server release.

If people are for what ever reason not upgrading to the latest version we should document what still works.

### DIFF
--- a/modules/administration_manual/pages/installation/system_requirements.adoc
+++ b/modules/administration_manual/pages/installation/system_requirements.adoc
@@ -1,5 +1,3 @@
-:icons: font
-
 System Requirements
 ===================
 
@@ -92,6 +90,15 @@ If you use Ubuntu 16.04:
 * openSUSE Leap 42.2 & 42.3
 
 NOTE: For Linux distributions, we support, if technically feasible, the latest 2 versions per platform and the previous https://wiki.ubuntu.com/LTS[LTS].
+
+Client Versions
+~~~~~~~~~~~~~~~
+
+Here are the oldest versions of the Desktop Client, Android app and iOS app supported with the latest server release:
+
+* Desktop Client 2.3.3
+* Android App
+* iOS App
 
 [[alternative-but-unsupported-options]]
 == Alternative (But Unsupported) Options


voroyam:
This is a first draft.

I need info about the iOS and Android client versions.

Don't merge this yet.

cdamken:
> I need info about the iOS and Android client versions.

@michaelstingl can you please help here?

cdamken:
Normally we recommend to use only the latest version. but @michaelstingl can provide more info

settermjd:
I agree with @cdamken and had heard that in the last 24 - 48 hrs. @michaelstingl, can you confirm, pls?

michaelstingl:
Maybe explain:

Only the latest stable version of desktop and mobile apps are supported. We don't test older versions of desktop and mobile apps with every new server release. For desktop, I'm aware of problems with chunked uploads with versions prior to 2.3.3.

voroyam:
@michaelstingl Okay, but that does not really tell users the minimal version for mobile clients.

If you say that we / you can't make such a statement, because you don't test this case / cases, then we / I can just write what you wrote: We don't test / support older versions of the mobile clients / desktop clients.

michaelstingl:
> I can just write what you wrote: We don't test / support older versions of the mobile clients / desktop clients.

I think this is what we should do 👍

modules/ROOT/pages/installation/system_requirements.adoc
```diff
@@ -16,42 +16,42 @@ Server
...
-|PHP Runtime 
+|PHP Runtime
 |* 5.6+, 7.0, & 7.1
 |===
 
-[NOTE] 
+[NOTE]
 ====
-ownCloud 10.1 requires a minimum php version of 7.1. 
+ownCloud 10.1 requires a minimum php version of 7.1.
```
settermjd: Is this correct? Is the support for earlier versions dropped in 10.1? Totally awesome if so.

cdamken: Well, when 10.1 comes out we have to decide ;)

phil-davis: core 10.1 or 11.0 or whatever number it is given is highly unlikely to happen before Christmas.
Death of PHP 5.6 and 7.0 security support is a 2018 Christmas gift anyway.
Given the above, the logical conclusion is PHP 7.1 minimum.

voroyam:
Okay, wrote this:

> You should always use the current/latest client versions for Windows, Linux, MacOS, iOS or Android. We test / support only the current releases.

The motivation to write this PR was to have something to point to in the documentation when people come and ask for such things like "what is the oldest client that supports 10.0.7?"

Now we have something "official".

mmattel:
@voroyam 
May I ask to add a small change to this commit fixing text in this document.
In Memory requirements:
```
ownCloud officially requires a minimum of 128MB RAM. But, we recommend a minimum of 512MB.
```
—>
```
ownCloud officially requires a minimum of 128MB RAM but recommends a minimum of 512MB.
```

```
Based on tests, server memory usage for scanning greater than 10k files uses about ...
```
—>
```
Based on tests, server memory usage for scanning more than 10k files uses about ...
```

settermjd:
@voroyam, https://github.com/owncloud/docs/pull/73#issuecomment-422943377?

voroyam:
> @voroyam
> May I ask to add a small change to this commit fixing text in this document.
> In Memory requirements:
> 
> ```
> ownCloud officially requires a minimum of 128MB RAM. But, we recommend a minimum of 512MB.
> ```
> —>
> 
> ```
> ownCloud officially requires a minimum of 128MB RAM but recommends a minimum of 512MB.
> ```
> ```
> Based on tests, server memory usage for scanning greater than 10k files uses about ...
> ```
> —>
> 
> ```
> Based on tests, server memory usage for scanning more than 10k files uses about ...
> ```


I don't agree with the first suggestion, but will work in the second one.

My reason:
ownCloud, in this sentence, is the software. A software is a non living thing. I can't recommend you stuff. We, on the other hand, the people who work at ownCloud, are a living thing, and can recommend you stuff.

This is how I see it. @settermjd is the native speaker and has as always the last word.

settermjd:
The second sentence correction suggestion looks fine to me. The first one, I'd rewrite as: "ownCloud officially requires a minimum of 128MB RAM. However, a minimum of 512MB is recommended.".

voroyam:
@settermjd I have made the required changes.

What is required to merge this?

settermjd:
@voroyam, just a nudge.

PR #87, opened 2018-09-17
added path for ransomware

voroyam:
added path in the container for the file extension blacklist

### DIFF
--- a/modules/ROOT/pages/appliance/login_information.adoc
+++ b/modules/ROOT/pages/appliance/login_information.adoc
@@ -1,5 +1,5 @@
-ownCloud Appliance Login Information
-====================================
+ownCloud Appliance Login Information and Custom Paths
+=====================================================
 
 Welcome to the ownCloud Appliance. Here are the login credentials.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/appliance/login_information.adoc
```diff
@@ -1,5 +1,5 @@
-ownCloud Appliance Login Information
-====================================
+ownCloud Appliance Login Information and Custom Paths
```
settermjd: As the document's stored inside the Appliance directory, then prefixing the header with "ownCloud Appliance" is redundant. I'd strongly suggest removing it, leaving the header as "Login Information and Custom Paths".

modules/ROOT/pages/appliance/login_information.adoc
```diff
@@ -1,5 +1,5 @@
...
+=====================================================
```
settermjd: Please use AsciiDoc headers instead of reStructuredText headers.

PR #107, opened 2018-09-20
Adding json to the nginx location directive

mmattel:
Small change: just adding json to the location directive.

Referencing core:

master: https://github.com/owncloud/core/pull/32718/ (When allowing pretty urls json files can also be loaded directly)
stable10: in progress (does not prevent merging)

### DIFF
--- a/modules/ROOT/examples/installation/nginx/subdirectory-configuration.conf
+++ b/modules/ROOT/examples/installation/nginx/subdirectory-configuration.conf
@@ -138,7 +138,7 @@
               access_log off;
           }
   
-          location ~ /owncloud/.*\.(?:svg|gif|png|html|ttf|woff|ico|jpg|jpeg|map) {
+          location ~ /owncloud/.*\.(?:svg|gif|png|html|ttf|woff|ico|jpg|jpeg|map|json) {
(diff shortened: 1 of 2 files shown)


modules/ROOT/examples/installation/nginx/subdirectory-configuration.conf
```diff
@@ -138,7 +138,7 @@
               access_log off;
           }
   
-          location ~ /owncloud/.*\.(?:svg|gif|png|html|ttf|woff|ico|jpg|jpeg|map) {
+          location ~ /owncloud/.*\.(?:svg|gif|png|html|ttf|woff|ico|jpg|jpeg|map|json) {
```
settermjd: Quick question; is there any reference to why the prefix `/owncloud/` is used, or is this a defacto convention that I've forgotten?

mmattel: Good question, will take a look

mmattel:
Answer: because we have defined in ``location ^~ /owncloud {`` that the subdirectory name is ``owncloud``. The name is free of choice.

PR #109, opened 2018-09-20
Updated the README

settermjd:
I added a "Quick Start" guide to help users get started as quickly as possible and fleshed out a section that provides users with two ways of working; one using just a text editor or IDE and a live preview plugin, and the other using Antora's command-line tools.


README.md
```diff
@@ -4,23 +4,50 @@ This project is a port of the ownCloud documentation, that was previously genera
...
+If you've been contributing to the previous version of ownCloud's documentation, which used reStructuredText and Sphinx-Doc, here's how to get started contributing to the new version of the documentation.
+Unlike the previous version of the documentation, you don't, necessarily, need to install a large number of tools and packages just to review your changes.
+
+### Only Making Text Changes?
+
+If you're only making text changes, then install either:
+
+1. The [AsciiDoc Live Preview plugin](https://asciidoctor.org/docs/editing-asciidoc-with-live-preview/) for *Firefox*, *Google Chrome*, or *Opera*
+2. The [AsciiDoc Live Preview plugin](https://asciidoctor.org/docs/editing-asciidoc-with-live-preview/#using-a-modern-text-editoride) for your text editor or IDE, *if it has one*.
+
```
mmattel: For the link description I would use:
``Browser AsciiDoc...`` and ``Editor AsciiDoc...``
Because when you render it, both look the same

README.md
```diff
@@ -4,23 +4,50 @@ This project is a port of the ownCloud documentation, that was previously genera
...
+Using one, or both, of these, you can quickly check if the changes you make are what you expect, and if there are any render errors.
+If the document renders as you expect, then you can commit the changes and push them to the docs repository.
+
+### Are You Working With Inline Code Examples, Images, and Attachments?
+
+If, however, you're working with inline code examples, images, and attachments, then you need to [install Antora's command-line tools](./docs/install-antora.md).
```
mmattel: Is it true for inline code examples too?
I thought it is only true when you link to things outside the .adoc file

mmattel: I also would add a note that this is the only way to avoid broken links before you push.

tboerger: It also makes sense for code examples to see that it's rendered like expected.

tboerger: It doesn't resolve broken links, so why should that help?

README.md
```diff
@@ -4,23 +4,50 @@ This project is a port of the ownCloud documentation, that was previously genera
...
+After that, you need to learn how to [build the docs](./docs/build-the-docs.md) from the command line and to how to review the changes in your browser.
```
tboerger: and to how to?

README.md
```diff
@@ -4,23 +4,50 @@ This project is a port of the ownCloud documentation, that was previously genera
...
-You need to do a few things to contribute to the documentation:
+### Now You're Ready to Contribute
 
-1. [Install Antora](./docs/install-antora.md).
-2. Learn [the AsciiDoc file format's basics](./docs/what-is-asciidoc.md).
-3. Update your toolset so that it supports both AsciiDoc and Antora.
-4. Learn how to [build the docs](./docs/build-the-docs.md).
-5. Create meaningful commits and a PR with your changes.
+Now that you've installed the required tools, you're ready to contribute.
+So dive on in and learn [the AsciiDoc basics](./docs/what-is-asciidoc.md).
```
tboerger: dive on in?

README.md
mmattel: @tboerger maybe misunderstood, imho rendering code examples would work with the minimal setup described above, as long they are in the text and not linked, so the need for the full installation of antora to render code examples is what I doubt as prerequisit...

mmattel: True, it will not resolve it, but as far I understand, when you do a preview with the plugin only, you never see what you would get using links versus with the full antora installation you would get it rendered if the link is correct. A link typo would render wrong and that is what you would see and then you are able to correct.

README.md
settermjd: Ah, I see your point. It took me a bit, however. :-)

README.md
settermjd: I think that I wasn't the clearest in what I said (I'll correct it). You can have code examples that directly contain the code, such as:

```asciidoc
.app.rb
[source,ruby]
----
require 'sinatra'

get '/hi' do
  "Hello World!"
end
----
```

And you can have them that have code sourced from a file, such as:

```asciidoc
[source,ruby]
----
include::app.rb[]
----
```

It's the second type that I'm referring to.

settermjd: @mmattel, not exactly. The browser/IDE/text editor plugin can properly render links that don't refer to local files, e.g., external sites, SFTP, etc. It's just links to local files where there would be an issue. Perhaps this sentence could do with a bit more depth.

README.md
settermjd: Perhaps slang that only I understand. Thanks for pointing that out.

settermjd:
@tboerger @mmattel all feedback has been addressed.

README.md
```diff
@@ -4,23 +4,53 @@ This project is a port of the ownCloud documentation, that was previously genera
...
+
+### Only Making Text Changes?
+
+If you're only making text changes, then add the AsciiDoc Live Preview plugin:
+
+1. To your [browser](https://asciidoctor.org/docs/editing-asciidoc-with-live-preview/) (The supported browsers are: *Firefox*, *Google Chrome*, or *Opera*).
+2. To your [text editor or IDE](https://asciidoctor.org/docs/editing-asciidoc-with-live-preview/#using-a-modern-text-editoride), *if it has one*.
+
+Using one, or both, of these, you can quickly check if the changes you make are what you expect, and if there are any render errors.
+If the document renders as you expect, then you can commit the changes and push them to the docs repository.
```
mmattel: I would add a text/note here that it renders text and external links. Internal links cant be rendered as discussed.

README.md
```diff
@@ -4,23 +4,53 @@ This project is a port of the ownCloud documentation, that was previously genera
...
-1. [Install Antora](./docs/install-antora.md).
-2. Learn [the AsciiDoc file format's basics](./docs/what-is-asciidoc.md).
-3. Update your toolset so that it supports both AsciiDoc and Antora.
-4. Learn how to [build the docs](./docs/build-the-docs.md).
-5. Create meaningful commits and a PR with your changes.
+Now that you've installed the required tools, you're ready to contribute.
+So it’s time to learn [the AsciiDoc basics](./docs/what-is-asciidoc.md).
+Then, you can create meaningful commits and a PR with your changes.
 
 ## Styling the Docs
```
mmattel: The sentence starts with ``If you want to change the look and feel``
Isnt it more about that we give a guideline about what we think the styling should look like?
Of course styling improvements can be made, but the text invites directly to change it...

settermjd: I think I understand what you're saying. I'll make an update to see if I do.

PR #111, opened 2018-09-21
Start to add client docs/repos

tboerger:
The referenced branches have been pushed by me, they are far away from being final but I wanted to start with that to get it into a working state.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -5,6 +5,18 @@ content:
   sources:
   - url: .
     branches: HEAD
+  - url: https://github.com/owncloud/android.git
+    branches:
+    - master-antora
+    start_path: docs/
+  - url: https://github.com/owncloud/ios.git
+    branches:
+    - master-antora
+    start_path: docs/
+  - url: https://github.com/owncloud/client.git
+    branches:
+    - master-antora
+    start_path: docs/
 
 ui:
   bundle:


settermjd:
That doesn't render/build properly, as the referenced repositories aren't based around AsciiDoc. As a stop-gap solution, links could be added to the top navigation.

bildschirmfoto 2018-09-21 um 10 47 36

settermjd:
What you do get is a link to the manuals, but the content may or may not render fully. 

bildschirmfoto 2018-09-21 um 10 48 26
bildschirmfoto 2018-09-21 um 10 49 06

tboerger:
The content have been rendered, but maybe it's not perfectly transformed from rst to adoc. You can see everything at:

* https://github.com/owncloud/android/tree/master-antora/docs/modules/ROOT/pages
* https://github.com/owncloud/ios/tree/master-antora/docs/modules/ROOT/pages
* https://github.com/owncloud/client/tree/master-antora/docs/modules/ROOT/pages

Besides adding the client repos we should also build some index page like it's already done for the old documentation, here we can take a look at the fedora documentation.

settermjd:
It might just be quicker to convert the three remaining repos to AsciiDoc. I don't imagine it would take that much time.

settermjd:
@tboerger, well that makes sense then. However, Pandoc only understands strict reStructuredText. It doesn't understand the sphinx-doc extensions.

tboerger:
But this still doesn't prevent this merge :)

The other branches are available within the correct repos to get them into the right state.

PR #115, opened 2018-09-24
User password cannot be "0"

voroyam:
In the upcoming 10.0.10 a user password cannot be set to "0".

### DIFF
--- a/modules/ROOT/pages/configuration/server/occ_command.adoc
+++ b/modules/ROOT/pages/configuration/server/occ_command.adoc
@@ -817,7 +817,7 @@ if one or more files are known to be in need of repair. Two examples of
 when files need repairing are:
 
 * If folders have the same entry twice in the web UI (known as a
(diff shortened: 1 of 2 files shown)


voroyam:
done. We should communicate our work better @mmattel

mmattel:
True 😃 
Still see that the file occ_command.adoc is present with changes (not fully reverted).
If fully reverted, the file should disappear from the change list...

mmattel:
@voroyam the easiest would be you close that PR and create a new one containing the fix for ``user_configuration.adoc``

mmattel:
This PR is now a bit messed up, because it contains an old version of occ_command.adoc which was updated in #113 (merged). Please either do what ever is necessary that the file will not longer be in the change list (maybe rebasing) or as already suggested close that PR an create a new one which only contains the changes for the user doc.

mmattel:
@voroyam 
Could you create a new PR containaing only the changes made in ``user_configuration.adoc`` ?

PR #117, opened 2018-09-25
Add PDF Theme Support

settermjd:
This PR

1. Adds a custom theme, which overrides the relevant settings, and copies in a lot of the other settings.
2. Extends the README, documenting how to generate the PDF versions of the manuals.

💁: The custom theme is required as, [asciidoctor-pdf](https://github.com/asciidoctor/asciidoctor-pdf#themes) uses a default theme. While fine, it isn't enough for ownCloud's needs, as it looks quite different to the PDF's currently being generated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This project is a port of the ownCloud documentation, that was previously genera
 Fundamentally, not that much has changed.
 
 All of the same information is still available.
-However, here’s what has changed:
+However, here's what has changed:
 
 1. The platform (and tools) used to build the documentation, which is [Antora](./docs/what-is-antora.md).
 2. The file format that the documentation is written in, which is [AsciiDoc](./docs/what-is-asciidoc.md).
@@ -12,7 +12,7 @@ However, here’s what has changed:
 
 ## Contributing to the Docs
 
-If you've been contributing to the previous version of ownCloud's documentation, which used reStructuredText and Sphinx-Doc, here’s how to get started contributing to the new version of the documentation.
+If you've been contributing to the previous version of ownCloud's documentation, which used reStructuredText and Sphinx-Doc, here's how to get started contributing to the new version of the documentation.
 
 You need to do a few things to contribute to the documentation:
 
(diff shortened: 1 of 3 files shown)


settermjd:
Just added a small section to the README, which shows how to generate PDFs (from the command-line, naturally).

README.md
```diff
@@ -25,3 +25,53 @@ You need to do a few things to contribute to the documentation:
...
+2. The custom theme directory and the custom theme file. This ensures that the defaults are overridden, where relevant, to ensure that the generated PDF is as close to the current ownCloud style as possible.
+
+### How To Generate The PDF Configuration File
+
+Book file generation isn't currently supported by Antora.
+As a result, it needs to be done separately.
+To save time and effort, you can use [the Antora Tools Phar file](https://github.com/settermjd/antora-tools/releases/download/0.0.1/antora-tools.phar), written specifically for the ownCloud docs.
+It provides a single command `antora:create-asciidoc-book-file`, which takes four options:
+
+| Setting | Description |
```
mmattel: you have no
table start/end ``|===``
``[width...``
?
Please also align ``|`` underneath
| text | more text
| text | more text

README.md
```diff
@@ -25,3 +25,53 @@ You need to do a few things to contribute to the documentation:
...
+To generate a PDF version of one or more of the manuals, you need to have `asciidoctor-pdf` installed.
+With it installed, run the command below, to generate a PDF version of a manual.
+Please be aware that, depending on the size of the manual, PDF generation may take some time.
+
+```console
+asciidoctor-pdf \
+    book.adoc \
+    -a pdf-stylesdir=resources/themes \
+    -a pdf-style=owncloud
+```
```
mmattel: From where in the directory do you start that command and what/where/name will the output be?

README.md
```diff
@@ -25,3 +25,53 @@ You need to do a few things to contribute to the documentation:
...
+
+It invokes [asciidoctor-pdf](https://github.com/asciidoctor/asciidoctor-pdf), passing it:
+
+1. [The configuration file](https://github.com/asciidoctor/asciidoctor-pdf/blob/master/docs/theming-guide.adoc) to use. This contains the list of files to use as the PDF's source material, along with front-matter. The front-matter includes details such as whether to render a table of contents, the icon set to use, and the images base directory.
+2. The custom theme directory and the custom theme file. This ensures that the defaults are overridden, where relevant, to ensure that the generated PDF is as close to the current ownCloud style as possible.
+
+### How To Generate The PDF Configuration File
+
```
mmattel: What is the difference in ``How To Generate PDFs`` and ``How To Generate The PDF Configuration File``.
What is that necessary for ?

README.md
settermjd: This is a [Markdown table](https://www.tablesgenerator.com/markdown_tables). I'll see about the formatting.

README.md
settermjd: Good point. Thanks for spotting that omission.

README.md
settermjd: The first one generates the physical (electronic) PDF file. The other generates its configuration file.

README.md
settermjd: I'm starting to wonder if aligning the table content is going to be workable over time; primarily as the auto-formatter plugins may override that, by default.

README.md
mmattel: 1 is clear, but what is 2 good for and where is it needed - not clear.

From my initial reading and understanding, you would if you want to create a pdf do the configuration first and not afterwards

settermjd:
Most of the feedback's been addressed, @mmattel.

Issue #79, opened 2018-09-10
Antivirus - missing occ command background true/false


voroyam:
missing occ command

The occ command is not yet in the documentation. As soon as it is, it should be documented here.

The command is question sets a value for the antivirus background job.

@cdamken for more info

settermjd:
Do you mean that an existing occ command for working with an Antivirus tool is missing, @voroyam?

voroyam:
As far as I understand there is occ command or it's planned to create one.

Then, when it's available, it should be documented.

VicDeo:
Why invent the wheel here?
https://doc.owncloud.org/server/10.0/admin_manual/configuration/server/occ_command.html#config-commands-label
```
config:app:set         Set an app config value
```
can be used for any app/any value

settermjd:
Does @VicDeo's [comment](https://github.com/owncloud/docs/issues/79#issuecomment-420693155) answer your question, @voroyam?

voroyam:
not really, because it's still not clear what value is to be set to configure the background anti virus check.

voroyam:
This can be closed imho, I don't know the command, don't know what to document.

Issue can be re-opened if we have further info.

mmattel:
@voroyam FYI
for apps not part of core like ldap or brute force protection, use the occ_apps_command.adoc file.
In particular when it comes to antivirus you can look at the brute force protection.
The only thing to be identified are the strings and their defaults which must be taken out of the code.

VicDeo:
According to https://github.com/owncloud/files_antivirus/pull/213/files#diff-0bd1ff9f0e0e556dfda091330bf237dfR53
options is called `av_scan_background`

mmattel:
Did a small research on the file above.

Following are the existing keys with their default values:
```
'av_mode' => 'executable',
'av_socket' => '/var/run/clamav/clamd.ctl',
'av_host' => '',
'av_port' => '',
'av_cmd_options' => '',
'av_path' => '/usr/bin/clamscan',
'av_max_file_size' => -1,
'av_stream_max_length' => '26214400',
'av_infected_action' => 'only_log',
'av_scan_background' => 'true',
```

Following are the existing keys with their possibe values, type and  text:
```
'av_mode' => 'executable', 'daemon', 'socket' (string)
'av_socket' => '/var/run/clamav/clamd.ctl', (string)
'av_host' => '', (string, 'Hostname or IP address of Antivirus Host')
[...]

PR #122, opened 2018-09-25
Document occ app files_antivirus

mmattel:
References: https://github.com/owncloud/docs/issues/79 (Antivirus - missing occ command background true/false)

Adds the documentation how to configure ``files_antivirus`` in the ``occ_apps_command`` file.

pls note that the description text is taken from the app.

(I also changed the sort order of the rows of the tables in brute force to be inline with files_antivirus)

### DIFF
--- a/modules/ROOT/pages/configuration/server/occ_app_command.adoc
+++ b/modules/ROOT/pages/configuration/server/occ_app_command.adoc
@@ -19,6 +19,7 @@ occ Apps Command Directory
 * xref:ocapps_brute_force_protection[Brute-Force Protection]
 * xref:ocapps_calendar[Calendar]
 * xref:ocapps_contacts[Contacts]
+* xref:ocapps_files_antivirus[Anti-Virus]
 * xref:ocapps_files_primary_s3[S3 Objectstore]
 * xref:ocapps_ldap[LDAP]
 * xref:ocapps_market_commands[Market]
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/configuration/server/occ_app_command.adoc
```diff
@@ -263,6 +264,141 @@ Marketplace URL: https://marketplace.owncloud.com/apps/contacts[Contacts]
 
 For commands for managing contacts, please see the DAV Command section in the occ core command set.
 
+[[files_Antivirus]]
+Anti-Virus
+----------
+
+Marketplace URL: https://marketplace.owncloud.com/apps/files_antivirus[Anti-Virus]
+
+Use these commands to configure the Anti-Virus app.
```
settermjd: It makes more sense, to me, to link "the Anti-Virus app", instead of having the link text on line 271.

modules/ROOT/pages/configuration/server/occ_app_command.adoc
```diff
@@ -263,6 +264,141 @@ Marketplace URL: https://marketplace.owncloud.com/apps/contacts[Contacts]
...
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+....
+sudo -u www-data php occ config:list files_antivirus
+....
+
+Set the setting
+~~~~~~~~~~~~~~~~
+
+To set a new value, use the command below and replace <Key> and value <Value> accordingly.
```
settermjd: I'd suggest the following so that "Key" and "Value" stand out more:

```rst
`<Key>` and value `<Value>` accordingly
```

modules/ROOT/pages/configuration/server/occ_app_command.adoc
mmattel: But the way it is done is what you requested and how I implemented it and you approved ...
It renders like that:
image
Please advice

modules/ROOT/pages/configuration/server/occ_app_command.adoc
mmattel: NP, will do including in brute force as we have the same thing there.

modules/ROOT/pages/configuration/server/occ_app_command.adoc
settermjd: When I was reading the text just now, it seemed that it would flow more naturally, as I suggested. But now that you mention it, I see that I was wrong.

mmattel: pfff  😄

PR #132, opened 2018-10-02
Remove remaining reStructuredText formatting

settermjd:
This PR removes any remaining reStructuredText formatting. Besides completing the migration from sphinx-doc to Antora, this PR also ensures that the content can be exported to PDF format (which rST formatting prevents).

Yes, it's a huge commit, at least judging by the number of files changed. However, aside of some whitespace changes that slipped through, here's what's changed:

- rST level-one through level-four headers have been converted to the native AsciiDoc format equivalent (= - level one header
- Fixed a set of broken AsciiDoc formatting


modules/ROOT/pages/configuration/files/external_storage/auth_mechanisms.adoc
```diff
@@ -36,7 +33,7 @@ credentials, and background file scanning does not work.
 
 [NOTE]
```
mmattel: Question:

This writes ``[NOTE]``

while in https://github.com/owncloud/docs/edit/master/modules/ROOT/pages/configuration/server/occ_command.adoc#L2852
we write ``Note :``
and in 
https://github.com/owncloud/docs/edit/master/modules/ROOT/pages/configuration/server/occ_app_command.adoc#L656
we write ``NOTE:``

Which one should we take?

settermjd: This is an example of [an admonition block](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#admon-bl), which can contain more complicated/sophisticated content.

PR #135, opened 2018-10-02
post installation script update

mmattel:
This is an update of the post installation script.

Following new:
- a bit more description
- automation if you want to upgrade an existing installation
- asks for the tar file
- renames the original directory including a timestamp
- creates a new directory with the correct name
- extracts correctly or stops in case of an error during extracting


modules/ROOT/examples/installation/post-installation-steps.sh
```diff
@@ -1,11 +1,25 @@
...
+# ocname	the name of your directory containing the owncloud files
+# ocroot	the path to ocname, usually /var/www
+# linkroot	the path to your source directory for linking data and apps-external
+# htuser	the webserver user
+# htgroup	the webserver group
+# rootuser	the root user
+
+ocname='owncloud'
+ocroot='/var/www/'
+ocpath=$ocroot$ocname
```
settermjd: How about quoting this and the following variables as an extra sanity check? Also, while it's a little more work, okay one character per variable, how about changing `ocroot` to `'/var/www'`, and then initialising the others with a forward slash between the variables, e.g., `ocpath=$ocroot/$ocname`. It could just be me, but I find compound variables easier to read when declared that way. Plus, I can't say for certain, but I thought it was an informal convention in shell scripts.

modules/ROOT/examples/installation/post-installation-steps.sh
```diff
@@ -17,15 +31,24 @@ rootuser='root'
...
+# New install using mkdir:     n/n/n/n (create missing directories, setup permissions and ownership)
+# Upgrade using mkdir:         y/n/n/n (you move/replace data, apps-external and config.php manually, set setup permissions and ownership)
+# New install using links:     n/y/y/n (link existing directories, setup permissions and ownership)
+# Upgrade using links:         y/y/n/y (link existing directories, copy config.php, permissions and ownership are already ok)
+# Post installation/upgrade:   either n/n/n/n or n/y/y/n
+# Reset all perm & own:        either n/n/n/n or n/y/y/n
 
 echo
+read -p "Do you want to upgrade an existing installation (y/N)? " -r -e answer
+if echo "$answer" | grep -iq "^y"; then
```
settermjd: You could simplify the if/else to `(echo "$answer" | grep -iq "^y") && do_upgrade="y" || do_upgrade="n"` instead, as it's just a simple test + assignment.

modules/ROOT/examples/installation/post-installation-steps.sh
```diff
@@ -47,6 +70,23 @@ else
   upgrdcfg="n"
 fi
 
+# check if to upgrade an existing installation
+if [ "$do_upgrade" = "y" ]; then
+  read -p "Please specify the file to extract with full path: " -r -e answer
```
settermjd: I see that the convention you're using is to use `$answer` for storing user input, which makes perfect sense, but would it not make more sense to name the variable based on it's intended use?

mmattel: it is just the given answer.
more important is the variable having the name which highlights its itention.

modules/ROOT/examples/installation/post-installation-steps.sh
```diff
@@ -83,7 +124,8 @@ else
     exit
   else
     printf "ln $ocapps_external \n"
-    ln -sfn $linkapps-external $ocapps_external
+	mkdir -p $linkapps_external
+    ln -sfn $linkapps_external $ocapps_external
```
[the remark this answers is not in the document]
mmattel: ups, this was not intended, it is a tab instead blanks

modules/ROOT/examples/installation/post-installation-steps.sh
settermjd: Fair point. It was worth raising.

modules/ROOT/examples/installation/post-installation-steps.sh
```diff
@@ -47,6 +70,23 @@ else
...
+  if [ ! -f "$answer" ]; then
+    echo "File to extract not found. Exiting."
+    echo
+    exit
+  fi
+  if [ -d ${ocpath} ]; then
+    mv $ocpath $oldocpath
+  fi
+  mkdir -p $ocpath
+  if ! tar xvf "$answer" -C $ocpath --strip-components=1 ; then
```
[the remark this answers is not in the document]
mmattel: Tried it now for hours. Atm where I put any kind of brackets around, it will no longer work...

mmattel:
Implemented recommendations.
Added one check before chmod to check, if files are present, else chmod would return an error. Identified when playing around with the tar command (if tar would succeed but no files have been extracted... very rare but possible)

mmattel:
I just adopted the script´s path to the new administration_manual path instead of root 😄

modules/administration_manual/examples/installation/post-installation-steps.sh
```diff
@@ -126,6 +164,7 @@ if [ "$chmdir" = "y" ] && [ "$uselinks" = "y" ]; then
   find -L ${ocapps_external}/ -type d -print0 | xargs -0 chmod 0750
 fi
 
+#chown
```
settermjd: Can you make this more descriptive, pls?

settermjd: Am I missing something? It doesn't seem to be visible.

mmattel:
I rebased and pushed. Now it is here 😄 
Sorry for the confusion...

modules/administration_manual/examples/installation/post-installation-steps.sh
settermjd: Unless mine's out of date, I don't see it, still.

Issue #97, opened 2018-09-19
PDF Generator


settermjd:

settermjd:
Closing as the related/implementing PRs have been merged.

Issue #92, opened 2018-09-19
Finish Antora Docs


settermjd:
## TODO 

- [x] Review labels
- [ ] Create a writing style guide https://github.com/owncloud/docs/issues/95
- [x] Augment the build process https://github.com/owncloud/docs/issues/96
  - [ ] Dead/broken link checker (@mmattel)
  - [x] rST formatting checker (if there is any leftover rST formatting)
- [x] PDF generator (@settermjd, @pvince81)
- [x] ~~Algolia subscription (see @pmaier1)~~
- [x] Implement Site Search https://github.com/owncloud/docs/issues/98
- [x] Update project README & documentation
  - [x] Document difference between when you should use the live preview plugin and when you should use Antora build tools
- [x] New Issue template (including "what needs to be documented (or a description), and "why is this being documented" (or what is the motivation), etc. A good example is https://github.com/owncloud/docs/issues/39 (thanks @voroyam).
- [x] Create redirects for old links

Leantricity:
I'd love to collaborate here. How to start?

settermjd:
Hey @Leantricity, thanks for wanting to contribute! Do you want to help out with this issue specifically, or contribute to the docs as a whole? 

Eitherway, start off by reading [the "Getting Started" guide](https://github.com/owncloud/docs/blob/master/docs/getting-started.md), so that you're abreast of what you need to know to get involved. 

Then, if you have any questions, if anything isn't clear enough, or if the docs could be written better, just let me know. I'm in the #documentation channel in https://talk.owncloud.com/home daily. 

Happy to answer any and all questions!

Leantricity:
Hi Mathew!

I want to collaborate as a whole, mostly if this is a goal, creating the Spanish version of the documentation (I think it does not exist yet).

I will read this Getting Started guide and follow up. I’ve already installed all the Antora required stuff and downloaded the first docs-ui folder as I saw I had to do but seems a demo setup not with the current ownCloud docs.

I will try to learn how to navigate here (I’m new to GitHub also…)

Thanks!

JP

settermjd:
Hoping to finalise the site search tomorrow.

settermjd:
Closing as it's sufficiently completed. The style guide can be left for later.

PR #136, opened 2018-10-05
Use Custom Font in the Custom ownCloud PDF Theme

settermjd:
This PR updates the custom PDF theme to use two custom fonts, *Times New Roman* and *Courier*, in place of the bundled Times Roman and Courier fonts. This is because the bundled fonts don’t support a wide enough character set to render the manuals correctly.

### Relates To 

#92 & #97.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,24 @@ asciidoctor-pdf \
     -a pdf-style=owncloud
 ```
 
+---
+
+**Note:** the custom ownCloud PDF theme references a custom font, *Times New Roman*, in place of the packaged Times Roman font, as Times Roman doesn't support a wide enough character set to render the manuals correctly.
+
+If you have this font available on your system, you need to copy it (all four variants: _normal_, _italic_, _bold_, and _bold-italic_), to the `fonts` directory. If you don't, then you'll have to copy the closest matching font family to _Times New Roman_ that you do have, and then update `resources/themes/owncloud-theme.yml`.
+
+If you want to use your available system fonts, here’s where you can find them:
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -39,6 +39,13 @@ asciidoctor-pdf \
     -a pdf-style=owncloud
 ```
 
+---
+
+**Note:** the custom ownCloud PDF theme references a custom font, *Times New Roman*, in place of the packaged Times Roman font, as Times Roman doesn't support a wide enough character set to render the manuals correctly.
+
+If you have this font available on your system, you need to copy it (all four variants: _normal_, _italic_, _bold_, and _bold-italic_), to the `fonts` directory. If you don't, then you'll have to copy the closest matching font family to Times New Roman that you do have, and then update `resources/themes/owncloud-theme.yml`.
+
```
mmattel: where did you got the custom font from?
that might be a valuable note for others... (...If you don't, then you'll have to...)

mmattel: This would be at least a good hint.
eg, on macOS, you can find the the font <whatever you have taken and is the closest match> in <....>, please look on other OS accordingly (make that sentence looking nice, but you understand the intention).
How is the naming of the font family files, do they have a particular extension (for eas of search),
are these fonts maybe available to be downloaded on the internet?

settermjd: Well said. I'll make the required changes.

settermjd: 277cc1b updates the previous commit to:

- list the default font directories (as best I understand them)
- link to further information per platform
- link to font directories

PR #142, opened 2018-10-09
Fix build issues

settermjd:
This PR corrects a series of small issues that cause the build process to throw warnings and errors.

### DIFF
--- a/modules/ROOT/pages/configuration/user/user_provisioning_api.adoc
+++ b/modules/ROOT/pages/configuration/user/user_provisioning_api.adoc
@@ -147,7 +147,7 @@ sending a Basic HTTP Authorization header.
 [cols=",,",options="header",]
 |==============================================================
 |Request Path |Method |Content Type
-|`Syntax: ocs/v1.php/cloud/users/{userid}` |`GET` |`text/plain`
+| `Syntax: ocs/v1.php/cloud/users/\{userid\}` | `GET` | `text/plain`
 |==============================================================
 
 [cols=",,",options="header",]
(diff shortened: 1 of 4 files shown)


modules/ROOT/pages/configuration/user/user_provisioning_api.adoc
```diff
@@ -147,7 +147,7 @@ sending a Basic HTTP Authorization header.
 [cols=",,",options="header",]
 |==============================================================
 |Request Path |Method |Content Type
-|`Syntax: ocs/v1.php/cloud/users/{userid}` |`GET` |`text/plain`
+| `Syntax: ocs/v1.php/cloud/users/\{userid\}` | `GET` | `text/plain`
```
mmattel: You are using a blank between `|`` and the next character, this is good and improves readability !
Also used in other tables like occ commands.
To be consistent, you should do that on all following tables too...

mmattel:
Lookes good, a small thing to improve

modules/ROOT/pages/configuration/user/user_provisioning_api.adoc
settermjd: That somehow slipped in. It was meant to only be a part of #143.

mmattel: without double checking, is this then addressed in #143 ?
If not pls add there as this PR is mergen now

settermjd: Yep, the change is there. As I said, this crept through, by accident.

PR #144, opened 2018-10-09
Backport owncloud/documentation#4405 to the new docs

### LINKED ISSUE #94: Document that occ config:system:set accepts values in JSON format
Referncing documentation: https://github.com/owncloud/documentation/pull/4405
Needs checking that the referenced PR gets implemented here in docs

@settermjd @phil-davis

settermjd:
This backport documents that occ config:system:set accepts values in JSON format, in addition to boolean, integer, float, and string. It fixes #94.

### DIFF
--- a/modules/administration_manual/pages/configuration/server/occ_command.adoc
+++ b/modules/administration_manual/pages/configuration/server/occ_command.adoc
@@ -453,20 +453,47 @@ sudo -u www-data php occ config:system:set \
 Value not updated, as it has not been set before.
 ....
 
-Note that in order to write a Boolean, float, or integer value to the
-configuration file, you need to specify the type on your command. This
-applies only to the `config:system:set` command.
+Note that in order to write a boolean, float, JSON, or integer value to the configuration file, you need to specify the type on your command.
+This applies only to the `config:system:set` command.
+The following values are known:
 
-When you want to e.g., disable the maintenance mode run the following
-command:
+* `boolean`
+* `integer`
+* `float`
+* `json`
+* `string` (default)
+
+Examples
+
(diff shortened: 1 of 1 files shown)


modules/administration_manual/pages/configuration/server/occ_command.adoc
```diff
@@ -453,22 +453,49 @@ sudo -u www-data php occ config:system:set \
...
+
+[source,console]
+....
+sudo -u www-data php occ config:system:set apps_paths \
+      --type=json \
+      --value='[
+        {
+            "path":"/var/www/owncloud/apps",
+            "url":"apps",
+            "writable":1
```
mmattel: even it works with ``0`` or ``1``
we have in all othe documents ``true`` or ``false``...
in the particular path for ``apps`` it is ``false``
for ``apps-external`` it is ``true``
which maches the othe docs too...

settermjd: Right you are. Thanks for spotting that one.

settermjd: This has been ammended in 21aa9b7.

mmattel:
May I ask for a favor, mind to add a little change:
Some lines above we have in the table following line:

--type=[TYPE] | Value type [string, integer, double, boolean, json]. [default: "string"].
-- | --

The change is to add ticks around the values as we do that on other places too

--type=[TYPE] | Value type [ `string`, `integer`, `double`, `boolean`, `json`]. [default: `string`].
-- | --

settermjd:
That's fine with me. Actually, would you mind making that change?

PR #154, opened 2018-10-12
Update PDF Configuration File

settermjd:
This PR contains a series of updates to the PDF configuration file to both make the process simpler, and to improve the quality of the finished files. The changes are:

- Replace proprietary fonts with open fonts
- Use a base font colour with a lower level of contrast (making the content easier to read)
- Increase the header font sizes to make them stand out better
- Update the footer to make it easier to read
- Remove note about need for custom fonts


settermjd:
I think that this PR should have been split into two separate PRs. Will do that from now on.

PR #153, opened 2018-10-11
Generate pdfs within Drone CI

tboerger:
Now the PDF documentation can be built within Drone, but where do we want to put it now?

PR #171, opened 2018-10-19
Fix missing images and paths

mmattel:
Here are images that have been missed to migrate plus paths corrected to access the images.

Identified via PR https://github.com/owncloud/docs/pull/153 (Generate pdfs within Drone CI)

For a list / details pls see issue comment: https://github.com/owncloud/docs/pull/153#issuecomment-431045778


tboerger:
Have you also tried if a `./` instead of the full path works?

settermjd:
Image references in Antora need to provide the path to the image _relative to the module's image directory_. For example, if the module name is `administration_manual`, then the path must be relative to `modules/administration_manual/assets/images/`. 

The same rule applies to images in other modules. The only difference is that you have to include the module's name in the image reference, **before** the relative path to the image. So, if you want to reference an image in a module named `developer_manual`, then the image reference would begin with `image:developer_manual:` and then specify the relative path to the image.

PR #176, opened 2018-10-23
Fix encryption type typo

paurakhsharma:


modules/administration_manual/pages/configuration/files/encryption_configuration.adoc
```diff
@@ -265,7 +265,7 @@ Then, enable the user-key, using the following command:
 
 [source,console]
 ....
-occ encryption:select-encryption-type user-key
+occ encryption:select-encryption-type user-keys
```
mmattel: Please write the complete command:
``sudo -u www-data php occ encryption:select-encryption-type user-keys``
as we have that in all commands described in occ commands // occ apps commands

settermjd: Well said, @mmattel. That's something to add to the upcoming style guide. Thanks for contributing the PR @paurakhsharma. Very much appreciated.

paurakhsharma:
@settermjd @mmattel, Thank you for the suggestion, as all commands in encryption do not have complete commands so, this will be confusing to have only this line with full command. If we want full commands in all commands then, So, in my opinion, this PR should only fix the above-mentioned typo. And if required new PR should be made where all existing commands are replaced with full commands.

What do you guys think? Please let me know.

phil-davis:
For this PR, I agree with @paurakhsharma 
The whole of this page has ``occ`` commands without ``sudo`` at the front. It is a different issue to decide if we want to write (guess) the ``sudo`` stuff at the start of every command example.

settermjd:
Well said @paurakhsharma. As such, I'm fine with the PR. A future PR can correct the reference to the command throughout the entire file.

PR #179, opened 2018-10-24
Fix build breaking issues

settermjd:
The title's a little overstated, but it fixes all of the warnings that were coming from asciidoctor-pdf during the build process, which were caused by duplicate anchors (ids in Antora-speak), text that it couldn't render, fonts that only support Win/US-Ansii, and missing or misnamed fonts.


tboerger:
These are all also the anchor links within the html documentation, right? In that case the anchors are getting worse pretty much...

settermjd:
Yeah, they are. Okay, I think that I'm being lazy here and need to fix them all. I'd passed on this because I thought it'd be massive. Maybe it's time to bite the proverbial bullet and fix them all.

settermjd:
Thanks for the push, @tboerger. Oh, and in answer to your question, yes they're the HTML anchors too.

settermjd:
After going down a number of dead ends, I ended up doing a project-wide search for any references to the ids (anchors) that were causing Asciidoctor-PDF to throw warnings. Turns out, nothing references them. So, afaik, there appears to be no harm in changing them as no broken links should result.

settermjd:
If you're happy with it, pls give it a 👍.

modules/developer_manual/pages/app/advanced/l10n.adoc
```diff
@@ -5,7 +5,7 @@ https://www.transifex.com/projects/p/owncloud/[Transifex]. To start
 translating sign up and enter a group. If translations for your community app should be
 added to Transifex follow the steps decribed at the end of this page.
 
-[[php]]
+[[translation-php]]
```
mmattel: Is there a reason why you named them ``translation`` instead of ``l10n`` while the others are more or less named after the filename like ``storage-backend``.
Means as an example ``translation-php`` (btw underline instead of dash) --> ``l10n_php``

modules/developer_manual/pages/app/advanced/storage-backend.adoc
```diff
@@ -1,6 +1,6 @@
 = Storage Backends
 
-[[external-storage-backends]]
+[[storage-backend-external-storage-backends]]
```
mmattel: ``storage-backend-external-storage-backends`` --> ``storage-backend_external-storage-backends``
pattern is name undersore topic
example represents for all of them

modules/developer_manual/pages/app/fundamentals/database.adoc
```diff
@@ -179,7 +179,7 @@ class Author extends Entity {
 }
 ----
 
-[[types]]
+[[database-types]]
```
mmattel: underscore

modules/developer_manual/pages/app/fundamentals/publishing.adoc
```diff
@@ -180,7 +181,8 @@ that:
 * major releases optionally featured on https://owncloud.org and sent to
 the owncloud-announce list.
 
-image:app/app-tile-verified.jpg[ownCloud "Verified" tag]
+.ownCloud "Verified" tag
+image:app/app-tile-verified.jpg[ownCloud 'Verified' tag]
 
 [[trusted]]
```
mmattel: same as above, maybe more ?

modules/developer_manual/pages/app/fundamentals/publishing.adoc
```diff
@@ -164,7 +164,8 @@ Apps with the `Enterprise` tag are official ownCloud enterprise apps.
 These can only be uploaded by ownCloud itself and represent ownCloud
 Enterprise Edition features.
 
-image:app/app-tile-enterprise.jpg[ownCloud "Enterprise" tag]
+.ownCloud "Enterprise" tag
+image:app/app-tile-enterprise.jpg[The ownCloud 'Enterprise' Tag]
 
 [[verified]]
```
mmattel: why don´t you have, for the sake of completeness:
``verified`` --> ``publishing_verified`` ?

modules/developer_manual/pages/app/fundamentals/testing.adoc
```diff
@@ -4,7 +4,7 @@ All PHP classes can be tested with http://phpunit.de/[PHPUnit],
 JavaScript can be tested by using
 http://karma-runner.github.io/0.12/index.html[Karma].
 
-[[php]]
+[[testing-php]]
```
mmattel: underscore

modules/developer_manual/pages/app/tutorial/javascript_and_css.adoc
```diff
@@ -2,7 +2,7 @@
 
 To create a modern web application you need to write xref:app/fundamentals/js.adoc[JavaScript ] and xref:app/fundamentals/css.adoc[CSS].
 
-[[javascript]]
+[[js-and-css-javascript]]
```
mmattel: underscore
``js-and-css-javascript`` --> ``js-and-css_javascript``

modules/developer_manual/pages/core/acceptance-tests.adoc
```diff
@@ -68,7 +68,7 @@ default:
 
 This folder can be used in tests to store temporary data.
 
-[[features]]
+[[acceptance-tests-features]]
```
mmattel: underscore

modules/developer_manual/pages/core/configfile.adoc
```diff
@@ -111,7 +111,7 @@ $CONFIG = [
 ]
 ----
 
-[[using-alternative-app-directories]]
+[[configfile_using-alternative-app-directories]]
```
mmattel: underscore

modules/developer_manual/pages/core/ocs-recipient-api.adoc
```diff
@@ -64,7 +64,7 @@ Get all shares from the user.
 [[curl]]
 ==== Curl
 
-[[php]]
+[[ocs-recipient-api-php]]
```
mmattel: underscore

modules/developer_manual/pages/core/ocs-share-api.adoc
```diff
@@ -6,10 +6,10 @@ pre-defined OCS calls.
 The base URL for all calls to the share API is:
 _<owncloud_base_url>/ocs/v1.php/apps/files_sharing/api/v1_
 
-[[local-shares]]
+[[ocs-share-api-local-shares]]
```
mmattel: underscore
for all items below too

modules/developer_manual/pages/core/theming.adoc
```diff
@@ -43,7 +43,7 @@ include::{examplesdir}/scripts/theme-bootstrap.sh[]
 include::{examplesdir}/scripts/read-config.php[]
 ----
 
-[[how-to-create-a-new-theme]]
+[[theming-how-to-create-a-new-theme]]
```
mmattel: undescore

modules/developer_manual/pages/core/unit-testing.adoc
```diff
@@ -339,7 +339,7 @@ JavaScript unit test examples can be found in apps/files/tests/js/ Unit
 tests for the core app JavaScript code can be found in
 core/js/tests/specs
 
-[[documentation]]
+[[unit-testing-documentation]]
```
mmattel: undescore

modules/developer_manual/pages/general/debugging.adoc
```diff
@@ -50,7 +50,7 @@ printed to Web server log files or command line output. Instead, errors
 are generally stored in ownCloud’s own log file, located at:
 /data/owncloud.log.
 
-[[using-alternative-app-directories]]
+[[debugging-using-alternative-app-directories]]
```
mmattel: underscore

tboerger:
Why underscores instead of dashes?

mmattel:
@settermjd did the following, which I found a very good idea:

``file-name-if-longer-with-dashes`` ``_`` ``topic-name-if-longer-with-dashes``

This makes it logic and better readable, escpecially if you ``xref`` that from another file.
You easily see where you are referencing to !

tboerger:
Maybe this works good for xrefs, but IMHO it reads pretty poorly within URLs. Beside that, to me, it sounds more like an `asciidoctor-pdf` issue that it can't differ between the tags per document. Using GLOBALLY unique anchors sucks pretty much.

mmattel:
@tboerger you are right, I have to admit the example with xref was not that good ...
It is as you said a pdf issue and we have to deal with it.
My point was only to have the use of underscore (or dash if that is preferred) consistent, but not mixed up.

tboerger:
I would really prefer short anchors, so maybe we should report the pdf issue upstream and look for an answer to get a proper fix instead of working around it :)

mmattel:
@settermjd maybe you could do the following:
Create a new PR fixing the font stuff so we can see if these issues will go away.
In parallel we can work on that topic.
This would bring in more flexibility

modules/developer_manual/pages/app/advanced/l10n.adoc
settermjd: I think there was at the time, but I don't remember them atm.

modules/developer_manual/pages/app/advanced/storage-backend.adoc
settermjd: Right you are.

modules/developer_manual/pages/app/fundamentals/publishing.adoc
settermjd: I don't follow what you mean here. All I did was to change the about text and to add a caption; same for the image below it.

modules/developer_manual/pages/app/fundamentals/publishing.adoc
settermjd: See above.

settermjd:
I'll chase the issue up in the Antora forums. Thanks for pointing this out @tboerger @mmattel

settermjd:
I know that this has been dragging for a little while now. So, in an attempt to get it resolved, I've reverted d69352e691fb9f938ddf3cdcd869aa53d004c2e7, which removed the duplicate anchors (ids). 

I've also asked about why there seems to be a restriction on using the same id/anchor name in different files, which comes from [this line](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/parser.rb#L1628) in the asciidoctor source, in [the asciidoctor forums](https://gitter.im/asciidoctor/asciidoctor). 

I've yet to get a resolution, but feel that this PR can be merged and then rebased on the Drone PDF PR. I'll update the PR when I hear more.

PR #183, opened 2018-10-24
More images to be processed

mmattel:
More images to be processed

Deleting all those images which ARE use in a file, BUT are stored on two locations,
where one location is correct and the other location is dead (unused, outside /assets/images/).

The last PR was to delete all orphand images (= they are nowhere used in a file)
This PR deletes the double present but unused version (all outside /assets/images/).

To check the result if there are images accidentially deleted:

The docs directory eg ``/var/www/owncloud/docs/modules/``
for each ``administration_manual``, ``developer_manual``, ``user_manual``
in directory ``pages``

```
grep -r ".png\[" | sed -e 's/\[image\]//g' | sed 's/^[^:]*://g' | sed 's/^[^:]*://g' | sort -u | awk -F'[' '{print $1}' | xargs -I {} ls ../assets/images/{} | grep "No such"
```
If you want to see what this command does, let it run up to each pipe ``|`` and watch the result...
There are for sure more effective ways, but I made it step by step to monitor the result.

One funny thing, I found an image which was NOWHERE present but requested in a file !
It was also not available in documentation !
I needed to download it from docs.owncloud.org...

If this PR gets merged, there should be no more orphand or dead images in docs :v: 😄


mmattel:
I am just a obstinately searcher on google 😄

PR #189, opened 2018-10-31
Implement site search

settermjd:
This PR implements site search into the documentation, powered by Lunr. There is an as yet unresolved bug generating the documentation with a Docker container, but the documentation can be generated from the command-line successfully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM antora/antora
+
+LABEL maintainer="matthew@matthewsetter.com"
+
+# Enable Lunr site search integration and generation of a Lunr search index by default 
+ENV DOCSEARCH_ENABLED true
+ENV DOCSEARCH_ENGINE lunr
+
+WORKDIR /antora
+
+COPY package* /antora/
+
+RUN apk add --no-cache build-base libressl-dev libcurl libgit2-dev python && \
+    ln -s /usr/lib/libcurl.so.4 /usr/lib/libcurl-gnutls.so.4 && \
+    BUILD_ONLY=true npm install nodegit
+
+# Install the required (custom) NPM packages. 
+# This is required as the ownCloud build requires extra packages so that Lunr
+# search can be integrated.
+RUN npm i
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
   ],
   "dependencies": {
     "@antora/cli": "^1.1.1",
-    "@antora/site-generator-default": "^1.1.1"
+    "@antora/site-generator-default": "^1.1.1",
+    "antora-lunr": "v0.1.0"
   },
   "description": "The ownCloud core documentation (created with Antora and AsciiDoc)",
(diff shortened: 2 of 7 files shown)


settermjd:
@tboerger, wondering if you could have a look at the Dockerfile to see what I might have missed?

settermjd:
@tboerger, this works without the Docker container. Time-permitting I can implement changes to get the Docker container to build the docs.

site.yml
```diff
@@ -1,5 +1,6 @@
 site:
   title: ownCloud Documentation
+  url: http://localhost:5000
```
tboerger: Hardcoded to localhost doesn't look good.

package.json
```diff
@@ -0,0 +1,35 @@
...
+  "version": "10.0.10",
+  "description": "The ownCloud core documentation, created with Antora and AsciiDoc",
+  "directories": {
+    "doc": "docs"
+  },
+  "dependencies": {
+    "@antora/cli": "^1.1.1",
+    "@antora/site-generator-default": "^1.1.1",
+    "antora-lunr": "v0.1.0",
+    "lunr": "v2.3.3",
```
tboerger: This is a peer dependency and not a direct dependency, right?

package.json
```diff
@@ -0,0 +1,35 @@
...
+    "write-good": "^0.13.1"
```
tboerger: Where is that required?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Enable Lunr site search integration and generation of a Lunr search index by default 
+ENV DOCSEARCH_ENABLED true
+ENV DOCSEARCH_ENGINE lunr
+
+WORKDIR /antora
+
+COPY package* /antora/
+
+RUN apk add --no-cache build-base libressl-dev libcurl libgit2-dev python && \
```
tboerger: Where are these packages required?

site.yml
settermjd: True. I got stuck here and need to fix this.

package.json
settermjd: I had to do [some reading](https://lexi-lambda.github.io/blog/2016/08/24/understanding-the-npm-dependency-model/) (_always good fun_ 👍) to know exactly what you meant. If I understand you correct, yes, it's a peer dependency. And [antora-lunr's package.json](https://github.com/Mogztter/antora-lunr/blob/master/package.json) shows that it's already a dependency. So it seems that I've added it unnecessarily. _Will fix._

package.json
settermjd: That slipped in. Will remove.

Dockerfile
settermjd: I installed them as the build failed without them.

site.yml
settermjd: I changed this to something akin to a Docker environment variable.

Issue #190, opened 2018-11-01
Broken link to install Antora command line tools


mmattel:
## What Type Of Content Change Is This?

- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [x] Bug Fix to Existing Content

## Which Manual Does This Relate To?

- [ ] Admin Manual
- [ ] Developer Manual
- [ ] User Manual

In https://github.com/owncloud/docs/blob/master/docs/getting-started.md

**The link to install Antora's command-line tools is dead:** https://github.com/owncloud/docs/blob/master/docs/docs/install-antora.md

Wrong: https://github.com/owncloud/docs/blob/master/docs/docs/install-antora.md
Correct: https://github.com/owncloud/docs/blob/master/docs/install-antora.md

## What Needs to be Documented?
Reestablish broken link

## Why Should This Change Be Made?
Fix

@settermjd , I asigned you this case as you originally wrote the Getting Started Guide

tboerger:
Can be closed?

mmattel:
sure, thanks

PR #192, opened 2018-11-01
fix link to install antora

mmattel:
Fix for issue: https://github.com/owncloud/docs/issues/190 (Broken link to install Antora command line tools)
``install-antora.md`` is in the same directory as ``getting-started.md``


mmattel:
Added a commit because I found another file with the same problem

Issue #182, opened 2018-10-24
Check documents for image titles 


mmattel:
## What Type Of Content Change Is This?

- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [x] Bug Fix to Existing Content

## Which Manual Does This Relate To?

- [x] Admin Manual
- [x] Developer Manual
- [x] User Manual

## What Needs to be Documented?

The name of the image that will be desplayed

## Why Should This Change Be Made?

Optical representation and better understanding what the image is referring to.

Consider having the documentation in ``/var/www/owncloud/docs/modules`` 
and you want to look for all files where you have an empty image title ``[image]``.

Running the command ``grep -rl --include=\*.adoc "[image]" . | wc -l``

I get 242 files --> a lot of improvement potential.

@settermjd fyi

mmattel:
**Update:**
Many of the image descriptions are set in ``documentation`` but not in ``docs``.
Seems that this could be somehow automated.

**Example:**
```
/var/www/owncloud/docs/modules/administration_manual/pages/appliance/Office.adoc
image:appliance/ucs/onlyoffice/001-ucs-portal.png[image]
```
```
/var/www/owncloud/documentation/admin_manual/appliance/Office.adoc
.. figure:: ../images/appliance/ucs/onlyoffice/001-ucs-portal.png
   :alt: UCS Portal
```

From the idea, we could use the same mechanism for orphaned images to identify all files in ``docs`` containing ``[image]`` and check in documentation for these files containing ``:alt:``. Maybe not a perfect match but a start to see how much the difference is.

mmattel:
Started with one file (Office.adoc) ...https://github.com/owncloud/docs/pull/184

PR #194, opened 2018-11-03
Fix available image :alt: (alternative) text

mmattel:
After thinking how to process, I wrote a program to identify missing image descriptions and (mostly) automate replacing the dummy description ``[image]`` with the correct one from the documentation repository.
This is the result + some manual post work as there were exeptions that had to be covered which would take too long to automate.

These descriptions would have gone lost if not migrated and would have missed on the future generated docs page.

Fixes:  https://github.com/owncloud/docs/issues/182

For a description what :alt: text is: https://moz.com/learn/seo/alt-text


modules/administration_manual/pages/appliance/Clamav.adoc
```diff
@@ -29,13 +29,13 @@ the checkboxes next to ``**clamav**'' (2), ``**clamav-freshclam**'', and
...
 
-image:appliance/ucs/clamav/install-clamav.png[image]
+image:appliance/ucs/clamav/install-clamav.png[UCS: Installing clamav, clamav-freshclam, and clamav-daemon.]
 
 After you do so, a confirmation dialog appears, as in the screenshot
 below, asking for confirmation to install the packages. Confirm the
 choice by again clicking ``**INSTALL**''.
 
-image:appliance/ucs/clamav/confirm-clamav-installation.png[image]
+image:appliance/ucs/clamav/confirm-clamav-installation.png[s2q]
```
tboerger: s2q?

mmattel: I did not qualify the content, this is the original from documentation.
Whatever this means, I migrated it.
I propose anyway to review a lot of content, but this would be a separate PR.

settermjd: As that's an alt tag, it'd make sense to make it more meaningful. I get, however, the script can't interpret these things on-the-fly.

mmattel: correct, I/we can change that to whatever we want to have written there, or even reset it to the default ``[image]``.
In case of changing, what text should be used?

PR #197, opened 2018-11-04
Improve getting-started.md

mmattel:
This is a small improvement.

The current ``Learn [the AsciiDoc file format]`` references to the same file as ``Learn [Antora's basics](./what-is-antora.md)`` which makes imho no sense.

Therefore I changed that line to a more relevant content, to get the syntax right, see links/references added.


settermjd:
Thanks for spotting it, @mmattel, and for making the change.

PR #199, opened 2018-11-05
Ignore the cache directory

settermjd:
No need to include it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 public/
 **/*.pdf
+cache/


tboerger:
```
public/
cache/
**/*.pdf
```

You don't need the `*` :)

settermjd:
@tboerger, I don't? I thought it was necessary to search/glob anywhere in the repo?

tboerger:
Not for gitignore.

settermjd:
Forget my earlier comment, I was, for some reason, referring to the pdf line, not the cache line. That and I misread your comment. .gitignore's now updated.

PR #201, opened 2018-11-05
Section Title out of Sequence

mmattel:
This PR fixes a warning:

``WARNING: file_xy.adoc: line 771: section title out of sequence: expected level 2, got level 3``

This happens when you have headers where headers jump over an expected level.

No Warning
``==``
``===``
``====``
Warning
``==``
``====``

The fix eliminates the warning by adding a header with the needed level.
There is no visibility break when you don´t do it, but fixing it quiets the antora docs build process.


settermjd:
I'd suggest that it's effectively a linter. Browser render engine's are so good these days that they work around HTML that isn't semantically correct. However, technically, mistakes such as these are still mistakes and should be removed.

Issue #200, opened 2018-11-05
https://owncloud.org/history is no longer a valid link


mmattel:
## What Type Of Content Change Is This?

- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [x] Bug Fix to Existing Content

## Which Manual Does This Relate To?

- [ ] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [x] General

## What Needs to be Documented?

Fix or remove link

## Why Should This Change Be Made?

When building docs, you get about 261 errors pointing to https://owncloud.org/history which is a dead link. According to @hurradieweltgehtunter, this link was present in the past but is no longer available.
The place to be fixed is in file: ``antora-ui-default/src/partials/footer-content.hbs``

@settermjd fyi

PR #202, opened 2018-11-05
Fix owncloud.org history and created a missing file

mmattel:
**Link corrections in the page footer:**
- ``owncloud.org/history`` is no longer available and there is no replacement for - line deleted
- other available links pointing to a new name have been corrected

**Missing file:**
- ``release_notes.adoc`` was missing
image
The file contains now a basic content which must be updated on new doc releases.

Fixes: https://github.com/owncloud/docs/issues/200 (https://owncloud.org/history is no longer a valid link)

**Update:**
After digging and drilling why the changes like ``history`` did not ave any effect, I found the solution.
If you change _any_ of the ``.hbs`` files, you also need to add/replace them in ``ui-bundle.zip``. Only then the changes take effect !

@settermjd @hurradieweltgehtunter


tboerger:
The ui bundle should be downloaded from the doc theme repo.

mmattel:
@tboerger, mind to chat on talk?

settermjd:
Just for clarity, the ui-bundle.zip isn't tracked in the repo, but one's available as a build artifact with both releases. To see the changes, you need to first rebuild the theme after you make any changes, using `gulp pack`, then you need to point the docs repo at the newly generated `ui-bundle.zip`, so that the changes are visible. I didn't track `ui-bundle.zip`, as it didn't make sense when it's a build artifact (at least it didn't make sense to me).

settermjd:
I think I need to turn [that last comment](https://github.com/owncloud/docs/pull/202#issuecomment-436309894) into a blog post.

PR #206, opened 2018-11-06
update readme.md

mmattel:
This is a small update of README.md to better reflect when to do changes in docs-ui.


README.md
```diff
@@ -18,7 +18,8 @@ To get started contributing to the documentation, please refer to [the Getting S
 
 To generate the documentation, whether in HTML or PDF format, please refer to [the Building the Documentation guide](./docs/build-the-docs.md).
 
-## Styling the Documentation
+## Common Content and Styling the Documentation
 
-If you find a bug in the look and feel of the ownCloud documentation or want to suggest an improvement, you can find all the information that you need in [the docs-ui repository](https://github.com/owncloud/docs-ui/blob/master/README.adoc).
+If you want to change eg. a footer, want to suggest an improvement of the layout or find a bug in the look and feel of the ownCloud documentation, you can find all the information that you need in [the docs-ui repository](https://github.com/owncloud/docs-ui/blob/master/README.adoc).
```
settermjd: How about the following instead?

> If you want to suggest an improvement to the layout of the ownCloud documentation or find a bug, all the information that you need is in [the docs-ui repository](https://github.com/owncloud/docs-ui/blob/master/README.adoc).
> Please be aware that changes made in `docs-ui` apply to the entire documentation.

mmattel: I did the rewrite in that way because the current writing highlights more bugs or features, but eg NOT that a footer is part of that. The lesson I learned from today was painful. Changed .hbs but no change in the result. by chance extracting the zip and saw that a change there made what was needed/wanted. If we reword that any change in docs-ui is for the entire documentation, we miss again the link that a footer is part of it. I want to link the readers brain to such issues. you know, I know now, but ask someone where to change the footer...

PR #208, opened 2018-11-07
Add a sample playbook file

settermjd:
[Antora](https://docs.antora.org/antora/)'s still a bit young, so it doesn't - **yet** - provide documentation in all forms necessary to answer the questions that every user could have (okay, no project does a perfectly comprehensive job, but I hope you get what I'm saying here). 

So this commit adds a sample [Playbook file](https://docs.antora.org/antora/1.0/playbook/#playbook-file-formats), one that is heavily commented, so that contributors can better understand what can be configured, what the sections do, and where to find more information.

**Note:** It is also usable when building the ownCloud docs in what Antora refers to as [Author Mode](https://docs.antora.org/antora/1.0/playbook/author-mode/).

### DIFF
--- a/site.sample.yml
+++ b/site.sample.yml
@@ -1,0 +1,70 @@
+#
+# This is an example Antora playbook configuration file. By default it's not
+# the one that's recommended in the documentation. It's included as a sample or
+# demonstration of what is possible, along with links to the relevant sections
+# of the documentation.
+#
+
+#
+# In this section, you can configure the core playbook configuration settings.
+# This section covers the site's title, base URL, and site start page. For more
+# information, see https://docs.antora.org/antora/1.0/playbook/configure-site/
+#
+site:
+  title: ownCloud Documentation
+
+content:
+  sources:
+  # This configuration enables what's referred to as "Author Mode". In this
+  # "mode", the local working copy is used to build the documentation. It
+  # doesn't refer to a specific branch, whether local or remote. This avoids
+  # generating only static assets, but no documentation. 
+  # For more information, see https://docs.antora.org/antora/1.0/playbook/author-mode/. 
+  # Normally, you would add standard content sources here. For more information
+  # about them, see https://docs.antora.org/antora/1.0/playbook/configure-content-sources/.
+  - url: ./
+    branches: HEAD
+  # Other components (modules) that you wish to include as part of this playbook.
+  - url: https://github.com/owncloud/android.git
+    branches:
+    - master-antora
+    start_path: docs/
+  - url: https://github.com/owncloud/ios.git
+    branches:
+    - master-antora
+    start_path: docs/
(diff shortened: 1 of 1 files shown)


site.local.yml
```diff
@@ -0,0 +1,70 @@
...
+
+#
+# In this section, you can configure the UI/theme for the documentation. This
+# covers the UI theme bundle to use, and whether to use a default page layout.
+# For more information, see
+# https://docs.antora.org/antora/1.0/playbook/configure-ui/
+#
+ui:
+  bundle:
+    url: https://github.com/owncloud/docs-ui/releases/download/1.0.0/ui-bundle.zip
```
mmattel: Why do we need ``releases`` instead using the directory ``bundle`` which exists and would be easy to un-gitignore (...gulp pack)? I may be wrong, but this approach implies a manual steep creating a release by someone after we have made a PR, commit, approval, merge process. Why a second process ?

site.local.yml
```diff
@@ -0,0 +1,70 @@
...
+  output_dir: assets
+
```
mmattel: If that gets included, I question myself, and that may be out of that PR (or not) which components (directory or files) are mandatory needed in the docs repro and which once are or can be included by the .yml file.
I am thinking of the docs/antora-ui-default/ directory. Tese things seem to be double present?

mmattel:
when finished, will we add that stuff to our working site.yml ?

site.local.yml
settermjd: It was where I was at at the time that I originally created the file. @tboerger's implementing a build pipeline for the docs. I'll change this after he's finished.

site.local.yml
settermjd: I don't follow, sorry.

site.local.yml
mmattel: Let us not forget to update this accordingly :smile:

mmattel:
``So this commit adds a sample Playbook file``...
Q: Why naming it ``site.local.ym`` and not ``site.sample.yml`` ?

site.local.yml
mmattel: Note: the directory ``docs/antora-ui-default/`` has been removed in the meanwhile by PR https://github.com/owncloud/docs/pull/209

settermjd:
@mmattel, the file's been renamed to `site.sample.yml`. I compared it to `site.yml` and don't see any differences worth migrating, except for the attributes, potentially. However, custom attributes aren't used, yet, in the docs. However, it might be handy to add the section, even if it's empty, as a starting point.

site.local.yml
settermjd: I'm planning to remove bundle.url for the production playbook file, as it should be specified on the command-line by the build tool as part of the build pipeline.

PR #220, opened 2018-11-09
Add a make target for running an Antora Xref validator

settermjd:
This is designed as another step for simplifying working with Antora projects. It uses a custom Antora build pipeline that makes use of a custom project generator, that can validate all Xrefs in a Playbook's
source files and output a report based on what it finds, if there are any invalid Xrefs found.

This is an initial concept. I'd like to integrate it with the ui-bundle.zip that's built, in the future, as part of the upcoming docs-ui build pipeline.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ STYLE         = owncloud
 BASEDIR       = $(shell pwd)
 APPVERSION    = 10.0.19
 BRANCH        = $(shell git rev-parse --verify HEAD)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -11,6 +11,7 @@ STYLE         = owncloud
 BASEDIR       = $(shell pwd)
 APPVERSION    = 10.0.19
 BRANCH        = $(shell git rev-parse --verify HEAD)
+UI_BUNDLE	  = https://github.com/owncloud/docs-ui/releases/download/1.1.0/ui-bundle.zip
```
mmattel: Why do you have teh ui-bundle with the old link instead the new minion link?

PR #224, opened 2018-11-13
Document WebDAV Search API

settermjd:
This PR documents two reports in the WebDAV Search API, file-filter and search-files. It fixes https://github.com/owncloud/core/pull/31873.

This PR makes extensive use of Antora's include directive, including both examples and partials. As a result, it may make the content a bit hard to review directly in GitHub. If you build the documentation locally, it will be much easier to review.


jvillafanez:
The search-files report will use the ownCloud's search API, with whatever search implementation is available. In particular, if search_elastic or search_lucene apps are installed, they'll be used. The report will remain the same, but the returned files might be different according to however those apps are configured. Not sure if it's worthy to comment it here.

For the rest, it looks good to me.

settermjd:
@jvillafanez, can you provide more context to your comment? I don't quite get where it fits in or what needs to be changed to accommodate it.

settermjd:
Ah, I understand now. Thanks for the clarification @jvillafanez. How is the search provider specified?

jvillafanez:
I guess the app itself takes care of the registration, so it's a matter of install the app the configure it properly. I don't think there is any additional thing to do.
So there is no need to specify the search provider.

settermjd:
Just specify the note about search options, based on the different search providers?

settermjd:
@jvillafanez, note added. Please let me know if you're okay with it.

modules/developer_manual/pages/webdav_api/search/search_files.adoc
```diff
@@ -7,7 +7,14 @@
 :method: REPORT
 :content_type: text/xml
 
-The search-files report allows for searching the available files in an ownCloud user's filesystem, based on a rudimentary filename pattern match.
+The `search-files` report search through the available files in an ownCloud user's filesystem, based on a rudimentary filename pattern match.
+
+By default, the report uses the default search provider to power the search functionality.
```
mmattel: ``uses the default`` --> ``uses ownCloud's default``
Should be the same as the sentence on line 14

PR #228, opened 2018-11-14
Remove environment-specific Antora playbook

settermjd:
As Playbook options can be passed on the command-line, such as the site URL, there's no point in going down the path of creating environment-specific playbooks. As such, the core Playbook's been renamed and the documentation that referred to it has been updated.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -29,7 +29,7 @@ pipeline:
     image: antora/antora:1.0.1
     pull: true
     commands:
-      - antora generate --cache-dir cache/ --ui-bundle-url ui-bundle.zip --redirect-facility netlify site.prod.yml
+      - antora generate --cache-dir cache/ --ui-bundle-url ui-bundle.zip --redirect-facility netlify site.yml
 
   cache-rebuild:
     image: plugins/s3-cache:1
--- a/docs/build-the-docs.md
+++ b/docs/build-the-docs.md
@@ -24,7 +24,8 @@ docker run -ti --rm \
       --cache-dir /antora/cache/ \
       --generator=./generate-site.js \
       --stacktrace \
-        site.prod.yml
+      --url <URL or path to your Antora instance> \
+      site.yml
 ```
 
 This command:
@@ -57,8 +58,8 @@ antora --clean --pull --quiet --silent \
     --redirect-facility static \
     --stacktrace \
     --ui-bundle-url https://github.com/owncloud/docs-ui/releases/download/1.1.0/ui-bundle.zip \
-    --url http://localhost:5000  \
-    site.prod.yml
+    --url <URL or path to your Antora instance> \
+    site.yml
 ```
 
 ### Update The Generated Search Index
--- a/site.yml
+++ b/site.yml
@@ -21,7 +21,7 @@ content:
 
 ui:
   bundle:
-    url: ui-bundle.zip
+    url: https://minio.owncloud.com/documentation/ui-bundle.zip
   output_dir: assets
 
 output:
--- a/site.yml
+++ b/site.yml
@@ -1,5 +1,6 @@
 site:
   title: ownCloud Documentation
+  url: https://owncloud-docs.netlify.com/ 
 
 content:
   sources:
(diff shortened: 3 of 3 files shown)


docs/build-the-docs.md
```diff
@@ -24,7 +24,8 @@ docker run -ti --rm \
       --cache-dir /antora/cache/ \
       --generator=./generate-site.js \
       --stacktrace \
-        site.prod.yml
+      --url <URL or path to your Antora instance> \
```
mmattel: I would write here:
``your Antora instance`` --> ``your Antora docs repro instance``
It may be that thre are more than one documentation repositories for other projects and this makes it more clear - imho.

.drone.yml
```diff
@@ -29,7 +29,7 @@ pipeline:
     image: antora/antora:1.0.1
     pull: true
     commands:
-      - antora generate --cache-dir cache/ --ui-bundle-url ui-bundle.zip --redirect-facility netlify site.prod.yml
+      - antora generate --cache-dir cache/ --ui-bundle-url ui-bundle.zip --redirect-facility netlify site.yml
```
tboerger: ui-bundle should be dropped as well, that should be part of the playbook.

docs/build-the-docs.md
tboerger: That should be: URL for the documentation

site.yml
```diff
@@ -1,6 +1,5 @@
 site:
   title: ownCloud Documentation
-  url: ${HOST}
```
tboerger: The default should be our production url

docs/build-the-docs.md
mmattel: I just looked at the antora documentation, https://docs.antora.org/antora/1.1/cli/
``--url <url>``
```
Base URL of the published site.
The URL should not include a trailing slash.
Overrides the site url value specified in the playbook
```
Means for me,
the playbook should contain as ``<url>`` --> ``https://docs.owncloud.org``
and should be overwritten when you build the docs by the url your local webserver uses to access the built documentation as this is for maintaining it.

site.yml
mmattel: yes, as I commented above

docs/build-the-docs.md
tboerger: This attribute defines the base URL, right. That's why this always have to be some local webserver URL. For production, this is anyway totally automated by drone.

settermjd: Thanks for the final point of clarity, @tboerger.

.drone.yml
settermjd: What are we setting this to? We didn't finalise this.

site.yml
settermjd: What is the production URL?

.drone.yml
tboerger: Look at my docs-ui pull request, the bundle gets uploaded to our Minio on every push to master. If you tag a git version it automatically builds the bundle and attaches it to a github release.

The only thing you got to do manually is writing the github release title and comment.

site.yml
tboerger: Currently it's https://owncloud-docs.netlify.com, in the end it will be https://doc.owncloud.com.

settermjd: Thanks. I'll change it to Netlify for now and then update it again later.

docs/build-the-docs.md
```diff
@@ -24,7 +24,7 @@ docker run -ti --rm \
       --cache-dir /antora/cache/ \
       --generator=./generate-site.js \
       --stacktrace \
-      --url <URL or path to your Antora instance> \
+      --url https://owncloud-docs.netlify.com/ \
```
tboerger: `--url <URL of your webserver for local testing>` should be better, the production URL is part of the `site.yml`

settermjd: Okay. I guess I misunderstood you previously.

settermjd: Fixed.

.drone.yml
tboerger: Still need to drop the bundle flag, and maybe even delete the bundle from this repo.

settermjd: I'll do that in a follow-up PR.

PR #229, opened 2018-11-14
Fixing more broken links

mmattel:
The next PR fixing broken links.

As usual:

- fixing broken links
- changing redirected links to the new location
- using link: in front of http
- fixing links with problematic URL (needs special attention)
- fixing (broken) references to the built site by using xref instead of http
- text fixes where found

There will be more PR´s with these kind of fixes (as you already hyve seen before) because I have to run the linkckecker each time again after merging. The resulting output file has to be parsed manually and there are so many entries to check...
But we are getting closer to an end :smile:

### DIFF
--- a/modules/developer_manual/pages/app/advanced/code_signing.adoc
+++ b/modules/developer_manual/pages/app/advanced/code_signing.adoc
@@ -45,7 +45,7 @@ release version branch in version.php to something else than ``stable`.
 === Is Code Signing Mandatory For Apps?
 
 Code signing is optional for all third-party applications. Applications
-with a tag of ``Official` on https://marketplace.owncloud.com/ require
+with a tag of ``Official` on link:https://marketplace.owncloud.com/ require
 code signing.
 
 [[technical-details]]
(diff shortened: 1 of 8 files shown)


modules/developer_manual/pages/app/advanced/code_signing.adoc
```diff
@@ -147,7 +147,7 @@ developers to share their private key.
 The following errors can be encountered when trying to verify a code
 signature. For information about how to get access to those results
 please refer to
-https://doc.owncloud.com/server/latest/admin_manual/issues/code_signing.html#fixing-invalid-code-integrity-messages[the
+xref:administration_manual:issues/code_signing.html#fixing-invalid-code-integrity-messages[the
```
settermjd: Nice one. That's one of my favourite Antora features.

modules/developer_manual/pages/general/security.adoc
```diff
@@ -153,7 +153,7 @@ because all `$_REQUEST` parameters are strings. However, if you expect
 text as a parameter, use
 link:http://php.net/manual/en/function.htmlspecialchars.php[PHP’s htmlspecialchars function] with 
 `ENT_QUOTES` or `strip_tags` to prevent
-link:https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)[Cross-site Scripting (XSS) attacks].
+link:++https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)++[Cross-site Scripting (XSS) attacks].
```
settermjd: What's with the surrounding `+`'s?

settermjd: Okay, I see the issue after experimenting with it locally. Strange that it should without the plus signs, it would cause the paragraph to italicise.

mmattel:
https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#links

PR #234, opened 2018-11-15
Redirect the existing documentation to the new documentation

settermjd:
This PR creates [a Netlify _redirects file](https://www.netlify.com/docs/redirects/) to redirect some content in the existing documentation to the new documentation so that there are no breaks in the user experience. It also creates a set of section index files, to make the process easier and provide "landing pages" so that the documentation makes that much more sense and is more professional.


tboerger:
Is the redirect facility flag of antora relevant for that?

settermjd:
I experimented with using the `page-redirect` attribute in combination with the `--redirects` option in Antora. However, it will only redirect from a page to another page. It doesn't support redirecting from a directory path to a page. When I attempted to do so, the result was something like `configuration/.adoc  configuration/index.adoc 301`. Given that, I decided to create a Netlify redirects file by hand.

mmattel:
If you look at #242 (More fixing links and other stuff)
redirecting here in file ``_redirects`` for ``/server/latest/`` is no longer necessary. 
All cleaned up by the above PR.
But (currently) true for ``ios`` and ``desktop``.

settermjd:
@mmattel, this redirect is necessary, as it redirects requests for the existing documentation to the new documentation when the new documentation is live. That way, there will be no break in functionality. Internal links won't solve this problem.

settermjd:
Thanks for the reminder. I didn't add it there automatically, as it would be overwritten in the future, if we need to create redirects using the `page-redirect` attribute and the `redirects` cli option. Would be good to update the Drone pipeline to prepend it to an existing file (if exists) or copy it to `public` if none exists. I'll have a go at getting that done.

tboerger:
> Thanks for the reminder. I didn't add it there automatically, as it would be overwritten in the future, if we need to create redirects using the `page-redirect` attribute and the `redirects` cli option. Would be good to update the Drone pipeline to prepend it to an existing file (if exists) or copy it to `public` if none exists. I'll have a go at getting that done.

Maybe we should just add another step to `.drone.yml` that is simply doing something like this?

```
cat _redirects >> public/_redirects
```

That way it will simply append the redirects if that file already exists.

tboerger:
BTW, we should do the same for headers:

## _headers

```
/*
  Content-Security-Policy: default-src 'self'
  X-Frame-Options: DENY
  X-Xss-Protection: 1; mode=block
  X-Content-Type-Options: nosniff
  Referrer-Policy: strict-origin-when-cross-origin
```

## script

```
cat _headers >> public/_headers
```

settermjd:
@tboerger, I just added a small script in 53ca392528445dd79451a1e0ef5c7ca44fc79b9c to do just that, with a little bit of defensive programming for good measure. Can you have a quick review, pls?

settermjd:
@tboerger, where does `_headers` come from?

settermjd:
@tboerger, I've updated [netlify.toml](https://github.com/owncloud/docs/pull/234/files#diff-5db06fd2327543bbb72119cd1e5761cf) and I believe that I've accounted for all manuals, regardless of version. It redirects the versions prior to the latest to the (yet to be created) `doc-old` subdomain that you mentioned in the #documentation channel earlier in the week. Can you proof it, to see if the rules are in the correct order.

settermjd:
@tboerger, after much trial and error, the redirect-map.conf file seems to be the correct approach, along with the following `location` directive, assuming redirect-map.conf's auto-included in NGINX's configuration:

```nginx
location / {
    if ( $redirect_uri ) {
        return 301 $redirect_uri;
    }
}
```

settermjd:
@tboerger, I've updated the redirect-map to point latest to 10.0

tboerger:
The index pages are totally fine, but I think we can drop the `redirect-map.conf` config.

.drone.yml
```diff
@@ -29,7 +29,7 @@ pipeline:
     image: antora/antora:1.0.1
     pull: true
     commands:
-      - antora generate --cache-dir cache/ --redirect-facility netlify site.yml
+      - antora generate --cache-dir cache/ --redirect-facility static site.yml
```
tboerger: This change should also be reverted, sorry

settermjd: You want netlify put back? Why? We're not using Netlify now.

settermjd:
@tboerger, I've remove redirect-map.conf and fixed the merge conflict. Just need feedback on your most recent question.

.drone.yml
tboerger: Master branch is using a make target, so just drop this change :)

PR #242, opened 2018-11-17
More fixing links and other stuff

mmattel:
There is the next PR fixing links with following changes

- write ``link:`` on all resolvable http(s) links
- fixing false used ``xref:`` instead of ``link:``
- fixing all links to internal pages using ``xref`` instead, mostly for 
doc.owncloud.com/server/admin_manual but also for the developer and user manual
(keeping links to old releases not having `latest` in the url as they are NOT touched !)
- fixing some tables for better readability
- some text fixes
- adding line breaks if there have been extra very long lines
- making lines shorter by removing a line break if the content reads better in the editor
- fixing ' ' --> ``
- ...

It seems that all links are now fixed, we have a big benefit beside proper rendering.
All links which are NOT preceeded by ``link:`` are non-resolvable links and therefore internal for eg examples. This also means that if you search for or want to differenciate resolvable links you can use the pattern ``:http``

In addition, all internal links pointing to a content inside docs are now 
``xref: ... .adoc`` and not ``link: ... .html``
No need to do any webserver redirect !

### DIFF
--- a/modules/ROOT/pages/index.adoc
+++ b/modules/ROOT/pages/index.adoc
@@ -3,7 +3,7 @@
 == Getting ownCloud
 
 You can download and install ownCloud on your own Linux server, try some prefab cloud or virtual machine images, or sign up for hosted ownCloud services. 
-See the xref:https://owncloud.org/install/[Get Started] page for more information.
+See the link:https://owncloud.org/install/[Get Started] page for more information.
 
 == Reporting Documentation Bugs & How to Contribute
 
@@ -12,7 +12,8 @@ Please refer to xref:how_to_contribute.adoc[the contributing guide] for instruct
 
 == Current Stable Server Release
 
-The _Administrator_, _User_, and _Developer_ manuals for the current stable release are always at https://doc.owncloud.org/#current-stable-server-release.
+The _Administrator_, _User_, and _Developer_ manuals for the current stable release are always at 
+link:https://doc.owncloud.org/#current-stable-server-release.
 
 == Latest Stable Release (version 10.0.10)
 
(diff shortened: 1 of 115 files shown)


modules/ROOT/pages/index.adoc
```diff
@@ -3,7 +3,7 @@
 == Getting ownCloud
 
 You can download and install ownCloud on your own Linux server, try some prefab cloud or virtual machine images, or sign up for hosted ownCloud services. 
-See the xref:https://owncloud.org/install/[Get Started] page for more information.
+See the link:https://owncloud.org/install/[Get Started] page for more information.
```
settermjd: I did some checking around this because its use was bugging me and found that prefixing external links with `link:` is only necessary "if the target is wrapped". This is according to @mojavelinux (the asciidoctor project lead, and Antora co-lead). Given that, I strongly suggest removing it, unless absolutely necessary.

modules/administration_manual/pages/appliance/Clamav.adoc
```diff
@@ -9,31 +9,29 @@ Appliance. It is composed of two parts:
 [[install-clamav-and-related-components]]
 == Install ClamAV and Related Components
 
-First, start the appliance and go to ``**System and domain settings**''.
+First, start the appliance and go to ``**System and domain settings**``.
```
settermjd: The text shouldn' be wrapped in double backticks, but with double-quotes, as the user's being directed to click on a specific section of the UI. Secondly, AsciiDoc only requires single backticks for inline code.

modules/ROOT/pages/index.adoc
mojavelinux: Correct. The link: prefix is only needed when the target is not a URI. That's because the URI protocol is an implicit macro (in other words, https: is recognized as a macro prefix of an implicit link).

mmattel: thanks for clarification :smile:

PR #247, opened 2018-11-19
Remove sudo www-data from acceptance test instructions

phil-davis:
Acceptance tests should be able to be run more independently of the system-under-test.
In particular, there should be no need to switch to being the Apache server user name.

See https://github.com/owncloud/documentation/pull/4472 for the same change in the documentation repo.

### DIFF
--- a/modules/developer_manual/pages/core/acceptance-tests.adoc
+++ b/modules/developer_manual/pages/core/acceptance-tests.adoc
@@ -243,8 +243,8 @@ Run a command like the following:
 
 [source,bash]
 ----
-sudo -u www-data make test-acceptance-api BEHAT_SUITE=apiTags
-sudo -u www-data make test-acceptance-cli BEHAT_SUITE=cliProvisioning
+make test-acceptance-api BEHAT_SUITE=apiTags
+make test-acceptance-cli BEHAT_SUITE=cliProvisioning
 ----
 
 === Running Acceptance Tests for a Feature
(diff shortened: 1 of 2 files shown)


modules/developer_manual/pages/core/acceptance-tests.adoc
```diff
@@ -254,8 +254,8 @@ Run a command like the following:
 
 [source,bash]
 ----
-sudo -u www-data make test-acceptance-api BEHAT_FEATURE=tests/acceptance/features/apiTags/createTags.feature
-sudo -u www-data make test-acceptance-cli BEHAT_FEATURE=tests/acceptance/features/cliProvisioning/addUser.feature
+make test-acceptance-api BEHAT_FEATURE=tests/acceptance/features/apiTags/createTags.feature
```
skshetry: Is this docs only for running on drone? Because I need to add `TEST_SERVER_URL` in the docs for tests to run. And, doing this works on every setups, doesn't it? 

EDIT: Pressed the wrong button

phil-davis: That should be described in sections above. There are about 100 ways to do it.
For example, when I open a terminal in the morning, I do:
```
export TEST_SERVER_URL=http://172.17.0.1:8080
export TEST_SERVER_FED_URL=http://172.17.0.1:8180
export BROWSER=chrome
```
and then I do not have to think about putting any of these before or after ``make`` commands.

So the "style" of using environment variables it is up to the developer. The documentation presents one "style".

PR #323, opened 2018-11-19
Add netlify headers file

settermjd:


tboerger:
Does Antora support something like the `static` folder of Hugo? Or do we really need to copy the file?

settermjd:
Every module can have an `assets/` directory, where you store static assets, specific to _that_ module. However, given Antora's modular nature, which Hugo doesn't follow, they're not a 1-1 concept. The reason that I created that script, is because antora's command-line tooling generates a _redirects file in `public/`, so I'm trying to be consistent there, and merging the file into the deployed doc root.

tboerger:
Just asking because on hugo you can put files into `static/` like https://github.com/kleister/kleister-docs/tree/master/static and they will be automatically copied to the resulting public folder, IDK if Antora got something comparable for additional files within the document root?

settermjd:
To the best of my knowledge, Antora doesn't have such a feature. But even if it did, how would it work if in the future, additional redirects would need to be created, that could be created by Antora's cli option? There would have to be two separate files that would be merged. So I see this as being required anyway, at least until search engines cache the new page locations.

tboerger:
But we are currently taking about headers, not redirects

settermjd:
don't mind me, was clearly a mile away when I wrote that.

settermjd:
@tboerger, I refactored _headers to netlify.toml in 3660097 to avoid having multiple Netlify configuration files and to keep the root directory cleaner.

PR #415, opened 2018-11-27
fixing failed to parse formatted text (image alt text)

mmattel:
This PR fixes an issue I found when looking in the drone result creating pdf´s, example:

Failed to parse formatted text: [Protecting files tagged with

The issue are the double quote inside the alternative text eg "Confidential".
This is not a problem when the docs are built to html because double quotes inside an alt text are rewritten to `&quot;` (see https://github.com/asciidoctor/asciidoctor/issues/2061) but in building the pdf´s seems they are.

The pdf generator seems to take the complete img src string and parses back `&quot;` to double quotes which generate this error...

Created an issue at: https://github.com/asciidoctor/asciidoctor-pdf/issues/977 (HTML entitiy &quot inside block causing 'Failed to parse formatted text')

### DIFF
--- a/modules/administration_manual/pages/configuration/server/antivirus_configuration.adoc
+++ b/modules/administration_manual/pages/configuration/server/antivirus_configuration.adoc
@@ -182,11 +182,11 @@ The Antivirus App for Files will show one of three warnings if it is
 either misconfigured, or ClamAV is not available. You can see an example
 of all three below.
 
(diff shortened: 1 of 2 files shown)


settermjd:
From what I understand, asciidoctor-pdf, while still supported, may not be for that much longer. So it's likely best to ask that question in the forums too: https://gitter.im/asciidoctor/asciidoctor.

PR #417, opened 2018-11-28
Correct wrong recommendation on keepAlive settings

micbar:
The recommended defaults are wrong. This has created severe downtimes at some instances. If the keepAliveTimeout is set to 100, Apache gets blocked by too many inactive connections.


settermjd:
ping @micbar

Issue #437, opened 2018-12-05
Replace email example in occ user:setting


PVince81:
See https://github.com/owncloud/core/issues/33711.

Since one should not set email vie occ user:setting but view user:modify as per https://doc.owncloud.org/server/10.0/admin_manual/configuration/server/occ_command.html#modify-user-details, we should modify the example in question and use another preference value.

mmattel:
Closed via #438

PR #438, opened 2018-12-06
update occ user setting command

mmattel:
References https://github.com/owncloud/docs/issues/437 (Replace email example in occ user:setting)

**Reason**
When you set an email address with occ like described at https://doc.owncloud.org/server/10.0/admin_manual/configuration/server/occ_command.html#setting-a-setting, this address is stored in oc_preferences as user preferences whereas changing the password in the UI only changes the values in oc_accounts (source).

At every login the password is synced from the preferences to the accounts table (called here and performed there) and thus changes from the UI are lost.

### DIFF
--- a/modules/administration_manual/pages/configuration/server/occ_command.adoc
+++ b/modules/administration_manual/pages/configuration/server/occ_command.adoc
@@ -2661,7 +2661,7 @@ will exit with 1. Only applicable on get.
 | `--error-if-not-exists`           | Checks whether the setting exists before deleting it.
 |====
 
-[IMPORTANT:] In case you want to change the email address, do not use `user:settings` but the `user:modify` command.
+[IMPORTANT:] In case you want to change the email address, use the `user:modify` command.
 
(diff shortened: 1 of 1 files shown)


modules/administration_manual/pages/configuration/server/occ_command.adoc
```diff
@@ -2663,18 +2661,20 @@ will exit with 1. Only applicable on get.
 | `--error-if-not-exists`           | Checks whether the setting exists before deleting it.
 |====
 
-Here's an example of how you would set the email address of the user `layla`.
+[IMPORTANT:] In case you want to change the email address, do not use `user:settings` but the `user:modify` command.
```
settermjd: I'd end this sentence as: "use the `user:modify` command, not the `user:settings` command." This approach should require less cognitive load on the user.

mmattel: fixed

PR #443, opened 2018-12-07
Just exit 0 if no validation errors

tboerger:
Without this fix the validator exists with code 1 even if there are not any validation issues.


mmattel:
Thanks a lot for adding that 👍 
I will add that suggestion to https://gitlab.com/antora/docs.antora.org/tree/master/generator so all can profit of this.

mmattel:
Just added https://gitlab.com/antora/docs.antora.org/merge_requests/18

PR #454, opened 2018-12-11
Update best-practices.md

mmattel:
The admonition example had square brackets but you must not use them.
Would not work otherwise. Was a copy/paste mistake.

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -204,7 +204,7 @@ Asciidoctor provides five admonition style labels: `NOTE`, `TIP`, `IMPORTANT`, `
 
 An `admonition` is written in following example style: `[<label>:]` Text
 
-Example: `[NOTE:]` If you have a large ownCloud installation and have shell access, you should ...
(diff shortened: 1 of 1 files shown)


docs/best-practices.md
```diff
@@ -204,7 +204,7 @@ Asciidoctor provides five admonition style labels: `NOTE`, `TIP`, `IMPORTANT`, `
 
 An `admonition` is written in following example style: `[<label>:]` Text
 
-Example: `[NOTE:]` If you have a large ownCloud installation and have shell access, you should ...
+Example: `NOTE:` If you have a large ownCloud installation and have shell access, you should ...
```
settermjd: Markdown doesn't support admonitions. What about just bolding the text?

mmattel: This is not Markdown !
We write BP for AsciiDoc 
This example is for AsciiDoc presented in Markdown :smile:

settermjd: There is such a thing?

mmattel: Yes, believe me

PR #472, opened 2018-12-20
Convert rST double backticks to adoc single backticks

settermjd:
There were a number of instances where rST-style double-backticks still persisted. These are unnecessary in AsciiDoc.

### DIFF
--- a/modules/administration_manual/pages/configuration/files/external_storage/dropbox.adoc
+++ b/modules/administration_manual/pages/configuration/files/external_storage/dropbox.adoc
@@ -58,7 +58,7 @@ However, the internal Dropbox functionality should be removed in ownCloud 10.0.4
 
 Then, you need to provide a name for the folder in the "Folder name" field, and a "client key" and "client secret", located in the
 "Configuration" column.
-The client key and client secret values are the "App key" and ``App secret" fields which you saw earlier in your Dropbox app’s configuration settings page.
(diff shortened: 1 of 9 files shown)


modules/administration_manual/pages/configuration/files/external_storage/dropbox.adoc
```diff
@@ -58,7 +58,7 @@ However, the internal Dropbox functionality should be removed in ownCloud 10.0.4
 
 Then, you need to provide a name for the folder in the "Folder name" field, and a "client key" and "client secret", located in the
 "Configuration" column.
-The client key and client secret values are the "App key" and ``App secret" fields which you saw earlier in your Dropbox app’s configuration settings page.
+The client key and client secret values are the "App key" and "App secret" fields which you saw earlier in your Dropbox app’s configuration settings page.
```
mmattel: shouldn´t hat be single backticks instead of quotes ?

mmattel:
You have to reset config.sample back to master and remove it from this PR.
Then do the changes in core and create an own PR and let confg-to-docs run (always for both files).
config.sample files in docs will always be overwritten, you need to change the source in core...

modules/administration_manual/pages/configuration/files/external_storage/dropbox.adoc
settermjd: Not in this case, as it has a trailing/wrapping double-quote.

settermjd:
@mmattel, config.sample.php's been removed from this PR.

PR #475, opened 2018-12-20
Integrate Algolia search

settermjd:
After much research, using a combination of Algolia and their docsearch-scraper Docker image to scrape the documentation and maintain the search index seems to be the best approach. Given that, this PR:

- Removes all references to lunr and antora-lunr (which was how things worked before)
- Updates the documentation to reflect the new strategy
- Adds a new Drone pipeline item

@tboerger, I need help to finish the Drone pipeline item please, as I'm sure it's not complete, and I need to get the secrets added to the configuration, which I can't do on my own.


mmattel:
It would be great if there would be a make command to update the search index like we have make html.

settermjd:
@tboerger, I've made some further changes to the pipeline item in .drone.yml and added in the relevant secrets. Just waiting to see if the build succeeds this time (_or at least does better than before_).

mmattel:
Docker tells on update the search index task:
```
Virtualenv location: /root/.local/share/virtualenvs/src-x3FbBUl6
Creating a Pipfile for this project...
/root/.local/share/virtualenvs/src-x3FbBUl6/bin/python: No module named src
```

.drone.yml
```diff
@@ -84,6 +84,25 @@ pipeline:
...
+  update-search-index:
+    image: algolia/docsearch-scraper
+    entrypoint: ""
+    pull: true
+    secrets: [ 
+      algolia_app_id, algolia_config, 
+      algolia_index_name, algolia_search_api_key, 
+      algolia_write_api_key 
+    ]
+    environment:
```
tboerger: Drop most of the environment block, secrets are injected as environment variables.

.drone.yml
```diff
@@ -84,6 +84,25 @@ pipeline:
...
+      - ALGOLIA_APP_ID=${ALGOLIA_APP_ID}
+      - ALGOLIA_API_KEY=${ALGOLIA_SEARCH_API_KEY}
+      - APPLICATION_ID=${ALGOLIA_APP_ID}
+      - API_KEY=${ALGOLIA_WRITE_API_KEY}
+      - ALGOLIA_INDEX_NAME=${ALGOLIA_INDEX_NAME}
+      - CONFIG=${ALGOLIA_CONFIG}
```
tboerger: Keep this var, but directly assign the value, e.g. `CONFIG=algolia-config.json`. BTW, we should move the config to `search/algolia.json` :)

.drone.yml
```diff
@@ -84,6 +84,25 @@ pipeline:
       local: false
       event: [ push ]
 
+  update-search-index:
+    image: algolia/docsearch-scraper
+    entrypoint: ""
```
tboerger: Drop that, Drone anyway overwrites the entrypoint.

.drone.yml
```diff
@@ -84,6 +84,25 @@ pipeline:
       local: false
       event: [ push ]
 
+  update-search-index:
+    image: algolia/docsearch-scraper
```
tboerger: Always use specific tags of used images, even if there is only `latest` available be more specific.

.drone.yml
```diff
@@ -84,6 +84,25 @@ pipeline:
...
+    environment:
+      - ALGOLIA_APP_ID=${ALGOLIA_APP_ID}
+      - ALGOLIA_API_KEY=${ALGOLIA_SEARCH_API_KEY}
+      - APPLICATION_ID=${ALGOLIA_APP_ID}
+      - API_KEY=${ALGOLIA_WRITE_API_KEY}
+      - ALGOLIA_INDEX_NAME=${ALGOLIA_INDEX_NAME}
+      - CONFIG=${ALGOLIA_CONFIG}
+    commands:
+      - pipenv run python -m src.index
+
```
tboerger: You should also add a `when` block, this should only be done on `push` event.

.drone.yml
```diff
@@ -92,3 +111,5 @@ pipeline:
       local: false
       event: [ push ]
       status: [ success, failure ]
+
```
tboerger: Drop the additional blank lines?

.drone.yml
```diff
@@ -84,6 +84,25 @@ pipeline:
...
+    ]
+    environment:
+      - ALGOLIA_APP_ID=${ALGOLIA_APP_ID}
+      - ALGOLIA_API_KEY=${ALGOLIA_SEARCH_API_KEY}
+      - APPLICATION_ID=${ALGOLIA_APP_ID}
+      - API_KEY=${ALGOLIA_WRITE_API_KEY}
+      - ALGOLIA_INDEX_NAME=${ALGOLIA_INDEX_NAME}
+      - CONFIG=${ALGOLIA_CONFIG}
+    commands:
+      - pipenv run python -m src.index
```
tboerger: Why does it need this pipenv stuff? Maybe we got to build our own Docker image for that...

settermjd:
Yeah, I'm still working on that.

.drone.yml
```diff
@@ -84,6 +84,25 @@ pipeline:
       local: false
       event: [ push ]
 
+  update-search-index:
+    image: algolia/docsearch-scraper
+    entrypoint: ""
+    pull: true
+    secrets: [
```
tboerger: Please use one or another approach for the formatting of array values, only one of these:

```
secrets: [ a, b, c ]
```

```
secrets:
- a
- b
- c
```

.drone.yml
settermjd: Okay, I'll do a bit of reading to work out how to do that.

.drone.yml
settermjd: It seemed to be what was required, as I understood it at the time.

.drone.yml
settermjd: Just trying to make it look neat and tidy.

.drone.yml
settermjd: I dropped the two that seemed okay to drop.

tboerger: The others like `ALGOLIA_API_KEY`, `APPLICATION_ID` and `API_KEY` should also be dropped from environment, just define them as secrets, if it's a secret at all.

.drone.yml
settermjd: From re-reading https://community.algolia.com/docsearch/run-your-own.html#running-your-config, the pipeline only needs the docsearch tool available, a Docker image isn't required, along with the config file.

.drone.yml
settermjd: They're not really secrets. Thanks for the suggestion.

.drone.yml
settermjd: Just made a further update to .drone.yml. How close is it now?

PR #481, opened 2018-12-21
Rename install make target to setup

settermjd:
After a discussion with @PVince81 "setup" is much more intuitive.


mmattel:
You also have to fix https://github.com/owncloud/docs/blob/master/docs/install-antora.md#install-the-antora-tools-locally accordingly

PR #491, opened 2019-01-07
Add a local playbook

settermjd:
This PR provides a local playbook, that will build the three core manuals, incorporating changes made in the local (GIT) working directory. It also updates the docs so that users know that this option is available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ help:
 	@echo "  clean        to clean the build directory of any leftover artifacts from the previous build."
 	@echo "  validate     to validate all xref links of all manuals defined within configuration."
 	@echo "  html         to generate the HTML versions of all manuals defined within configuration."
+	@echo "  html-local   to generate the HTML versions of all manuals defined within configuration from the local working directory."
 	@echo "  pdf          to generate the PDF versions of the administration, developer, and user manuals."
 	@echo "  check-prose  to lint English prose to support developers"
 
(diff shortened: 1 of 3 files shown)


mmattel:
We have now `site.yml`and `site.local.yml` —> great,
But `make html` still points to `site.yml` which is imho counteracting the great `make` idea.
Why not having two make commands, one for the external site and one for local.
That would not break things and eases handling and you can generate the site as it looks local and on the web.

settermjd:
Thanks for the further feedback, @mmattel. However, after reading further through [the GNU Make documentation](https://www.gnu.org/software/make/manual/make.html#Environment) (_thanks for the inspiration/motivation to do so_), there's no need to create multiple make targets. 

All that needs to be done is to set the `PLAYBOOK` variable appropriately (it defaults to `site.yml`) when calling `make html`; for example `make html PLAYBOOK=site.local.yml`. 

I'll update the documentation to reflect this.

settermjd:
@mmattel, the `make` section's been updated. Can you re-read it and share your feedback, pls?

mmattel:
Two comments:
great that make can define the playbook in the command, but if we look at daily life, we 99,99% would run make html (or any short command) to create a site to check local changes. The command where you add a quite long parameter is not that handy.
In addidtion running `make` alone will give you that help text:
``html         to generate the HTML versions of all manuals defined within configuration.``
Will you remember the long possible parameter versus a own make command for local and external?
For easy handling, I would prefer two make commands, one for local and one for external.

settermjd:
@mmattel, the Makefile now includes a new `html-local` target, which does what you suggest, and `docs/build-the-docs.md` has also been updated to reflect the new target as well.

PR #509, opened 2019-01-14
Document static tagging functionality

settermjd:
This relates to owncloud/enterprise#2748, by documenting the ability to create static tags, that are only editable by admins but can be assigned and unassigned by whitelisted group members.

This change:

1. Converts the previous formatting of the tag types into a description list to provide them with greater emphasis
2. Better emphasises that the Collaborative Tag Management app needs to be installed
3. Reflows some text as one sentence per line, so that it's easier to edit and maintain.


settermjd:
@PVince81, do you want the documentation to be more detailed? If so, I'm more than happy to do that. I've only not done it, as the current documentation's not that detailed, and I didn't want to go overboard, unnecessarily. Please advise.

PVince81:
@pmaier1 ^

PR #513, opened 2019-01-15
Fix list item index error

mmattel:
This PR fixes
- An error when building the site with make html because of list indexing starting with the same number
Now the numbering and indent is correct and no error is shown.
- Some line breakes for better source readability (no rendering impact)
- Fix link to config sample file


settermjd:
I approved this a little too quickly, as I'm not a fan of line-breaks for readability.

PR #516, opened 2019-01-15
Clean up the SMB/CIFS documentation

settermjd:
As part of making the changes required for #411, the SMB/CIFS documentation needed to be cleaned up. This commit makes the required changes.

Issue #411, opened 2018-11-26
Document min samba version for using windows network drive listener and smb protocol


patrickjahns:
## What Type Of Content Change Is This?

- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [x] Bug Fix to Existing Content

## Which Manual Does This Relate To?

- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual

## What Needs to be Documented?
In light of our blogpost https://owncloud.com/windows-network-drive-integration-benefits-from-samba/ the documentation of the windows network drive listener and the correlation with the samba protocol versions should be documented.

When the windows network drive listener is used, and the target file server requires the usage of the protocol SMB2+, we require at least Samba 4.7.8 ( https://www.samba.org/samba/history/samba-4.7.8.html // BUG13382 ) - or Samba 4.8.1 ( https://www.samba.org/samba/history/samba-4.8.1.html // BUG13382 ) on the ownCloud server.

With any Samba version lower then the above mentioned, the windows network drive listener from ownCloud would only work on samba protocol version 1

## Why Should This Change Be Made?
Keep the documentation up-to-date and avoid known errors for windows network drive listener deployments

settermjd:
Closing as the implementing PR's been merged.

patrickjahns:
@settermjd - reopening as #517 is still not merged

cdamken:
The information is only in: https://doc.owncloud.com/server/admin_manual/configuration/files/external_storage/smb.html
but this relevant issue is for WND:
https://doc.owncloud.com/server/admin_manual/enterprise/external_storage/windows-network-drive_configuration.html
Here is the whole information missing 😢 

@settermjd please add a link or add this information in the WND information.

settermjd:
Closing, as the related PRs (#860, #976) have been merged

PR #517, opened 2019-01-15
Document min samba version for using wnd listener

settermjd:
*Note*: Don't merge before #516 has been merged.


modules/administration_manual/pages/configuration/files/external_storage/smb.adoc
```diff
@@ -3,9 +3,18 @@
...
+:samba-481-url: https://www.samba.org/samba/history/samba-4.8.1.html 
+:smb2-url: https://en.wikipedia.org/wiki/Server_Message_Block#SMB_2.0
 
 ownCloud can connect to Windows file servers, and other SMB-compatible (Samba) servers, by using the {smbcifs-url}[SMB/CIFS] backend.
 
+[IMPORTANT]
+====
+ownCloud requires at least {samba-478-url}[Samba 4.7.8] or {samba-481-url}[Samba 4.8.1] on the ownCloud server, when the Windows Network Drive Listener is used and the remote Windows/Samba file server requires at least {smb2-url}[SMB 2.0].
+The Windows Network Drive Listener only supports version 1 of the SMB protocol with earlier Samba versions. 
+====
```
mmattel: This sentence sounds for me the following way - maybe I am wrong:
For the owncloud instance we need samba version greater than 4.7.x, for the target host SMB version must be minimum SMB2.0. WND listener on the other hand only supports SMB1. So host must be higher than listener - hmmm.

settermjd: I can't vouch for the differing versions of the listener, Samba, and the SMB protocol, but your text gave me reason to have a think. I like it mostly as it is, but I agree it can be updated so that it's clearer. I'll do that momentarily, and would appreciate your, further, feedback.

settermjd: Change made.

mmattel: That sounds strange because first you require a minimum samba version but then you exclude wnd listener in the xref line from it (with earlier Samba versions). Proposal: we need to contact a knowledged colleague to clarify that, else it will be a pain/guess only...

settermjd: Agreed. Just sent a message to the relevant channel.

mmattel: Any update ?

settermjd: Nothing yet.

patrickjahns:
Just pulling out https://github.com/owncloud/docs/issues/411 - as this is how it started.

To clarify:
A Samba **server** (usually in ownCloud context a microsoft share/server is used) can enforce min/maximum version of protocols to be used when a client talks to him. 
In light of the wannacry attack, microsoft has patched their windows server to only allow SMB2 protocol by default ( as SMB1 is insecure ).

The ownCloud windows network drive listener utilizes a the SMB notification feature. This has been working all good with SMB1 in conjunction with most samba versions. However - when the minimum protocol a server accepts is SMB2, we require samba 4.7.8+ ( 4.8+ etc. ) to be able to properly work, as prior versions of samba had a bug that would cause this feature not to work at all

So it boils down - if the  samba server ( windows server ) allows SMB1 as protocol - any samba is sufficient. If it requires SMB2+ - we need a more recent version

Please let me know if this clarifies the situation for you

patrickjahns:
@settermjd @mmattel  - anything missing here? Just got another request from a customer which has issues with 4.7.6 smb - would be great to have this in the documentation deployed

settermjd:
@patrickjahns will get this out, today!

PR #526, opened 2019-01-16
Remove SWIFT from the List of usable Storages

haukman:
SWIFT was still listed as usable backend storage - removed it.

### DIFF
--- a/modules/administration_manual/pages/installation/deployment_considerations.adoc
+++ b/modules/administration_manual/pages/installation/deployment_considerations.adoc
@@ -123,8 +123,7 @@ supported with ownCloud Enterprise Edition only.
 
 While many customers are starting with NFS, sooner or later that
 requires scale-out storage. Currently the options are GPFS or GlusterFS,
-or an object store protocol like S3 (supported in Enterprise Edition
-only). S3 also allows access to Ceph Storage.
+or an object store protocol like S3. S3 also allows access to Ceph Storage.
 
 [[session-storage]]
 == Session Storage
--- a/modules/administration_manual/pages/installation/deployment_considerations.adoc
+++ b/modules/administration_manual/pages/installation/deployment_considerations.adoc
@@ -124,7 +124,7 @@ supported with ownCloud Enterprise Edition only.
 While many customers are starting with NFS, sooner or later that
 requires scale-out storage. Currently the options are GPFS or GlusterFS,
 or an object store protocol like S3 (supported in Enterprise Edition
-only) or Swift. S3 also allows access to Ceph Storage.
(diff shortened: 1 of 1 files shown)


modules/administration_manual/pages/installation/deployment_considerations.adoc
```diff
@@ -124,7 +124,7 @@ supported with ownCloud Enterprise Edition only.
 While many customers are starting with NFS, sooner or later that
 requires scale-out storage. Currently the options are GPFS or GlusterFS,
 or an object store protocol like S3 (supported in Enterprise Edition
-only) or Swift. S3 also allows access to Ceph Storage.
+only). S3 also allows access to Ceph Storage.
```
mmattel: The text: `(supported in Enterprise Edition only)` needs to be removed because S3 is now for all editions.
To check see: https://marketplace.owncloud.com/search/s3

cdamken: I think it's correct that way, in the first line we specified "customers", the community tar does not include those apps.

mmattel: @cdamken this is correct that the apps are not included in the tarball, but you can download it from marketplace without any restriction to EE only. Therefore EE only is wrong. It was like that in the past, but has changed. This needs to be reflected by removing that wrong statement.

haukman: grafik

But the "old" S3 App ist still Enterprise - so it´s still kind of true :)

haukman:
> Great to see, @haukman. However, there are a few other references to Swift in the documentation. Are you planning to remove them as well?
> 
> They're in:
> 
>     * modules/administration_manual/pages/configuration/server/occ_command.adoc, lines 1263 and 1450.
> 
>     * modules/administration_manual/pages/configuration/files/external_storage/openstack.adoc, lines 3 & 4.
> 
>     * modules/administration_manual/pages/installation/apps_supported.adoc, line 26
> 
>     * modules/user_manual/pages/files/access_webdav.adoc, line 336

Yes, will do!

modules/administration_manual/pages/installation/deployment_considerations.adoc
settermjd: So, what's the resolution on this one?

mmattel: The old S3 is depriciated in favor the new one, is in the process to be phased out and not longer maintained...

settermjd:
@haukman, any further progress on this one?

PR #532, opened 2019-01-18
Better clarify statement about code signing being optional

settermjd:
As the statement was, it could be misleading. So this commit updates it to make a distinction between apps that are to be uploaded to the Marketplace, and those that aren't (_just used directly in ownCloud, via manual installation_).

### DIFF
--- a/modules/developer_manual/pages/app/advanced/code_signing.adoc
+++ b/modules/developer_manual/pages/app/advanced/code_signing.adoc
@@ -44,9 +44,8 @@ release version branch in version.php to something else than `stable`.
 [[is-code-signing-mandatory-for-apps]]
 === Is Code Signing Mandatory For Apps?
 
-Code signing is optional for all third-party applications. Applications
-with a tag of `Official` on https://marketplace.owncloud.com/ require
-code signing.
+If you intend to upload your app to the Marketplace, yes, code signing _is_ mandatory.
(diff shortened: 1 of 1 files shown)


modules/developer_manual/pages/app/advanced/code_signing.adoc
```diff
@@ -44,9 +44,10 @@ release version branch in version.php to something else than `stable`.
 [[is-code-signing-mandatory-for-apps]]
 === Is Code Signing Mandatory For Apps?
 
-Code signing is optional for all third-party applications. Applications
-with a tag of `Official` on https://marketplace.owncloud.com/ require
-code signing.
+If you intend to upload your app to the Marketplace, yes, code signing _is_ mandatory.
+If the app will only be installed directly in an ownCloud installation, then code signing is _optional_, for all third-party applications. 
+
+IMPORTANT: Applications with a tag of `Official` on https://marketplace.owncloud.com/ require code signing.
```
IljaN: @settermjd Remove/reword this line because there is no concept of tags at the marketplace at the moment. But I think owncloud can display official tag in the Apps view (not sure)

settermjd: Thanks for the feedback, @IljaN. As there's no concept of tags at the marketplace at the moment, then it makes sense to remove it. I'll get that done.

settermjd:
@IljaN, the change has been made. Any other changes required?

PR #538, opened 2019-01-21
Update the GitHub new issue template

settermjd:
This PR reverts the previous changes to the GitHub new issue template, as they made the template less helpful to those who are first-time contributors to the documentation. Secondly, it adds more manuals to the manuals list and adds a "Where should this be documented section".

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -17,6 +17,18 @@ To make it possible for us to help you, please fill out below information carefu
 - [ ] Admin Manual
 - [ ] Developer Manual
 - [ ] User Manual
+- [ ] Android
+- [ ] iOS
+- [ ] Branded Clients
+- [ ] Desktop Client
+- [ ] Other
+
+## Where Does This Need To Be Documented?
+<!--
+If this is a new content addition, please provide the path(s)
+to the relevant file(s), if known, where the new information needs
+to be added.
+-->
 
 ## What Needs to be Documented?
 <!--
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,14 +1,10 @@
-## What Needs to be Documented?
+<!--
+Thanks for wanting to contribute to the ownCloud documentation!
 
+However, before reporting any issues, please make sure that you're using the latest available version of master.
 
-## Why Should This Change Be Made?
-
-
-## Where Should This Change Be Made?
-<!-- GitHub page and Docs Page -->
-
-
-<!-- Create the issue, check those check boxes afterwards -->
+To make it possible for us to help you, please fill out below information carefully.
+-->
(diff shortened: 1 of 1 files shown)


voroyam:
why are the check boxes on the top?

I have no idea what the right format is to check them correctly, nor the average user.

That's why I put them at the bottom, so users can fill in the info, create the issue, and then click on the check boxes later on. That's way more comfortable and intuitive than guessing and trying the right combination of "x" and "space".

voroyam:
Please provide path(s) to file(s) can be confusing.

Users can interpret this as "I have found this in the admin manual section"
Or "this is in the LDAP section"
Have fun asking the user what he meant and searching for the document.

I think what I wrote was sufficient "GitHub link and Documentation Link"
2 links should be enough to get an overview where to write what.

voroyam:


Who do you expect to read all this without forgetting what they wanted to add to the docs?

Example - I need to document one information to the occ command about ecryption.

If I read all what you wrote I have forgotten all that I had in mind.

Also had you watched people who are not doc people writing issues?

They don't read anything, they glance at the things that are there and only write what they read in a few seconds. If the text is more than 1 line long it does not get read.

.github/issue_template.md
```diff
@@ -1,14 +1,10 @@
-## What Needs to be Documented?
-
-
-## Why Should This Change Be Made?
+<!--
+Thanks for wanting to contribute to the ownCloud documentation!
 
+However, before reporting any issues, please make sure that you're using the latest available version of master.
```
mmattel: `latest available version of master.`--> `latest available documentation version of master.`

mmattel:
> Please provide path(s) to file(s) can be confusing.

It is not, because you have "Edit this Page" on the web pointing clearly to the file in question.
(Except new file addition where you have to tell where you want to show it but then you use the same machanism as above)

> Also had you watched people who are not doc people writing issues? They don't read anything...

If they do not read anything, they will not write anything. And non-readers do not add value to the documentation. Just to open a bin and tell - throw your stuff into creates more work and confuses much more. Look at the great quality docs has now and is continuing to improve. This has been achieved by clear approaches and not by non-readers...

It is a quite bad idea to create a throw-in-bin-one-fits-all with the lowest level.
That will decrease quality dramatically.

Just my 2c

voroyam:
Okay, then my contribution approach is too low level for this new docs. 

What I wanted was to lessen the amount of work you and @settermjd have to do. That is only possible if we include more people in the contribution process. To do that you have to make the entry point as easy as possible. The first contribution station to the docs is opening an issue. If this is already to confusing - they won't go any further.

I don't want to ruin your new doc. So I will be hands off of this project.

PR #541, opened 2019-01-21
Document the inability to explicitly end OAuth2 web sessions

### LINKED ISSUE #534: Document OAuth2 Limitations regarding web sessions.

- [x] New Content Addition
- [ ] Old Content Deprecation

- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual

Limitations regarding web sessions. It's not possible to end user sessions (web sessions) with OAuth.

## Why Should This Change Be Made?

A customer wanted to manage user sessions, and found no explicid documentation about the limitation of OAuth2 regarding web sessions.

------------

Why is here no "where does this need to be documented?" section like in my idea for the template.

Now if I write just what in the current template people will ask "where do you think this should be documented instead of knowing it because it is in the template...

https://doc.owncloud.com/server/administration_manual/configuration/server/security/oauth2.html#limitations

https://github.com/owncloud/docs/blob/master/modules/administration_manual/pages/configuration/server/security/oauth2.adoc

settermjd:
This fixes #534.


voroyam:
hm. I am having trouble formulating my thoughts  on this matter.

Maybe someone here can help.

Here is what OAuth2 can't do - end specific user sessions for a client. You can end all user sessions on a specific type of client asfaik ( desktop, iOS, Android ) but not a single one as an admin.

You can't end web Sessions with OAuth2.

I would like to have these 2 thinks in the documentation in case a customer would think that if it's not explicitly mentioned as a non feature then it can be done.

settermjd:
@IljaN, can you provide further input on this topic, pls?

voroyam:
for some reason the tests were failing. So I took the content of this PR and created a new one:
https://github.com/owncloud/docs/pull/1507

voroyam:
can we close this @settermjd ?

PR #545, opened 2019-01-23
Update the issue template to make it simpler

settermjd:
This PR primarily aims to ensure that if someone is in a hurry, they can write what they need to as quickly as possible, and not have to fill out unnecessary forms. Thanks @voroyam.

It also:

1. Reorders the existing template items, top-loading them from the most to the least important.
2. Marks some items as optional.
3. Marks some items as used only by the docs team, not for people creating issues.

💁‍♂️ I hope this balances the requirements for people who are in a hurry, and those who have more time.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,19 +1,44 @@
+<!-- Short on time? Just write what you’re thinking. -->
+
+
+
+<!-- Have more time? Please answer out as many of the questions below as you can. -->
+## What Needs to be Documented?
 <!--
-Thanks for wanting to contribute to the ownCloud documentation!
-
-However, before reporting any issues, please make sure that you're using the latest available version of master.
-
-To make it possible for us to help you, please fill out below information carefully.
+Describe, in detail, what changes are required.
+However, don’t make it a Ph.D. thesis or "War and Peace".
 -->
-## What Type Of Content Change Is This?
-<!-- Please choose all that apply. -->
+
+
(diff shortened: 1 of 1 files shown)


voroyam:
Well, I appreciate the olive branch you are trying to hand me here.

In my opinion - to work an the docs efficiently you need the answer to 3 questions: what, where, why.

What needs to be changed so I can change it.
Where do I need to change it so I don't have to search for it and waste time.
Why are those changes necessary and later on - how can I then justify making those changes.

I understand that this is kinda robotic approach to the whole contribution process, but it makes it easier in my opinion to get started.

That's why I don't think the why and where are optional. Or should be optional. I watched Chris make an issue, and had to tell him that in the where section a GH and Docs link should go. It's not obvious to people that we need a where section if we already have a "what manual goes this change in to" section. From their POV they already told us that it should go in the admin manual, what else do you want?

Also if things are optional - people don't fill them out, because their time is ( or they thing it is ) valuable. 

Thant's why I don't think the where and why sections should be optional.

settermjd:
@voroyam, it's a little bit of an olive branch, and a little bit of integrating the points you made as best as possible to build the best new issue template. I discussed them with @PVince81, and we decided to reorder the questions to put the most important ones first, and to label some questions as optional, so that people wouldn't feel compelled to answer them - if their time was short. 

You're welcome to contribute further changes to the PR.

PR #547, opened 2019-01-23
Remove "Hidden users" information

voroyam:
"will be hidden users"

There are no hidden users in ownCloud.

Wrong information. Not reproducible.


voroyam:
> This statement is really questionable...
> Thanks for catching it 👍

I stumbled upon it when looking for a way to implement a customer request. I though I had the solution right there. Boy was I disappointed :D

PR #549, opened 2019-01-24
Move Algolia config directory

settermjd:
This PR moves the Algolia config directory so that it's in the location that `.drone.yml` expects. 

💁‍♂️ I might otherwise have changed the Drone configuration file to point to the previous path, but the path that it uses keeps the directory structure more intuitive and tidier.


settermjd:
@tboerger, can you review pls?

settermjd:
That was your idea, I thought.

tboerger:
> That was your idea, I thought.

It wasn't my idea to put it into search/algolia/. I wanted to put it into algolia/. You were just getting the paths wrong ;)

PR #551, opened 2019-01-24
Clarify where to create the transifex config

PVince81:
See PR title

### DIFF
--- a/modules/developer_manual/pages/app/advanced/l10n.adoc
+++ b/modules/developer_manual/pages/app/advanced/l10n.adoc
@@ -187,7 +187,7 @@ app.
 [[how-to-automatically-sync-translations]]
 == How to automatically sync translations
 
-1) Create a initial Transifex config within the app repository:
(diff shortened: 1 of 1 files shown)


modules/developer_manual/pages/app/advanced/l10n.adoc
```diff
@@ -187,7 +187,7 @@ app.
 [[how-to-automatically-sync-translations]]
 == How to automatically sync translations
 
-1) Create a initial Transifex config within the app repository:
+1) Create a initial Transifex config within the app repository under "l10n/.tx/config":
```
settermjd: The sentence should start: "Create an initial".

tboerger: Shouldn't the path be an inline code block for proper highlighting instead of quotes?

settermjd: Either way's fine, but an inline code block would be that much clearer, yes.

PR #557, opened 2019-01-28
Revise update-search drone task

settermjd:
This change should fix the update script so that it won't break at the end of the process, when it's asked whether it wants to update the `nb_hits` value in the config file. How to deal with the change in value with respect to git, that I'm not sure about, yet.


tboerger:
I'm already working a proper fix as part of a different branch.

PR #558, opened 2019-01-28
Add version secret in Drone build-pdfs task

settermjd:
As far as I can see, while Makefile uses a default value for `VERSION`, its not overridden as it's not set as a secret in `.drone.yml`. This commit rectifies that situation.


tboerger:
Why should the version be a secret? That should be part of the repo and not externally set.

settermjd:
It seemed simpler to do it this way, than having a commit each time. However, on thinking about it, the logic makes sense.

PR #559, opened 2019-01-29
Set the revision number and date correctly

settermjd:
[According to the documentation](https://asciidoctor.org/docs/user-manual/#revision-number-date-and-remark) the revision number and date were being set from incorrect attributes. Based on that, this PR has been created to ensure that the PDF configuration files follow the format correctly.

This relates to:

- owncloud/administration-internal#101
- owncloud/enterprise#2528

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ STYLES_DIR ?= resources/themes
 CACHE_DIR ?= cache
 
 STYLE ?= owncloud
-VERSION ?= 10.0.10
+VERSION ?= 10.1.0
 REVDATE ?= "$(shell date +'%B %d, %Y')"
 REDIRECTS ?= static
 PLAYBOOK ?= site.yml
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,8 @@ STYLES_DIR ?= resources/themes
 CACHE_DIR ?= cache
 
 STYLE ?= owncloud
-VERSION ?= 10.1.0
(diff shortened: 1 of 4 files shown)


PVince81:
conflicts ? @settermjd

settermjd:
Looking at that now.

Makefile
```diff
@@ -9,7 +9,8 @@ STYLES_DIR ?= resources/themes
 CACHE_DIR ?= cache
 
 STYLE ?= owncloud
-VERSION ?= 10.1.0
+VERSION ?= 10.0.10
```
mmattel: FYI in just merged https://github.com/owncloud/docs/pull/560 you set
`VERSION ?= 10.1.0`
which you would now "revert" back in case you merge this PR...

settermjd: That was merged by @PVince81 for the next release. I'm currently discussing, in #documentation, about the branching strategy, going forward. Thanks for commenting on this.

tboerger:
This is the master branch, so it's 10.1. We should branch off stable10 or 10.0.x for that.

settermjd:
As per https://github.com/owncloud/docs/issues/563, I'm thinking of a simple versioning strategy, just plain semver, such as 10.1.0. Can you add your feedback to that issue, @tboerger, pls?

Makefile
settermjd: @mmattel, reverted.

PR #567, opened 2019-01-30
Limit the amount of document indexing

settermjd:
This change seeks to limit the amount of document indexing, by restricting it to only happen when the master branch is pushed to, and the build is a success.

💁‍♂️ I'm still trying to figure out how to limit it to just once a day; though, one step at a time. @tboerger/@pjahns, any suggestions?

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -103,7 +103,9 @@ pipeline:
       - cd /root && pipenv run python -m src.index
     when:
         local: false
+        branch: [ master ]
         event: [ push ]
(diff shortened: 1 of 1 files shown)


tboerger:
The current changes doesn't change anything, the master branch is already limited for the whole pipeline, and the step gets anyway only executed on success.

settermjd:
Good to know. I couldn't make that out from the Drone docs. Any suggestions about how to limit indexing?

PR #569, opened 2019-01-31
Update the Drone build-html pipeline configuration

settermjd:
This PR adds two missing secrets from the `build-html` pipeline configuration, and removes the `algolia_app_id` secret mapping. There's no reference to the target `algolia_application_id`.

💁‍♂️: from what I've been able to piece together, if a secret is to be used, it has to be explicit in the configuration; it can't be implicit. Please let me know if I'm wrong in this assumption.


mmattel:
Question,
does this have an doc impact to https://github.com/owncloud/docs/blob/master/docs/build-the-docs.md  ?

settermjd:
I'm not considering that until the configuration's worked out. At this stage, I don't know if it is. Need to somehow figure out how to setup Drone locally so that I can test it.

tboerger:
I'm already working a proper fix as part of a different branch.

PR #571, opened 2019-02-01
fix for jsonlint executable in make

mmattel:
This is a fix for the jsonlint executable.
Even the package is called jsonlint, the command is called `jsonlin-php` (not `jsonlint`)
Pls see: 
https://askubuntu.com/questions/966331/jsonlint-command-not-found-although-jsonlint-is-installed
https://packages.ubuntu.com/xenial/jsonlint
Experienced after installing `sudo apt-get install jsonlint`


settermjd:
It's not called that for me. I'll have to check further.

mmattel:
Seems that there are two possibilities.
Hmmm, on Ubuntu it is called `jsonlint-php` on others seems to be `jsonlint`.
So we need to check both variants in make and use the available one and if none is available report the error.

mmattel:
@settermjd I adopted the jsonlint call to be flexible for Ubuntu (`jsonlint-php`) and other OS (`jsonlint`).
I have checked it for Ubuntu and it works, pls check for your OS.

PR #579, opened 2019-02-05
Update links to latest docs

settermjd:
This change updates the home page references to the latest and previous version of the docs, and updates the component descriptor to ensure that the previous version of the docs is both built and accessible from the UI.

💁‍♂️ Do not merge until #578 is merged!

### DIFF
--- a/modules/ROOT/pages/index.adoc
+++ b/modules/ROOT/pages/index.adoc
@@ -1,4 +1,5 @@
 = ownCloud Documentation Overview
+:docs-base-url: https://doc.owncloud.com/server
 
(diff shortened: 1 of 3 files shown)


tboerger:
I'm against branching every version, imho it's only maintainable to do 10.0.x and 10.1.x branches.

settermjd:
So long as we settle on something, I'm fine with it.

tboerger:
The client is branching as 2.5 and so on. Imho it's sufficient to do the same for the core docs

settermjd:
Okay, I agree with you.

PR #585, opened 2019-02-06
updating the sessions doc

voroyam:
This should be an update for the sessions documentation.

A new screenshot and a rework of the whole page.


mmattel:
@voroyam looking at the location of the file, there is already a picture named `settings_sessions.png`
which content seems the one you want to have updated with this PR.
But your file is named `settings_sessions_pic.png`.
Which means that the current PR does not update teh file but creates an orphan.
Can you check please?

voroyam:
@mmattel my goal was to create a new file with new name, and reference this file, delete the current ones as they are outdated.

voroyam:
But I screwed up and now I need a new screenshot ( new new ) because on the one I uploaded information was missing.

mmattel:
Overwrite the current file with a new one.
No need to do the double work.

PR #587, opened 2019-02-07
Update playbook and component descriptor for version 10.0

settermjd:
This will ensure that the playbook and component descriptor for version 10.0 of the docs only builds the latest docs from the 10.0 branch and no longer include master.

### DIFF
--- a/modules/ROOT/pages/index.adoc
+++ b/modules/ROOT/pages/index.adoc
@@ -1,4 +1,5 @@
 = ownCloud Documentation Overview
+:docs-base-url: https://doc.owncloud.com
 
 == Getting ownCloud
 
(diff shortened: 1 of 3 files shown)


antora.yml
```diff
@@ -1,6 +1,6 @@
 name: server
 title: ownCloud Server
-version: master
+version: '10.0'
```
tboerger: Sorry, I have pushed that fix directly to the branch by accident, please drop this change.

settermjd: I thought 10.1 is the one that you pushed.

tboerger: i had pushed both before i checked that i forgot to create a separate branch ;)

settermjd: np. thanks.

settermjd: fixed.

settermjd:
If you're happy with this, please give it a tick and I'll merge it.

tboerger:
Please revert the site config changes, I will include that within the build process finishing. Than I will merge the PR.

PR #588, opened 2019-02-07
Remove Piwik integration

settermjd:


mmattel:
Isnt it necessary to remove Piwik support from docs-ui too?
Pls see merged PR: https://github.com/owncloud/docs-ui/pull/27

settermjd:
It's not, at least not for the time being, as it won't break anything, as it won't be activated. I need to raise the issue in https://gitter.im/antora/users to see how to do this. I'm just taking it out for the time being so that builds don't break.

settermjd:
As always, @mmattel, thanks for spotting this.

PR #589, opened 2019-02-07
Update docs for version 10.1

settermjd:
This PR:

- Updates the docs to build version 10.1 and move version 10.0 to be the previous version.
- Update the links to the three core manuals to reference the new latest version and to move the others back a version.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -1,6 +1,6 @@
 name: server
 title: ownCloud Server
-version: master
+version: 10.1
 nav:
 - modules/ROOT/nav.adoc
 - modules/administration_manual/nav.adoc
--- a/modules/ROOT/pages/index.adoc
+++ b/modules/ROOT/pages/index.adoc
@@ -27,12 +27,12 @@ The _Administrator_, _User_, and _Developer_ manuals for the current stable rele
 
 == Previous Stable Release (version 9.1)
 
-* https://doc.owncloud.com/server/9.1/admin_manual/[Administration Manual]
-(https://doc.owncloud.com/server/9.1/ownCloud_Server_Administration_Manual.pdf[Download PDF])
-* https://doc.owncloud.com/server/9.1/developer_manual/[Developer Manual]
-(https://doc.owncloud.com/server/9.1/ownCloudDeveloperManual.pdf[Download PDF])
-* https://doc.owncloud.com/server/9.1/user_manual/[User Manual]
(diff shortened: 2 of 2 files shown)


antora.yml
```diff
@@ -1,6 +1,6 @@
 name: server
 title: ownCloud Server
-version: master
+version: '10.1'
```
tboerger: Sorry, I have pushed that fix directly to the branch by accident, please drop this change.

settermjd: I just pulled the latest version of 10.1 and don't see it there.

settermjd: Ah, I see it now, in the conflict with `master`.

tboerger: No need to quote it... Just drop the change ;)

settermjd: I'm keeping the `version` value in this branch, as it should work as a reminder to quote the value supplied. I set it to `10.0` for the 10.0 branch and the fraction was truncated.

site.local.yml
```diff
@@ -5,7 +5,9 @@ site:
 content:
   sources:
   - url: .
-    branches: HEAD
```
tboerger: site.local.yml have been dropped from master branch.

site.yml
```diff
@@ -8,7 +8,8 @@ content:
   sources:
   - url: https://github.com/owncloud/docs.git
     branches: 
-    - master
```
tboerger: This branch should only build `HEAD` and url `.`

settermjd: Why so? I get that HEAD would make sense, but `.`? Why would that value be required?

antora.yml
tboerger: Where? I haven't seen such a behavior locally.

site.yml
tboerger: Because it should only build the changes from the current branch.

antora.yml
settermjd: I saw it happen when I was creating the patch/PR for 10.0 branch.

site.yml
settermjd: But as this will be merged into the 10.1 branch, it needs to reference the preceding branch, no? Or have I misunderstood you?

site.local.yml
settermjd: Thanks for reminding me of that. I've just removed that file from the PR.

tboerger:
Please revert the site config changes and the antora.yml changes, I will include that within the build process finishing. Than I will merge the PR.

settermjd:
@tboerger, done.

PR #593, opened 2019-02-12
first draft of a single guide how to install ownCloud for a specific distro

haukman:
First draft to improve the manual installation - open for discussions. 

Idea is to have one guide for each distro which can be used to install ownCloud on a single server from the start to the end, without the need to lookup things on google etc.

### DIFF
--- a/modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
+++ b/modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
@@ -1,0 +1,163 @@
+= Manual Installation on Linux - Ubuntu 16.04
+:nginx-app-tracing-url: https://www.nginx.com/blog/application-tracing-nginx-plus/
+
+== Introduction
+This guide describes how to install the required dependencies, the database and ownCloud itself in a
+single server setup.
+
+
+[[install-the-required-packages]]
+== Prepare your Server
+
+*Necessary steps to prepare the installation of ownCloud*
+
+Before we can start to install ownCloud, we need to make sure that everything needed
+is available.
+
+Steps to take:
+
+* Install MariaDB Database
+* Install Redis Server
+* Install Apache and the necessary dependencies
+* Configure Apache
+* Download and install ownCloud 
+
+
+=== Install MariaDB Database
+
+Perform the following steps to install MariaDB and setting up a Database for ownCloud.
+
+....
+apt update
+apt install -y mariadb-server-10.0
+....
+After this has finish, create the database and a database user for ownCloud to use:
+
+....
+mysql -u root
+
+MariaDB [(none)]> create database owncloud;
+MariaDB [(none)]> GRANT USAGE ON owncloud.* TO owncloud@localhost IDENTIFIED BY 'password';
+MariaDB [(none)]> GRANT ALL PRIVILEGES ON owncloud.* TO owncloud@localhost IDENTIFIED BY 'password';
+MariaDB [(none)]> flush privileges;
+MariaDB [(none)]> exit
+....
+
+=== Install Redis Server
+Redis in needed for Transactional File Locking in ownCloud. Perform the following step to install it:
+....
+apt install -y redis-server
+....
+
+=== Install Apache and the necessary dependencies
+The following steps will install the Apache Webserver and the dependencies needed for ownCloud:
+....
+apt install -y \
+apache2 \
+libapache2-mod-php \
+php-gd \
+php-json \
+php-mysql \
+php-curl \
+php-intl \
+php-mcrypt \
+php-imagick \
+php-zip \
+php-xml \
+php-mbstring \
+php-soap \
+php-ldap \
+php-apcu \
+php-redis \
+smbclient \
+php-smbclient \
+gettext-base && \
+a2enmod ssl rewrite headers env dir mime ldap && \
+service apache2 restart
+....
+
+=== Configure Apache
+We will create a new apache config file for ownCloud:
+....
+vim /etc/apache2/sites-available/owncloud.conf
+....
+Please add the following to the file and adjust parameters like SERVER-NAME and
+the paths to the certificates.
+....
+<VirtualHost *:80>
+    ServerAdmin webmaster@localhost
+    ServerName  SERVER_NAME
+    Redirect permanent / https://SERVER_NAME/
+</VirtualHost>
+
+<IfModule mod_ssl.c>
+    <VirtualHost _default_:443>
+        ServerAdmin webmaster@localhost
+        ServerName  SERVER_NAME
+        DocumentRoot /var/www/owncloud/
+        <Directory />
+                Options FollowSymLinks
+                AllowOverride All
+        </Directory>
+
+        <IfModule mod_headers.c>
+                Header always set Strict-Transport-Security "max-age=15768000;  includeSubDomains; preload"
+        </IfModule>
+        ErrorLog ${APACHE_LOG_DIR}/error.log
+
+        # Possible values include: debug, info, notice, warn, error, crit,
+        # alert, emerg.
+        LogLevel warn
+
+        SSLEngine on
+        SSLProtocol all -SSLv2 -SSLv3
+        SSLCipherSuite EDH+CAMELLIA:EDH+aRSA:EECDH+aRSA+AESGCM:EECDH+aRSA+SHA25 6:EECDH:+CAMELLIA128:+AES128:+SSLv3:!aNULL:!eNULL:!LOW:!3DES:!MD5:!EXP:  !PSK:!DSS:!RC4:!SEED:!IDEA:!ECDSA:kEDH:CAMELLIA128-SHA:AES128-SHA
+        SSLHonorCipherOrder on
+        SSLCompression off
+        SSLCertificateFile      /path/to/certificate
+        SSLCertificateKeyFile   /path/to/key
+        CustomLog /var/log/apache2/ssl_request_log "%t %h %{SSL_PROTOCOL}x  %{SSL_CIPHER}x \"%r\" %b"
+
+        BrowserMatch "MSIE [2-6]" nokeepalive ssl-unclean-shutdown  downgrade-1.0 force-response-1.0
+        BrowserMatch "MSIE [17-9]" ssl-unclean-shutdown
+
+    </VirtualHost>
+</IfModule>
+....
+
+After that enable the new config and reload Apache:
+....
+a2ensite owncloud
+service apache2 reload
+....
+
+== Download and install ownCloud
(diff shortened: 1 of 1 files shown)


haukman:
@michaelstingl @voroyam @patrickjahns

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+=== Change permissions
+This step is necessary to make sure that the apache user can read and access the
+ownCloud directory:
+....
+find /var/www/owncloud/ -type f -print0 | xargs -0 chmod 0640
+find /var/www/owncloud/ -type d -print0 | xargs -0 chmod 0750
+chown -R www-data:www-data /var/www/owncloud
+....
+=== Install ownCloud
```
voroyam: This can be done with a occ command, so no need for pictures.

like this for example

```
occ maintenance:install \
    --database "mysql" --database-name "owncloud" \
    --database-user "admin" --database-pass "admin" \
    --admin-user "admin" --admin-pass "admin"
```

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
```diff
@@ -0,0 +1,163 @@
...
+tar xjf owncloud-10.1.0.tar.bz2
+....
+
+=== Change permissions
+This step is necessary to make sure that the apache user can read and access the
+ownCloud directory:
+....
+find /var/www/owncloud/ -type f -print0 | xargs -0 chmod 0640
+find /var/www/owncloud/ -type d -print0 | xargs -0 chmod 0750
+chown -R www-data:www-data /var/www/owncloud
```
voroyam: This command is sufficient, no need for the 2 above.

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
```diff
@@ -0,0 +1,163 @@
...
+....
+
+== Download and install ownCloud
+After we have successfully prepared the Server, we can continue to install ownCloud.
+
+=== Download ownCloud
+Create/enter the directory that was configured for ownCloud in your vhost configuration,
+download and unpack the ownCloud tarfile.
+....
+mkdir /var/www/owncloud
```
voroyam: no need to make a folder "owncloud"

If you extract the tar ball, it's in the folder owncloud. Just extract it to `/var/www ` or extract it to whatever and move it to `/var/www`

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
```diff
@@ -0,0 +1,163 @@
...
+....
+a2ensite owncloud
+service apache2 reload
+....
+
+== Download and install ownCloud
+After we have successfully prepared the Server, we can continue to install ownCloud.
+
+=== Download ownCloud
+Create/enter the directory that was configured for ownCloud in your vhost configuration,
```
voroyam: no need for that

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
```diff
@@ -0,0 +1,163 @@
...
+        BrowserMatch "MSIE [2-6]" nokeepalive ssl-unclean-shutdown  downgrade-1.0 force-response-1.0
+        BrowserMatch "MSIE [17-9]" ssl-unclean-shutdown
+
+    </VirtualHost>
+</IfModule>
+....
+
+After that enable the new config and reload Apache:
+....
+a2ensite owncloud
```
voroyam: there are several other modules we have to enable:

a2enmod headers
a2enmod env
a2enmod dir
a2enmod mime
a2enmod unique_id

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
```diff
@@ -0,0 +1,163 @@
...
+service apache2 restart
+....
+
+=== Configure Apache
+We will create a new apache config file for ownCloud:
+....
+vim /etc/apache2/sites-available/owncloud.conf
+....
+Please add the following to the file and adjust parameters like SERVER-NAME and
+the paths to the certificates.
```
voroyam: what if the user does not have certificates? I suggest we have 2 files, one for http only and one for https

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
```diff
@@ -0,0 +1,163 @@
...
+php-smbclient \
+gettext-base && \
+a2enmod ssl rewrite headers env dir mime ldap && \
+service apache2 restart
+....
+
+=== Configure Apache
+We will create a new apache config file for ownCloud:
+....
+vim /etc/apache2/sites-available/owncloud.conf
```
voroyam: I would write "open it with the editor of your choice.  because not all people know how to use vim. ( or most importantly - how to get out of vim)

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
```diff
@@ -0,0 +1,163 @@
...
+....
+apt update
+apt install -y mariadb-server-10.0
+....
+After this has finish, create the database and a database user for ownCloud to use:
+
+....
+mysql -u root
+
+MariaDB [(none)]> create database owncloud;
```
voroyam: this can be done with one line.

```
mysql -u root -e "CREATE DATABASE IF NOT EXISTS owncloud; \
GRANT ALL PRIVILEGES ON owncloud.* TO admin@localhost IDENTIFIED BY 'admin'";
```

mmattel:
Regarding setting the correct permissions, please take a look at:
https://doc.owncloud.com/server/administration_manual/installation/installation_wizard.html#post-installation-steps. You will find a script that does all the job for you AND you can use the same script in case of updating the owncloud code, which eases life dramatically. You could quite easily xref to it.

mmattel:
Regarding Ubuntu 16.04
As far I remember, but I may be wrong, the PHP version 16.04 provides does not match the prerequisite and you have to manually install a matching PHP version via ppa.
This is not the case for 18.04 where PHP matchhes, but that OS version needs some other PHP module prerequisites, have different namings or are at least packed differently or obsolete at all.
Please also consider this as the official PHP release available for 16.04 had end of support last year...

voroyam:
> As far I remember, but I may be wrong, the PHP version 16.04 provides does not match the prerequisite and you have to manually install a matching PHP version via ppa.

I think you have to add the version number before the package number

mmattel:
Needs to be proved, because I remember it just not available and therefore not selectable.

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
settermjd: It'd be good to strongly encourage users to use HTTPS, but, not all users may see the value in that option nor want to put in the effort to implement it. So, I agree that both an HTTP and HTTPS option should be made available.

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
settermjd: Well said. Perhaps we should also provide a list of suggested editors, such as VIM, ed, pico, nano, etc.

settermjd:
I agree that we should always place emphasis on the latest supported version, as an implicit way of encouraging users to use the latest version. However, we need to make sure that we provide documentation that shows users how to work with all the versions that we support. Otherwise, we're not truly supporting those versions.

settermjd:
I like the concept of this addition. However, it'd be great if we could instead have one large page (built by including a number of small guides) that uses a tabbed interface that allows users to pick the instructions for the OS distribution and version that they've chosen. I think there was a custom addition to Antora that already provides for a tabbed interface. I'll have to check in the community discussion if that's the case and where to find it.

haukman:
@mmattel @settermjd: This first PR was just an initial draft that we used to start discussing a new possible structure for the install docs. Please don´t invest to much time for now in technical details etc. 

Meeting Minutes:
- Add a disclaimer on top to explain that this is just a fast forward install (no hardening)
- Three installation options: Docker, Manual Install (prepare your server, install ownCloud), Appliance

Next steps:
- Hauke prepares structure overview
- Invite Martin & Matthew to next call and discuss

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
```diff
@@ -0,0 +1,163 @@
+= Manual Installation on Linux - Ubuntu 16.04
+:nginx-app-tracing-url: https://www.nginx.com/blog/application-tracing-nginx-plus/
+
+== Introduction
+This guide describes how to install the required dependencies, the database and ownCloud itself in a
+single server setup.
+
+
+[[install-the-required-packages]]
```
settermjd: You don't need to create custom anchors, as anchors are [auto-generated for headers](https://asciidoctor.org/docs/user-manual/#auto-generated-ids).

modules/administration_manual/pages/installation/manual_installation_ubuntu_16_04.adoc
```diff
@@ -0,0 +1,163 @@
...
+=== Install Redis Server
+Redis in needed for Transactional File Locking in ownCloud. Perform the following step to install it:
+....
+apt install -y redis-server
+....
+
+=== Install Apache and the necessary dependencies
+The following steps will install the Apache Webserver and the dependencies needed for ownCloud:
+....
+apt install -y \
```
settermjd: I strongly suggest formatting this command example differently, so that it's more compact, yet still readable. For example:

```
apt install -y \
    apache2 gettext-base && libapache2-mod-php \
    php-apcu php-curl php-gd php-imagick \
    php-intl php-json php-ldap php-mbstring \
    php-mcrypt php-mysql php-redis php-smbclient \
    php-soap php-xml php-zip smbclient && \
a2enmod ssl rewrite headers env dir mime ldap && \
service apache2 restart
```

voroyam:
@haukman I am seeing some mixed messaging here

> start discussing

> don´t invest to much time for now in technical details

If you throw a first incomplete untested draft in a room full of experts they gonna make suggestions how to improve it. Even if you don't want the suggestions now, they are important for later, when you want to make something real out of the draft.

haukman:
@voroyam, that was the reason why I asked you if it would be a good idea to create a PR as we are still not sure about the final structure ;) 

Before we start to add/create new content we need to define a structure that makes sense and work on the content afterward. I will close this PR and take this work offline to don´t waste anybody's time.

PR #616, opened 2019-02-19
fix shell used in makefile

mmattel:
This is a small fix for the makefile.
Running Ubuntu 18.04,

following error gets reported using the original makefile:
```
make help
/bin/sh: 2: Syntax error: "(" unexpected
Makefile:49: recipe for target 'help' failed
make: *** [help] Error 2
```
With the fix:
```
make help
target                         help
------                         ----
help                           Print a basic help about the targets
setup                          Install Antora's command tools locally
clean                          Remove build artifacts from output dir
pdf                            Generate PDF version of the manuals
pdf-user                       Generate PDF version of the user manual
pdf-admin                      Generate PDF version of the administration manual
pdf-developer                  Generate PDF version of the developer manual
validate-xml                   Validate all XML files
validate-php                   Validate all PHP files
validate-yaml                  Validate all YAML files
validate-json                  Validate all JSON files
```

It seems that conditional setting the shell if not already set to bash does not work on Ubuntu 18.04.


tboerger:
That got to be changed on the branches 10.0 and 10.1 beside the other repos like client, ios, android, branded_clients as well as all are using nearly the same makefile.

PR #623, opened 2019-02-19
Change system tag internal name to collaborative

davitol:
Change system tag internal name to collaborative, due to system tags is the internal code name.
. Collaborative tags is the user facing name.

### DIFF
--- a/modules/user_manual/pages/files/webgui/tagging.adoc
+++ b/modules/user_manual/pages/files/webgui/tagging.adoc
@@ -13,12 +13,12 @@ there is no need to use single or double-quotes. Type as many words as
 you want for the tag name. When you press the return key your tag will
 be saved.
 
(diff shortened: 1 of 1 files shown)


modules/user_manual/pages/files/webgui/tagging.adoc
```diff
@@ -13,12 +13,12 @@ there is no need to use single or double-quotes. Type as many words as
 you want for the tag name. When you press the return key your tag will
 be saved.
 
-NOTE: All tags are system tags, so they are shared by all users on your ownCloud server.
+NOTE: All tags are collaborative tags, so they are shared by all users on your ownCloud server.
```
settermjd: I recommend changing "_, so they_" to "_, this means that they_", which makes it that much clearer.

PR #630, opened 2019-02-20
Document the documentation branching workflow

settermjd:
As the new branching workflow has come into effect with the core manuals, then it needs to be added to the contributor/maintainer docs. This PR attempts to do so, succinctly, yet in as detailed as is necessary.


mmattel:
I would add the following for clarification, text to be created, but from the idea:

...so it has three branches: `master`, `10.1`, and  `10.0`. --> which one is used for what?

10.0 (former release) ... bugfix only --> pushed to web
10.1 (current release) ... bugfix only --> pushed to web
master ... not pushed to web but will become the next current release eg 10.2

So the intention what to commit where becomes more clear.

tboerger:
Master gets pushed to the web but it's a moving target as it's not tied to a released version.

mmattel:
ok, then this should be stated.
Q: how do we deal with upcoming (merged) but unreleased (tar) changes from core? where to write those?

phil-davis:
@tboerger @settermjd I guess that 10.0 branch gets pushed to an online docs URL with `10.0` somewhere in it, and 10.1 branch gets pushed to an online docs URL with `10.1` somewhere in it. And the "default" "latest" online would point to the latest official release (so to the `10.1` as of today)

What URL format do we use to see where `master` has been pushed to the web?

tboerger:
Master branch is the latest version selector

phil-davis:
This will work fine. 

- For doc of new features that are release ready, it goes into `master` and sits there and `10.2` branch will happen off `master` at the right time.

- For PRs that correct mistakes in the docs, they go to `master` then are back-ported to any branches that have the mistake.

- For PRs that expand/fill-in things that are "just missing" from the documentation, they will likely need to be back-ported to all the branches (that are currently being published)

phil-davis:
> Master branch is the latest version selector

So when a user/admin goes to the "latest" on the web, they will see documentation about new features that have been merged into core somewhere and documented, but are not yet released.

settermjd:
Hey, @phil-davis, from https://github.com/owncloud/docs/pull/630#issuecomment-465553037, I thought that you might not be too familiar with the version chooser component of Antora's UI, and how the URLs changed, based on the selected version. If so, have a look at the animated gif below.

Show the antora version changer

phil-davis:
@settermjd thanks - I had no idea to actually look down there and select a version!

settermjd:
Hey @phil-davis, that's fair enough. It's new and takes a bit of getting used to.

settermjd:
Just added a few further minor changes. Would love one final review, @tboerger @phil-davis @mmattel, pls.

mmattel:
Thanks for the picture.
This defines the current WF for release 10.1 and below and a rolling master.
Question: how does this picture look like when there is 10.2 (or any new release). There must be a process which is not clear to me.

docs/the-branching-workflow.md
```diff
@@ -0,0 +1,33 @@
+# The Branching Workflow
+
+![The Branching Workflow](images/the-branching-workflow.png.png)
```
tboerger: Duplicated file extension

tboerger:
We should replace the current theme with a theme like https://docs.mulesoft.com/general/ which would show the version switches much more obvious.

settermjd:
Is this what you mean?
 
bildschirmfoto 2019-02-25 um 10 40 27

docs/the-branching-workflow.md
settermjd: Thanks. Removed.

settermjd: Thanks. Removed.

tboerger:
Yes, and also the full list of available docs.

docs/the-branching-workflow.md
```diff
@@ -0,0 +1,33 @@
+# The Branching Workflow
+
+![The Branching Workflow](images/the-branching-workflow.png)
```
tboerger: Now you are linking to a file that doesn't exist?

settermjd: Have another look.

settermjd:
I'm not against a change to the navigation structure, but I'd want to test it out first, and discuss it fully, before enacting the required changes.

tboerger:
> I'm not against a change to the navigation structure, but I'd want to test it out first, and discuss it fully, before enacting the required changes.

This doesn't mean you have to do that as part of this PR, but we should consider to build a theme like that to reduce the confusion.

settermjd:
I wouldn't be doing it as part of this PR. Just continuing the discussion where it began. Please create a new Issue for this.

PR #636, opened 2019-02-21
Use Antora version changer animated gif in what-is-antora docs

settermjd:
This change uses the recently created animated gif, showing how to navigate between different versions of a manual in Antora, rather than the previous still image.


docs/what-is-antora.md
```diff
@@ -27,7 +27,7 @@ In the screenshot above, you can see an example of what Antora documentation loo
 
 While not revolutionary, the layout uses well-recognized and understood navigation conventions. However, it has a rich navigational feature. Notice the link right at the bottom<sup>(5)</sup>. If you click it, it opens up a sub-navigation menu.
 
-![Antora documentation version navigation](./images/antora-version-navigation.png)
```
tboerger: If the png isn't used anymore, why don't you just delete it?

Issue #284, opened 2018-06-19
[OC11] Modify doc for files:scan --groups


phil-davis:
https://github.com/owncloud/core/pull/31719 proposes to modify the ``files:scan --groups`` option so that each of multiple groups are listed by specifying the option multiple times e.g.:
```
occ files:scan --group=foo --group=bar --group=a,b
```
This allows the sysadmin to scan a group that has a literal comma in its name - e.g. ``a,b``

New help output is:
```
occ files:scan --help
Usage:
  files:scan [options] [--] [<user_id>]...

Arguments:
  user_id                Will rescan all files of the given user(s).

Options:
      --output[=OUTPUT]  The output format to use (plain, json or json_pretty, default is plain). [default: "plain"]
  -p, --path=PATH        Limit rescan to this path, e.g., --path="/alice/files/Music", the user_id is determined by the path and the user_id parameter and --all are ignored.
  -g, --group=GROUPS    Scan user(s) under the group(s). This option can be used as --group=foo --group=bar to scan groups foo and bar (multiple values allowed)
  -q, --quiet            Do not output any message
      --all              Will rescan all files of all known users.
      --repair           Will repair detached filecache entries (slow).
      --unscanned        Only scan files which are marked as not fully scanned.
  -h, --help             Display this help message
  -V, --version          Display this application version
      --ansi             Force ANSI output
      --no-ansi          Disable ANSI output
  -n, --no-interaction   Do not ask any interactive question
      --no-warnings      Skip global warnings, show command output only
  -v|vv|vvv, --verbose   Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

Help:
  Scans the filesystem for changes and updates the file cache accordingly.
```

Update the doc, if the PR is merged.

phil-davis:
PR https://github.com/owncloud/core/pull/31719 was merged to core master 6 Aug 2018.
This issue can be documented and then wait until it is known when/what release it will go in.

phil-davis:
PR #642 has the needed changes.

phil-davis:
PR merged to master. It will get released in the "normal" workflow whenever the corresponding core master code is released.

PR #642, opened 2019-02-26
[OC11] Issue 284 document files:scan --group

phil-davis:
Issue #284 

This change should be released with the next core major version release.

So it should be able to be merged to `docs` `master`, but do not backport to any `10.*` branch because it will not be released in any of those versions.

### DIFF
--- a/modules/admin_manual/pages/configuration/server/occ_command.adoc
+++ b/modules/admin_manual/pages/configuration/server/occ_command.adoc
@@ -922,8 +922,8 @@ sudo -u www-data php occ files:scan --help
 | `--output=[OUTPUT]`    | The output format to use (`plain`, `json` or `json_pretty`, default is `plain`).
 | `-p --path=[PATH]`     | Limit rescan to this path, eg. --path="/alice/files/Music",
 the user_id is determined by the path and the user_id parameter and --all are ignored.
-| `-g --groups=[GROUPS]` | Scan user(s) under the group(s).
-This option can be used as --groups=foo,bar to scan groups foo and bar.
+| `-g --group=[GROUP]`   | Scan user(s) under the group(s).
+ This option can be used as --group=foo --group=bar to scan groups foo and bar (multiple values allowed)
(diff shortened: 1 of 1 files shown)


settermjd:
Having this in the documentation also highlights the point you made last week about having master be deployed to the live documentation, and being the default view. It will be confusing for users seeing features that aren't available in their release.

PVince81:
thinking that we might want to backport this change as well as long as it's not a breaking feature @phil-davis

phil-davis:
> Thanks for raising the issue, @phil-davis. Looks good to me. Quick question though; would it not have been simpler to implement further CSV functionality in the options supplied to group, rather than change groups to group? For example, by surrounding fields with double-quotes, as in this example, `--groups="group1",group2,group3,"group,1,2,3"` groups with and without commas in their names can be supplied to the `groups` option.

As I remember, other `occ` commands already have the repeated `--group`` way of doing it, so we kept the same system here.

phil-davis:
> thinking that we might want to backport this change as well as long as it's not a breaking feature @phil-davis

See comment https://github.com/owncloud/core/pull/31719#issuecomment-410631045 - it breaks the old/existing way to type the `occ` command for a group scan.

phil-davis:
> Having this in the documentation also highlights the point you made last week about having master be deployed to the live documentation, and being the default view. It will be confusing for users seeing features that aren't available in their release.

That was the reason that I made the PR now. It presses the point that we should be able to merge doc changes for core master features to docs master branch, and have it "always up-to-date and ready-to-release" (oh, I hear the pigs flying overhead again ;)

PVince81:
hmm right :-/

settermjd:
@phil-davis, re https://github.com/owncloud/docs/pull/642#issuecomment-467353559, I agree with you. That should be the case.

settermjd:
On reviewing the playbook (`site.yml`) again, `master`'s been removed, so showing features for future releases is no longer an issue.

Issue #156, opened 2018-10-12
Document Behat-gherkin step text


phil-davis:
## What Type Of Content Change Is This?
- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [x] Bug Fix to Existing Content

## Which Manual Does This Relate To?
- [ ] Admin Manual
- [x] Developer Manual
- [ ] User Manual

## What Needs to be Documented?
The top part of the developer acceptance-tests.adoc describes the layout of the ``tests/acceptance`` folders and how to write new acceptance test features. It has got a long way out of date.

- [x] Port https://github.com/owncloud/documentation/pull/3870 to the docs repo.
- [x] document the standard grammar forms that we are using to write ``Given`` ``When`` and ``Then`` step text #398 
- [x] document the tags that we use #624 
- [ ] document what steps are available in core (probably just the really common ones that people will make use of a lot in writing new scenarios)
- [ ] give an overview of other steps that are available, and how to search for, find and use
- [x] give guidelines for how to write the code for new API acceptance test steps #650 
- [x] give guidelines for how to write the code for new CLI acceptance test steps #650 (nothing special here that is in principle different to API tests)
- [ ] give guidelines for how to write the code for new webUI acceptance test steps
- [x] give guidelines on what makes a "good" test scenario #398 
- [x] how to write scenarios for bugs (ref: https://github.com/owncloud/QA/issues/601 ) #633 

## Why Should This Change Be Made?

Developers (and managers/designers/end users?) need a reference to help them write good and standardized acceptance tests.

phil-davis:
Issue transferred and updated from the original issue in https://github.com/owncloud/documentation/issues/3829

settermjd:
@phil-davis, it appears, from all the merged PRs and closed issues, that this is resolved. If so, can you pls close it?

settermjd:
@phil-davis, can you help me to complete the final three steps of this issue?

phil-davis:
@settermjd I added the `QA-team`  label (that helps me not miss it - I often filter the whole ownCloud GitHub organisation for the `QA-team` label.
I put it in the next sprint.
This week I am working on some "standardized" Gherkin step text formats for use in test scenarios across clients (Phoenix, desktop, Android, iOS...). So that will get my brain in the space for documenting "typical" Gherkin step examples.

settermjd:
Sweet!

voroyam:
[...]

PR #650, opened 2019-02-26
How to add new acceptance test steps

phil-davis:
Part of issue #156 

This is meant to just give some clues/guidelines for the code of `Given` `When` and `Then` steps. For gory detail of how the whole Behat infrastructure fits together, I think developers need to actually read the Behat User Guide - we do have our `behat.yml`, suites and format of context PHP in the standard Behat way.

It applies equally-well to core `10.0` and `10.1` so can be backported.


phil-davis:
Backports
`10.1` #651 
`10.0` #652

PR #653, opened 2019-02-26
fix Available APIs navigation link

individual-it:
the navigation link of "Availiable APIs" links to "Application Configuration"

image


individual-it:
backports in #662 and #663

Issue #660, opened 2019-02-28
Rename make test-php to test-php-unit


phil-davis:
core https://github.com/owncloud/core/pull/34638 changes `make test-php` to `make test-php-unit`
This standardizes the `make` command for running the PHP unit tests in core to be the same as in the apps.

Update the documentation.

The change is only merged to core `stable10` today. So it only applies when using current core git, not when using a `10.0` or `10.1` system.

PR #661, opened 2019-02-28
Rename make test-php to test-php-unit

phil-davis:
Issue #660 

due to core PR https://github.com/owncloud/core/pull/34638

### DIFF
--- a/modules/developer_manual/pages/core/unit-testing.adoc
+++ b/modules/developer_manual/pages/core/unit-testing.adoc
@@ -80,10 +80,12 @@ available.
 
 Testing apps
 
-* To run test for a specific app with the provided PHPUnit version,
-change into `<webroot>/apps/<appnname>/<testfolder>` and call
-`<webroot>/lib/composer/phpunit/phpunit/phpunit` plus optional
-parameters when needed.
+* To run the tests for a specific app with the provided PHPUnit version,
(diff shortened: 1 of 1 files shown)


phil-davis:
@settermjd this is an interesting case for what to backport.

For changes to the ongoing day-to-day developer environment, we want the developers to be able to find the up-to-date documentation. But we do not want to backport this to `10.1` because actually if a developer checks out the `v10.1.0` tag and tries to `make test-php-unit` then it will not work.

So in this case we do want the "master" documentation to be findable online.

mmattel:
Is backporting required ?

phil-davis:
If you are a developer and have actually chcked out a 10.1 or 10.0 branch:
```
git checkout v10.1.0
git checkout v10.0.10
```
then you have to use the previous `make test-php` to run the unit tests.

If you are using a current in-development branch (which developers will be doing 99% of the time):
```
git checkout stable10
git checkout master
```
then you have to type `make test-php-unit` 

This is an example where we do not backport, but where the "current fluid docs master" needs to be published so that developers can see doc of the current development environment. And developers need to understand that they should usually look at "docs master" when reading developer docs.

mmattel:
Your text is very clear and informative.
Could this be somehow adopted to be a unified text for all doc branches?
That would be imho very helpful and would avoid confusion - or?

PR #664, opened 2019-03-01
DELETE commands do not return any data

individual-it:
sending a DELETE does not result in any returned data, why should it, you just deleted that share.


phil-davis:
LGTM

individual-it:
backports on #666 :imp: and #667

PR #665, opened 2019-03-01
accepting a pending remote share does not return any data

individual-it:
similar to  #664, accepting a pending remote share does not return any data, just the status


phil-davis:
@individual-it I guess this needs to be backported to 10.1 and 10.0 ?

phil-davis:
Backports:
10.1 #672
10.0 #673

PR #667, opened 2019-03-01
[10.1] DELETE commands do not return any data

individual-it:
backport of #664


settermjd:
Sorry, I jumped in and did this already.

PR #681, opened 2019-03-04
Development Docs Update: Install composer and yarn

mmattel:
This is a small update to the development documentation how to install composer and yarn for Debian/Ubuntu systems.


mmattel:
Some comments,
I take as example core, but you also can take other repros.
You need at least make, composer, sometimes yarn and/or others. It is true that these are shared requirements but the documentation to install these prerequisites is very limited. It is expected to either be known or you you have to look by yourself. On the other hand side, unzip is not only used in make but also as running requirement for an owncloud instance, so why naming it here again (see the whole document)? The procedures described here are very good and ease preperation life AND we could link from other repros in github directly to that description if necessary. Therefore I vote to approve and merge.

settermjd:
@mmattel, can you review the changes that I've made, please?

mmattel:
Thanks for the changes - looks all good.

PR #717, opened 2019-03-09
Remove db:generate-change-script

phil-davis:
core PR https://github.com/owncloud/core/pull/34721 removed this.
The core changed is merged to core `stable10` by PR https://github.com/owncloud/core/pull/34722
So it will be released in the next core release, likely to be called 10.2


phil-davis:
It seems an easy deletion. The descriptive text no longer mentions the `db:generate-change-script` command anyway.

PR #720, opened 2019-03-11
Explicitly specify the local branch to use in site.yml

settermjd:
This PR explicitly specifies the local branch to use (`HEAD`) in `site.yml`. Without it specified, [the branches default to `[master, v*]`](https://docs.antora.org/antora/2.0/playbook/playbook-schema/#branches-key). Without it, running `yarn antora` results in `Error: Duplicate nav: master@server:ROOT:nav.adoc`, as `master` is listed as the `version` for a number of branches.

PR #725, opened 2019-03-11
[10.0] [PR 720] Explicitly specify the local branch to use in site.yml

mmattel:
Backport of PR #720

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -4,6 +4,8 @@ site:
 content:
   sources:
   - url: .
+    branches:
+    - HEAD
   - url: https://github.com/owncloud/docs.git
     branches:
     - master
--- a/site.yml
+++ b/site.yml
@@ -9,7 +9,7 @@ content:
   - url: https://github.com/owncloud/docs.git
     branches:
     - master
-    - '10.0'
+    - '10.1'
   - url: https://github.com/owncloud/android.git
     branches:
     - master


site.yml
```diff
@@ -4,6 +4,8 @@ site:
 content:
   sources:
   - url: .
+    branches:
+    - HEAD
   - url: https://github.com/owncloud/docs.git
     branches:
     - master
```
tboerger: Just replace 10.0 below with 10.1

settermjd: I've not seen that work in any testing that I've done locally, as I covered in #720, @tboerger.

tboerger: We are on the `10.0` branch, so the list below must be `master` and `10.1`, everything else is simply wrong, and that's why it's failing now.

settermjd: As I wrote in the initial PR, the reason why it is failing is that no branches were explicitly listed for the first URL. To quote [Dan Allen in the antora/users chat](https://gitter.im/antora/users?at=5c86245dd3b35423cb9d7ef2): 

remember that the default branch pattern is [master, v*]. so that first entry is effectively:

```
- url: .
  branches: [master, v*]
```

settermjd: I may have misunderstood what you were meaning, so just in case; if you mean that `master` should be the branch for the first URL and `10.1` should replace master for the second, please confirm.

tboerger: It's already applied what I was talking about.

```diff
-   - '10.0'
+   - '10.1'
```

settermjd:
Thanks, @tboerger.

PR #731, opened 2019-03-12
Remeber the base branch and switch back when finished

mmattel:
Another backporting script improvement.
Useful if you want to do consecutive backports, eg first to 10.0 and after to 10.1.

**Before** you manually had to switch back eg to the master branch and then do backporting.
**Now** you just can recall the last backport command and change the target branch.

### DIFF
--- a/modules/developer_manual/examples/scripts/backport.sh
+++ b/modules/developer_manual/examples/scripts/backport.sh
@@ -21,11 +21,13 @@ fi
 
 commit=$1
 targetBranch=$2
+baseBranch=$(ggit rev-parse --abbrev-ref HEAD)
+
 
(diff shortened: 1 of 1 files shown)


modules/developer_manual/examples/scripts/backport.sh
```diff
@@ -21,11 +21,13 @@ fi
 
 commit=$1
 targetBranch=$2
+baseBranch=$(ggit rev-parse --abbrev-ref HEAD)
```
settermjd: what's `ggit`?

mmattel: :man_facepalming: thanks catching it, copy/paste error

PR #739, opened 2019-03-13
Development environment based on docker containers

nilo72:
The documentation describes the setup of a development environment based on docker containers where the owncloud code is mounted into the container from the machine you are developing in.


modules/developer_manual/pages/general/devenv_docker.adoc
```diff
@@ -0,0 +1,192 @@
...
+
+Make sure that you have created a folder parallel to the owncloud sources.
+
+[source,console]
+....
+/home/<your-username>/src/oc_dev_docker
+....
+
+[[build-a-docker-container]]
+=== Build a docker container
```
tboerger: You don't even need that, for CI and development you could use `owncloudci/base` which also includes make and npm.

nilo72: Ok, cool... that make things even easier.

nilo72: Hm... using owncloudi/base leads to following error:

`ERROR: The image for the service you're trying to recreate has been removed. If you continue, volume data could be lost. Consider backing up your data before continuing. Continue with the new image? [yN]n ERROR: pull access denied for owncloudi/base, repository does not exist or may require 'docker login'`

using owncloud/base downloads the image but if I try to execute make:

`bash: make: command not found`

nilo72: ok ok dyslexia trap.... owncloud **c** i/base works....

nilo72: I removed the section 'Build a docker container' and modified the image to use in the compose file.

settermjd:
@nilo72, I've made a change to the PR with the aim of helping to streamline the content. Please have a read through and let me know what you think.

mmattel:
Backport with fix
10.0: https://github.com/owncloud/docs/pull/881
10.1 https://github.com/owncloud/docs/pull/882

PR #748, opened 2019-03-14
[WIP]: Update latest manual version and fix previous breaking changes.

settermjd:
This cherry-picks ed7836e61e84d157b5201c91950ade22e64bb787 to master.

💁‍♂️ this needs a fix for the 10.1 link to the admin manual. I'm waiting on community feedback about what might be wrong.


books/ownCloud_Admin_Manual.adoc
```diff
@@ -10,42 +10,71 @@ The ownCloud Team <docs@owncloud.com>
 :icon-set: octicon
 
 include::modules/admin_manual/pages/index.adoc[leveloffset=+1]
+
```
tboerger: Why is that needed? That doesn't make sense...

settermjd: It did, at least previously, as I commented in the PR that removed the newlines. mojavelinux gave a recommendation to do that when working with asciidoctor-pdf. I don't have a link to further information, but not having them caused corrupted build artifacts previously.

tboerger: What does corrupted artifacts mean? So far the PDFs are building fine, it's just printing warnings while building it.

settermjd: I'd have to dig into what was happening previously to be more specific. If it's working, then I'll update the PR.

settermjd:
Extra spacing in book configuration file removed.

PR #756, opened 2019-03-14
Fix wrong include directives in the developer manual

mmattel:
- Found a bunch of wrong defined include::example/partials directives in the developer docs which were not rendered therefore.
- Fixed missing includes for existing example files
- Moved/fixed/include some example files to the correct examples directory
- Fixed tag render errors
- Improved/fixed some text

There is still a `developer_manual/pages/examples/` directory which needs checking where these examples are used (if?) and then moved/corrected/deleted to `developer_manual/examples/`


phil-davis:
Also note issue #758 
But we can fix the structure in this PR, then fix the content in a later PR.

Issue #758, opened 2019-03-15
Invalid code in MyStorage.php


phil-davis:
## What Needs to be Documented?
The example code shows various warnings an errors in PHPstorm, like:

```
	public function getConnection() {
		if ($this->connection === null) {
			// do the connection to the storage lazily
			$this->connection = new \SomeVendor\FakeStorageLib\Connection($params);
		}
		return $this->connection;
	}
```
Undefined variable $params

```
	public function fopen($path, $mode) {
		switch ($mode) {
			case 'r':
			case 'rb':
				// this works if the library returns a PHP stream directly
				return $this->connection->getStream($path);
			case 'w':
			case 'w+':
			case 'wb':
			case 'wb+':
			case 'a':
			case 'ab':
			case 'r+':
			case 'a+':
			case 'x':
			case 'x+':
			case 'c':
			case 'c+':
				// most storages do not support on the fly stream upload for all modes,
				// so we use a temporary file first
				$ext = pathinfo($filename, PATHINFO_EXTENSION);
				$tmpFile = \OC::$server->getTempManager()->getTemporaryFile($ext);

				// this wrapper will call the callback whenever fclose() was called on the file,
				// after which we send the file to the library
				$result = CallbackWrapper::wrap(
					$source,
					null,
					null,
					function () use ($tmpFile, $path) {
						$this->connection->putFile($tmpFile, $path);
						unlink($tmpFile);
					}
				);
		}
		return false;
	}
```
Undefined variable $filename

```
	public function fopen($path, $mode) {
		switch ($mode) {
			case 'r':
			case 'rb':
				// this works if the library returns a PHP stream directly
				return $this->connection->getStream($path);
			case 'w':
			case 'w+':
			case 'wb':
			case 'wb+':
			case 'a':
			case 'ab':
			case 'r+':
			case 'a+':
			case 'x':
			case 'x+':
			case 'c':
			case 'c+':
				// most storages do not support on the fly stream upload for all modes,
				// so we use a temporary file first
				$ext = pathinfo($filename, PATHINFO_EXTENSION);
				$tmpFile = \OC::$server->getTempManager()->getTemporaryFile($ext);

				// this wrapper will call the callback whenever fclose() was called on the file,
				// after which we send the file to the library
				$result = CallbackWrapper::wrap(
					$source,
					null,
					null,
					function () use ($tmpFile, $path) {
						$this->connection->putFile($tmpFile, $path);
						unlink($tmpFile);
					}
				);
		}
[...]

PR #763, opened 2019-03-16
moved all scripts to examples dir and added many to the documentation

### LINKED ISSUE #757: Missing examples in developer docs
`developer_manual/pages/examples/` contains example files which:
- should be moved to `developer_manual/examples/`
- are currently orphaned. Means present but "nowhere" used.
Example:
`developer_manual/pages/examples/curl/create-share.sh`
Should be included in file `developer_manual/pages/core/ocs-share-api.adoc` line 286
- The example above is important as there are example files in `curl, ruby, php, go` to be added accordingly

developer docs

## Why Should This Change Be Made? (Optional)
Move and fix missing examples so that they are shown

- [ ] New Content Addition
- [ ] Old Content Deprecation

- [ ] Admin Manual
- [x] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

One first step fixing this issue is addressed in: https://github.com/owncloud/docs/pull/756
(the part with cron)

mmattel:
Fixes #757 (Missing examples in developer docs)

**Old:**
There have been many scripts in directory `developer_manual/pages/example` and these scripts have nowhere been used.

**New:**
All these scripts are now moved to `developer_manual/examples` directory, plus many of the scripts have been added to the `ocs-share-api.adoc` page

**Notes:**
There is a directory `developer_manual/examples/core/scripts/responses` with responses/scripts I cant identify to add to the above or any other document. We need support to add them to the correct location/page or delete them if obsolete.

I have added a text **Missing** in `ocs-share-api.adoc` line 86 because it is clear that a xml response example needs to be included but I do not know which one.

We also need your response if those scripts are up to date and working as I have NOT checked for working content but relationship for the include directive only.

There are maybe **OTHER** scripts which are linked properly but may be outdated and do not work, see issue: https://github.com/owncloud/docs/issues/758 (Invalid code in MyStorage.php) which could be fixed in this PR too.

It is possible that there are files which are orphaned, completely outdated and/or are not needed anymore like:
`developer_manual/examples/core/scripts/php/vendor/guzzlehttp/guzzle/src/Client.php`
Please identify and delete or include/correct them in the right page/location.

Backporting most likely to be necessary but to be decided.

Please add additional reviewers if they can help.


mmattel:
Thanks for your comments @phil-davis 
Please consider the **Missing** text I added in the file above which either needs to be removed or with a include directive replaced.

PR #764, opened 2019-03-16
Improving Linkchecker

mmattel:
This is a small improvement with regards of the command
`yarn linkcheck http://localhost:8080`

Added grep to filter broken results as the rest is not of value
`yarn linkcheck http://localhost:8080 | grep "BROKEN"`
Adopted the output accordingly and added a text line to highlight that there are real broken links.

### DIFF
--- a/docs/checking-broken-links.md
+++ b/docs/checking-broken-links.md
@@ -68,8 +68,8 @@ yarn linkcheck http://localhost:8080 | grep "BROKEN"
 ├─BROKEN─ https://example.com/owncloud (HTTP_404)
 ...
 ```
-Please consider that only the first entry is a broken link as the others in the list
-are example links and therefore not broken.
+Note: in the example output above, only the first entry is a genuine broken link.
+All the others in the list are example links, and therefore not broken.
 
 ### A general linkchecker utility
 
--- a/docs/checking-broken-links.md
+++ b/docs/checking-broken-links.md
@@ -54,21 +54,22 @@ worktree: /var/www/owncloud/docs | component: server | version: master
 If you already installed the Antora dependencies via `yarn install` you already got a broken link checker included, you can simple execute the following command:
 
 ```console
-yarn linkcheck http://localhost:8080
+yarn linkcheck http://localhost:8080 | grep "BROKEN"
 ```
 
 #### Example Output
 
 ```console
 ...
-$ broken-link-checker --filter-level 3 --recursive --verbose https://doc.owncloud.com
-Getting links from: http://localhost:8080/client/automatic_updater.html
-├───OK─── https://github.com/owncloud/client/edit/master-antora/docs/modules/ROOT/pages/automatic_updater.adoc
-├───OK─── https://doc.owncloud.org/branded_clients/
-├─BROKEN─ https://owncloud.org/history/ (HTTP_404)
-Finished! 60 links found. 57 excluded. 1 broken.
+├─BROKEN─ http://mechanics.flite.com/blog/2014/07/29/using-innodb-large-prefix-to-avoid-error-1071/ (BLC_UNKNOWN)
+├─BROKEN─ http://bucket.hostname.domain/ (ERRNO_ENOTFOUND)
+├─BROKEN─ http://hostname.domain/bucket (ERRNO_ENOTFOUND)
+├─BROKEN─ https://example.com/owncloud (HTTP_404)
+├─BROKEN─ https://example.com/owncloud (HTTP_404)
 ...
 ```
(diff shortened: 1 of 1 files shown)


docs/checking-broken-links.md
```diff
@@ -54,21 +54,22 @@ worktree: /var/www/owncloud/docs | component: server | version: master
...
-├─BROKEN─ https://owncloud.org/history/ (HTTP_404)
-Finished! 60 links found. 57 excluded. 1 broken.
+├─BROKEN─ http://mechanics.flite.com/blog/2014/07/29/using-innodb-large-prefix-to-avoid-error-1071/ (BLC_UNKNOWN)
+├─BROKEN─ http://bucket.hostname.domain/ (ERRNO_ENOTFOUND)
+├─BROKEN─ http://hostname.domain/bucket (ERRNO_ENOTFOUND)
+├─BROKEN─ https://example.com/owncloud (HTTP_404)
+├─BROKEN─ https://example.com/owncloud (HTTP_404)
 ...
 ```
+Please consider that only the first entry is a broken link as the others in the list
```
phil-davis: Maybe:
```
Note: only the first entry is a real broken link. The others in the list are example links and therefore not broken.
```
@settermjd how would you write this?

mmattel: @phil-davis info, this is page markdown (.md), not asciidoc (.adoc). md does not have the directive NOTE: and it will not render like we are used in adoc. The only thing we can do is write Note: and some text which will all be all rendered as text. I am open for any improvement.

tboerger:
I haven't checked it again, but is there some flag to ignore specific domains?

docs/checking-broken-links.md
settermjd: @phil-davis, good question. I'll have to review and think it over.

settermjd: tbh, I'm not sure what the comment means.

settermjd:
@tboerger, the `--exclude` option would be a good fit.

docs/checking-broken-links.md
settermjd: Please clarify what you mean, @mmattel.

tboerger:
Than we should add all the example domains to the excludes within the package.json command definition.

mmattel:
@tboerger your attempt having an exclude directive (which is available) was one of the first things that came to my mind. The bad thing is, our documentation has a so many different example links. We could use this directive, but then we would have to harmonize in the complete documentation the way we define example links making an exclude usable.

@settermjd 
> I'm not sure what the comment means.

Take a look to the change,
```
├─BROKEN─ http://mechanics.flite.com/blog/2014/07/29/using-innodb-large-prefix-to-avoid-error-1071/ (BLC_UNKNOWN)
├─BROKEN─ http://bucket.hostname.domain/ (ERRNO_ENOTFOUND)
...
```
The first line is a real broken link while the next lines are example links and therefore not broken.

settermjd:
@mmattel, gotcha. Then it's clear that we need to standardise on an example domain. CAn you careate a new issue to cover this please?

docs/checking-broken-links.md
settermjd: @mmattel, given your further clarification, I suggest the following sentence "_Note: in the example output above, only the first entry is a genuine broken link. All the others in the list are example links, and therefore not broken._".

tboerger:
Then I would prefer a consolidation of the example links to make the exclude useful instead of grepping through the link checker results.

settermjd:
@tboerger, I added a list of all the custom domains in the discussion in #770, along with a suggestion or two.

mmattel:
@tboerger,
in any case you need this PR.
- First grep will filter for broken links, else you get all findings.
-Second you cant cover all excludes.

But @settermjd upcoming work will reduce the broken output massively!

settermjd:
@mmattel, am I missing something? If we centralise the example domains, then no grepping, et al, will be required.

settermjd:
It'd be ideal to add some kind of git hook or other tool to help enforce a standard.

mmattel:
BLC output shows ALL results.
We wanto to focus on BROKEN once only.
And reduce the result list by exclude as suggested.
Give it a quick try without/with grep and you get it :smile:

mmattel:
Would it be ok to approve/merge this PR and file additional PR´s for 
- issue #770 (Standardize how example domains are written in docs)
- adding `--exclude` (s) to the underlaying yarn command

These are separate tasks and it would make things faster easier.

docs/checking-broken-links.md
mmattel: Done

PR #766, opened 2019-03-16
yarn run (build-the-docs.md)

mmattel:
Add `yarn run` to get a list of available  yarn commands which also can be started from that command.


settermjd:
Also, how about adding a link to [the yarn run documentation](https://yarnpkg.com/lang/en/docs/cli/run/)?

PR #769, opened 2019-03-18
Add Linters for Kotlin and Golang

settermjd:


mmattel:
Referencing https://github.com/owncloud/docs/pull/764 because there are script examples in Kotlin and Go which are now included and we need checking therefore.

tboerger:
Seriously? Why don't we start to add linters for every existing language???

settermjd:
Simple; because I'm adding linters for the languages that have existing code samples. No more. No less.

PR #783, opened 2019-03-19
Improving the cherry pick procedure in backporting script

mmattel:
This is an improvement of the backporting script which handles the commits to be picked in a better way. Plus some output tuning and text improvements.

### DIFF
--- a/modules/developer_manual/examples/scripts/backport.sh
+++ b/modules/developer_manual/examples/scripts/backport.sh
@@ -23,11 +23,9 @@ commit=$1
 targetBranch=$2
 baseBranch=$(git rev-parse --abbrev-ref HEAD)
 
-
 is_merged=$(git branch --contains $1 | grep -oP '(?<=\*).*')
-
 if [ -z "$is_merged" ]; then
-    echo "$commit has not been merged or $baseBranch is not pulled/rebased. Exiting"
+    echo "$commit has not been merged or branch $baseBranch is not pulled/rebased. Exiting"
     echo
     exit
 fi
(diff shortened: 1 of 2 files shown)


modules/developer_manual/examples/scripts/backport.sh
```diff
@@ -51,6 +52,17 @@ now=$(date +%s)
 # time remaining in HMS
 remaining=$(date -u -d @$remaining +%H:%M:%S)
 
+# check if there are commits to cherry pick and list them in case
+if [[ -z "$commitList" ]]; then
+  echo "There are no commits to cherry pick. Exiting"
+  exit
+else
+  lineCount=$(echo "$commitList" | grep '' | wc -l)
+  echo "$lineCount commits beeing cherry picked:"
```
phil-davis: s/beeing/being

modules/developer_manual/examples/scripts/backport.sh
```diff
@@ -38,6 +36,9 @@ pullId=$(git log $1^! --oneline 2>/dev/null | tail -n 3 | grep -oP '(?<=#)[0-9]*
 # get the repository from commit
 repository=$(git config --get remote.origin.url 2>/dev/null | perl -lne 'print $1 if /(?:(?:https?:\/\/github.com\/)|:)(.*?).git/')
 
+# get the list of commits in PR without any merge commit
+commitList=$(git log --no-merges --reverse --format=format:%h $1^..$1)
```
phil-davis: Note to other readers who (like me) might have to think about what this does: `$1^` means the first parent of the merge commit (that is passed in as `$1`). Because `$1` is a "magically-generated" merge commit, it happily "jumps back" to the point on the main branch just before where the PR was merged. And so the commits from that point are exactly the list of individual commits in the original PR.
`--no-merges` leaves out the merge commit itself, and we get just what we want - works nicely.

mmattel: Thanks for your feedback.
I have fixed the grammar from above and added your note as comment for other readers

phil-davis:
Now you can test the script by making the backports :) @mmattel

PR #802, opened 2019-03-22
Document the HSM Daemon

### LINKED ISSUE #586: How to enable HSM with encryption
This ticket is regarding documentation for integrating HSM with encryption  reference https://github.com/owncloud/encryption/pull/90. From the UI point of view its enabling HSM and adding the JWT secret and mentioning the URL in the text box. But there are other changes required to be made in the instance where oC is running.

#### Configuring host machine running oC instance ####
I am using Ubuntu 18.04 ( Neon KDE )
- Install `softhsm2` 
- After installation of `softhsm2` we would see 2 additional packages installed `libsofthsm2` and `softhsm2-common`
- Now we can configure file `/etc/softhsm/softhsm2.conf`. I have configured as shown below in my machine:
```
# SoftHSM v2 configuration file

directories.tokendir = /var/lib/softhsm/tokens/
objectstore.backend = file

# ERROR, WARNING, INFO, DEBUG
log.level = INFO
```
- Not the line `directories.tokendir = /var/lib/softhsm/tokens/`. If this directory is not there then please do create in your machine. In my machine the permissions for this directory is as shown below
```
drwxr-sr-x 3 root softhsm 4096 Jan 17 20:40 /var/lib/softhsm/tokens/
```
- Now we need to initialize tokens in softhsm
     1. Inorder to initialize the token execute command:
```
sudo softhsm2-util --init-token --slot 0 --label "My token 1"
```
      2. This would ask for pin number. Enter a value, say `1234` and kindly note it down.
      3. Next it would ask for the secret value. Enter `secret`
- Now install `opensc`, which is a command line tool to use `pkcs11`
```
sudo apt install opensc
```
- In the command line you may try with
```
sudo pkcs11-tool --module /usr/lib/softhsm/libsofthsm2 -l --pin 1234 -O
```

- Now try to build the hsmdaemon. There is a makefile for the same.
- Run `make` in the hsmdaemon folder
- An executable `hsmdaemon` would be created.
- Create a file `/etc/hsmdaemon/hsmdaemon.toml` with contents as shown below:
```
# This is a .ini like TOML document. See https://github.com/toml-lang/toml#example for detailed synax

#debug = "*"

[server]
#hostname = "localhost"
#hostname = "" # to listen on all interfaces
#port = 8513

[jwt]
secret = "secret"

[pkcs11]
module = "/usr/lib/softhsm/libsofthsm2.so" # softhsm v1
# module = "/usr/lib/x86_64-linux-gnu/softhsm/libsofthsm2.so" # softhsm v2
# module = "/opt/nfast/toolkits/pkcs11/libcknfast.so" # Thales nShield

pin = "1234"
slot = 615007925
#slot = 11110 # find your slot id with `hsmdaemon listslots`

persistent-connection = false
```
- For testing you may run with `DEBUG="*" sudo -E ./hsmdaemon` on terminal

#### Now come back to the UI ####
- From the UI enable encryption app
- Navigate to the encryption page in the admin settings, you should see the page as shown 
encryption_page
- To enable HSM tick the check box and fill the text box. The JWT secret is `secret` and the URL is `http://localhost:8513`
hsm_config_added
- Now try to Enable the encryption 
enabling_encryption
- Now enable master key
- Logout and login back ( as admin )
- Now verify that private key in the files_encryption/OC_DEFAULT_MODULE should be less than 1000 bytes 
files_encryption_dir
- You should see messages in the terminal where `hsmdaemon` running as 
```
✘ sujith@sujith-ownCloud  ~/owncloud/hsmdaemon   dev  DEBUG="*" sudo -E ./hsmdaemon
2019-02-07T14:44:58.771+0530    INFO    hsmdaemon/hsmdaemon.go:145      Server listening        {"host": "localhost", "port": 8513, "version": "", "build": "2019-02-06T13:50:48+00:00"}
2019-02-07T14:45:39.802+0530    DEBUG   hsmdaemon/webserver.go:47       Requested       {"remote": "127.0.0.1:59978", "url": "/keys/new"}
2019/02/07 14:45:39 Token extracted: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJvY2h5ZjJlMDZyanAiLCJzdWIiOiJvYzpwdWJTaGFyZV81MWI1ODgyMCIsImF1ZCI6ImhzbWRhZW1vbiIsImV4cCI6MTU0OTUzMTA1OSwicmlkIjoiM3JsUTAyR1BCelNqNGNMeXZFaEIiLCJpYXQiOjE1NDk1MzA5MzksImp0aSI6IjVjNWJmNzNiYTkwNzk2Ljc0NDY4OTIyIn0.3B-4hu3l3DKJcj5M8NXVFnHr9YriHdGi0pxq8I_VAXc
2019/02/07 14:45:39 JWT: &{Raw:eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJvY2h5ZjJlMDZyanAiLCJzdWIiOiJvYzpwdWJTaGFyZV81MWI1ODgyMCIsImF1ZCI6ImhzbWRhZW1vbiIsImV4cCI6MTU0OTUzMTA1OSwicmlkIjoiM3JsUTAyR1BCelNqNGNMeXZFaEIiLCJpYXQiOjE1NDk1MzA5MzksImp0aSI6IjVjNWJmNzNiYTkwNzk2Ljc0NDY4OTIyIn0.3B-4hu3l3DKJcj5M8NXVFnHr9YriHdGi0pxq8I_VAXc Method:0xc42011c660 Header:map[typ:JWT alg:HS256] Claims:0xc42011cde0 Signature:3B-4hu3l3DKJcj5M8NXVFnHr9YriHdGi0pxq8I_VAXc Valid:true}
2019/02/07 14:45:39 Claims: &{RequestId:3rlQ02GPBzSj4cLyvEhB StandardClaims:0xc4201d8060}
2019-02-07T14:45:39.807+0530    WARN    hsmdaemon/webserver.go:63       Generating new key
```

- [x] New Content Addition
- [ ] Old Content Deprecation

- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

settermjd:
I'm not quite sure if this is what you want, but I've done my best to take the existing hsmdaemon documentation, to work through it and revise it, where necessary; after which I've integrated it into the ownCloud docs.

This fixes #586.


modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,514 @@
...
+=== ownCloud Setup
+
+==== Shut Down Apache
+
+If anyone accesses ownCloud while encryption is enabled, it will automatically generate the keys. 
+To prevent this, shut down Apache until encryption is appropriately configured.
+
+==== Patch ownCloud
+
+NOTE: The PR for HSM support https://github.com/owncloud/core/pull/34205[is in review].
```
sharidas: The PR for HSM support https://github.com/owncloud/core/pull/34527 ( merged ). The one mentioned https://github.com/owncloud/core/pull/34205 is not required. It is a UI enhancement, we do not want to expose this to UI, hence the PR is closed.

mmattel:
Not everyone is aware about the term HSM. I would therefore take the long name into the navigation and page header and then continue with the abbreviation.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,514 @@
...
+----
+
+===== Installation Example
+
+To save time, copy and paste the commands below to compile the hsmdaemon.
+
+[source,console]
+----
+# Install hsmdaemon's dependencies
+go get
```
mmattel: Is `go` part of the standard distribution or does it needs to be installed first?

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,514 @@
...
+==== Patch ownCloud
+
+NOTE: The PR for HSM support https://github.com/owncloud/core/pull/34205[is in review].
+
+For now, we assume a fresh ownCloud installation. 
+Patch the encryption app, by following the example below.
+
+[source,console]
+----
+$ cd /path/to/owncloud/installation
```
mmattel: Proposal:
`$ cd /path/to/owncloud/installation` —> `$ cd /path/to/owncloud_root/`
Else it could be understood to change to directory installation

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,514 @@
...
+$ patch -p1 < /path/to/hsmdaemon/patch/hsmdaemon.patch
```
mmattel: Where do you get this pathch from?

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: Just read the PRs. Understood. I'll remove this segment.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: Well said! I didn't think of that at the time. I'll update the content accordingly.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,588 @@
...
+Because of this, a separate process is needed to decrypt anything with the private key stored in an HSM. 
+
+Running {php-exec-function-url}[exec()] to decrypt the key with a command line command to do the encryption might leak the HSM credentials if the admin lists the currently running processes. 
+To prevent that an "HSM" daemon will be used that can open a session to the HSM upon startup. 
+
+This daemon will be used by ownCloud to decrypt the current master key upon request. 
+The communication happens via {unix-sockets-url}[UNIX sockets] or {network-sockets-url}[TCP sockets] and is authorized by a shared token that the daemon stores in the ownCloud database via a REST/JSON route.
+
+ownCloud internally uses OpenSSL to en-/decrypt keys and needs to be extended to support en-/decrypt operations via the new daemon. 
+The current solution would be to encrypt the current ownCloud master key with a key from the HSM.
```
mmattel: Do I read correct, that you encrypt a key by a key or is it meant the the keys are replaced

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,588 @@
...
+== Quick Overview
+
+HSM support consists of two core parts:
+
+. An actual HSM PKCS11 module.
+. An hsmdaemon that provides a {jwt-url}[JWT]-protected web API for the PKCS11 stack to generate key pairs and decrypt data.
+
+== Deployment Recommendation
+
+We recommend running hsmdaemon on every Apache server to reduce latency.
```
mmattel: could we stay neutral and say web server instead or is it mandatory apache. if so this must be highlighted.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,588 @@
...
+. An actual HSM PKCS11 module.
+. An hsmdaemon that provides a {jwt-url}[JWT]-protected web API for the PKCS11 stack to generate key pairs and decrypt data.
+
+== Deployment Recommendation
+
+We recommend running hsmdaemon on every Apache server to reduce latency. 
+
+== Installation
+
+Integrating the hsmdaemon with ownCloud requires X steps; these are:
```
mmattel: `requires X steps` ?
`requires some steps`

modules/admin_manual/nav.adoc
```diff
@@ -64,6 +64,7 @@
 **** Security
 ***** xref:configuration/server/security/password_policy.adoc[Password policy]
 ***** xref:configuration/server/security/oauth2.adoc[OAuth2]
+***** xref:configuration/server/security/hsmdaemon.adoc[The HSM Daemon]
```
mmattel: Not everyone is aware about the term HSM. I would therefore take the long name Hardware Security Module into the navigation and page header and then continue with the abbreviation HSM.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,588 @@
...
+----
+
+==== Test Data Encryption
+
+For testing data encryption, run the `hsmdaemon encrypt` command, as in the following example.
+
+[source,console]
+----
+The first argument is the "Id:" value from running the genkey command above.
+$ sudo hsmdaemon encrypt 9bac3719-2b8d-11e9-aeab-0242b5ece4c3 Zm9vYmFy
```
mmattel: what is the second value `Zm9vYmFy` ?
where does it come from?

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,588 @@
...
+}
+----
+====
+
+=== Configure ownCloud
+
+[TIP]
+====
+If anyone accesses ownCloud while encryption is enabled, it will automatically generate the keys. 
+To prevent this, shut down Apache until encryption is appropriately configured.
```
mmattel: `shut down Apache`
`shut down your web server`

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,588 @@
...
+
+////
+==== Configure Authorization
+
+TBW.
+////
+
+==== Initialize and Check Generated Keys
+
+Now start Apache, and log in with any user to initialize the keys, have a look at the output of the hsmdaemon to see key generation and decryption requests.
```
mmattel: `start Apache`
`start your web server`

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,588 @@
+= The HSM Daemon (hsmdaemon)
+:base64-encoding-url: https://en.wikipedia.org/wiki/Base64
+:hsm-url: https://en.wikipedia.org/wiki/Hardware_security_module
+:jwt-url: https://jwt.io/
+:network-sockets-url: https://en.wikipedia.org/wiki/Network_socket
+:go-install-url: https://golang.org/doc/install
```
mmattel: now you have that link but it is nowhere used - intended?

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: I see your point. While Apache (I believe) is the only supported web server, others are community-supported.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: Well spotted. I need to add some kind of generic marker to remind me when I do this.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: Likely was in an earlier revision. Well spotted.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: Well spotted. It's the base64-encoded data to be encrypted.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: Not sure. Checking.

settermjd: @butonic, can you answer this please?

sharidas: Not exactly. The keys are not replaced. From the technical point of view the Crypt class is extended to handle the key generation in the HSM device and also to get the key from HSM. So basically for the read/write operation on a file, the request goes to the HSM and then based on the keys fetched from HSM the files are encrypted/decrypted. Does that answer your query?

settermjd: Thanks for the further input, @sharidas.

settermjd: If fe414b4efa8ce8cb718ec648184bf55b2d64418c is okay, @sharidas, can you approve the PR, pls?

sharidas: Looks ok to me :+1:

mmattel: Sorry I need to insist...
The answer to my question is not clear nor the impact.
It currently says that you encrypt oC masterkey with a key from HSM.
Is it what it reads or is it different. What does this mean. What is the impact of a key re-encrypted by HSM.
The sentence also says „would be“. This hands over the insecurity of the process to the admin and also tells that we do not know. Either we say it clear - this is the process (with a description of it) or we rewrite this statement and describe the opportunities/possibilities and their impact.

mmattel:
Before merging, pls see my comment https://github.com/owncloud/docs/pull/802#discussion_r270768907

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: @sharidas, can you clarify, pls?

settermjd: @mmattel, can you have another look and let me know if the change resolves your concerns?

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,592 @@
...
+:pkcs11-tool-url: https://linux.die.net/man/1/pkcs11-tool 
+:softhsm2-url: https://www.opendnssec.org/softhsm/
+:unix-sockets-url: http://beej.us/guide/bgipc/html/multi/unixsock.html
+
+The hsmdaemon is a daemon to delegate encryption to an {hsm-url}[HSM (Hardware Security Module)].
+It can be necessary, as PHP cannot directly interface with {pkcs11-url}[a PKCS11 stack]; neither by an API wrapper, because one does not exist, yet, nor via the OpenSSL bindings
+Because of this, a separate process is needed to decrypt anything with the private key stored in an HSM. 
+
+Running {php-exec-function-url}[exec()] to decrypt the key with a command line command to do the encryption might leak the HSM credentials if the admin lists the currently running processes. 
+To prevent that an "HSM" daemon will be used that can open a session to the HSM upon startup.
```
mmattel: I am not a native speaker, but doesn´t it write `a "HSM"`?

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
mmattel: Concern solved, thanks for taking care of. Marking as resolved

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,592 @@
...
+We recommend running hsmdaemon on every web server to reduce latency. 
+
+== Installation
+
+Integrating the hsmdaemon with ownCloud requires 4 steps; these are:
+
+. xref:install-a-pkcs11-module[Install a PKCS11 Module]
+. xref:install-and-configure-the-hsmdaemon[Install and Configure the hsmdaemon]
+. xref:configure-owncloud[Configure ownCloud]
+. xref:install-the-hsmdaemon-as-a-system-service[Install the hsmdaemon as a System Service]
```
mmattel: Question: why is `configure-owncloud` before `install-the-hsmdaemon-as-a-system-service` ?
Shouldnt it be install pkcs11, install hsmdeamon, hsm as system service, configure owncloud?
It may has a reason but the sort order sounds strange

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
mmattel: For better readability, I would add a comma after „To prevent that“

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: Good point. I'm not sure why.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: Re https://github.com/owncloud/docs/pull/802#discussion_r271321567, no. If the sound of the word following the preposition comes from a vowel, then using 'a' would be fine, but as it comes from a consonant, 'an' is correct.

settermjd: Re https://github.com/owncloud/docs/pull/802#discussion_r271375317, done!

settermjd:
I've added meta keywords and a meta description to aid in SEO discoverability.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
mmattel: thanks for the explanation :smile:

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -0,0 +1,595 @@
...
+:network-sockets-url: https://en.wikipedia.org/wiki/Network_socket
+:opensc-wiki-url: https://github.com/OpenSC/OpenSC/wiki
+:opensuse-security-repositories-url: https://download.opensuse.org/repositories/security/
+:php-exec-function-url: https://www.php.net/manual/en/function.exec.php
+:pkcs11-url: https://en.wikipedia.org/wiki/PKCS_11
+:pkcs11-tool-url: https://linux.die.net/man/1/pkcs11-tool 
+:softhsm2-url: https://www.opendnssec.org/softhsm/
+:unix-sockets-url: http://beej.us/guide/bgipc/html/multi/unixsock.html
+
+The hsmdaemon is a daemon to delegate encryption to an {hsm-url}[HSM (Hardware Security Module)].
```
mmattel: It is just a small thing.
On the installation of the daemon below, it is first time mentioned that this daemon is provided by owncloud.
I would propose therefore to write here at the beginning:
`The hsmdaemon is a daemon to` --> `The hsmdaemon is a daemon, provided by ownCloud, to`.
Additionally I propose to add a note here in this top section how to obtain it, who is eligable to obtain it, as this document is NOT in the enterprise section.

settermjd: As always, well spotted.

settermjd: @sharidas or @patrickjahns, can you provide further details on how the daemon is obtained and who is able to receive a copy?

mmattel:
I did the following exercise to improve overview (not commited, just tried), added:
```
= The HSM ...
:toc:
...
== Introduction
...
```
See the result:
image

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
patrickjahns: This is something for @hodyroff and @pmaier1 ( or even @CaptionF )

pmaier1: > @sharidas or @patrickjahns, can you provide further details on how the daemon is obtained and who is able to receive a copy?

So, there are two components to this:
- The `encryption` app is part of ownCloud Server and will have out-of-the-box support for the `hsmdaemon` with the upcoming 10.2 release.
- The `hsmdaemon` itself will then be available via our customer portal as part of the Enterprise Edition.

pmaier1:
Wow, this is huge. Thanks for the work!

Some feedback:
- If I'm not mistaken, this documentation should reflect how you set up "Storage encryption with a master key stored in a HSM" and setting up the `hsmdaemon` is a part of this.
- Not sure whether we need exhaustive explanations on how to set up softhsm here. In the end you would only use this for testing purposes. Correct me when I'm wrong, please :)

settermjd:
@sharidas @butonic can you provide feedback on https://github.com/owncloud/docs/pull/802#issuecomment-479806837 from @pmaier1, pls?

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -14,12 +14,12 @@
...
 
-== Introduction
-
-The hsmdaemon is a daemon to delegate encryption to an {hsm-url}[HSM (Hardware Security Module)].
-It can be necessary, as PHP cannot directly interface with {pkcs11-url}[a PKCS11 stack]; neither by an API wrapper, because one does not exist, yet, nor via the OpenSSL bindings
+The hsmdaemon is a daemon, provided by ownCloud, to delegate encryption to an {hsm-url}[HSM (Hardware Security Module)].
+This can be necessary, as PHP cannot, directly, interface with {pkcs11-url}[a PKCS11 stack]; neither with an API wrapper, because one does not exist, nor via the OpenSSL bindings.
 Because of this, a separate process is needed to decrypt anything with the private key stored in an HSM. 
 
+NOTE: The encryption app, which is bundled with ownCloud Server, will have out-of-the-box support for the hsmdaemon with the upcoming 10.2 release.
```
mmattel: Would´nt it be better to write:
`for the hsmdaemon with the upcoming 10.2 release` --> `for the hsmdaemon with release 10.2 upwards`.
Because at one point in time upcoming will be present respectively past...

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
```diff
@@ -56,6 +56,10 @@ HSM support consists of two core parts:
 
 We recommend running hsmdaemon on every web server to reduce latency. 
 
+== Getting the hsmdaemon?
```
mmattel: Using a questionmark does not look good (imho)

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: It will. The text can be updated when 10.2's released to remove the specific reference to that version.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
settermjd: Will remove that.

modules/admin_manual/pages/configuration/server/security/hsmdaemon.adoc
pmaier1: I think you can just remove the reference as the productizing of this whole feature will anyway take place with/after 10.2. Maybe just state that you need Server >= 10.2 for HSM support.

sharidas:
> * Not sure whether we need exhaustive explanations on how to set up softhsm here. In the end you would only use this for testing purposes. Correct me when I'm wrong, please :)

Yes softhsm would only be used for testing purpose.

settermjd:
As the work was already invested to create the instructions, I've moved the SoftHSM2 installation instructions to a separate file and linked to it from the core hsmdaemon documentation.

mmattel:
@settermjd pls see my comment in talk

mmattel:
Q: will this PR be backported and in case to which branch?
If there is no BP, we should add the no-backport label...

settermjd:
@mmattel, I believe it's to be backported to 10.1 and 10.0.

modules/admin_manual/pages/configuration/server/security/hsmdaemon/index.adoc
```diff
@@ -0,0 +1,501 @@
...
+====
+
+=== Install a PKCS11 Module
+
+==== Install Using a Preconfigured PKCS11 Module
+
+At least one PKCS11 library is necessary, which is typically provided by an HSM vendor. 
+However, if none is available, you can xref:configuration/server/security/hsmdaemon/install-softhsm2.adoc[install a software HSM (SoftHSM2)].
+SoftHSM2 is the most simple approach to employ storage encryption with a master key in an HSM. 
+This is because it hands off the master key material to a user other than the web server user.
```
pmaier1: That's not the reason for being the "most simple approach". It's a benefit compared to not using an HSM at all. 
=> The point here is that using softhsm is not at all as secure as using an actual **hardware** device. But still there's this benefit that another user on the system is used for the hsmdaemon.

sharidas: SoftHSM2 would only give the upper edge when compared with oC instance without HSM.
How about rephrasing this sentence as:
```
SoftHSM2 is an approach which user could opt, when no HSM hardware is available. It would simulate the same procedure how a hardware device would behave. Like creating key pairs, providing the key to decrypt the file of oC instance when oC user tries to view the file etc.
```

The sentence structure might not be accurate ( pardon me for that ). Let me know if this looks ok?

settermjd: I've been playing around with that paragraph and have come up with the following, based on the existing information. Can you review it and let me know if it's correct?

> SoftHSM2 can also substitute for an HSM when one is not available. For example, when creating key pairs, SoftHSM2 can provide the key to decrypt the file of an ownCloud instance, when the ownCloud user tries to view the file.

I'm not confident on the accuracy of the last sentence.

sharidas: Looks fine to me.

pmaier1: > For example, when creating key pairs, SoftHSM2 can provide the key to decrypt the file of an ownCloud instance, when the ownCloud user tries to view the file.

I don't understand the above. I think the point is:
- Storage encryption with master key on the storage => master key is located on oC storage, next to the actual data, accessible by apache web server user
- Storage encryption with master key in softhsm => master key is located on within softhsm on any storage or machine of your choice, accessible only by softhsm user on that machine
- Storage encryption with master key in an actual HSM => master key is located in a dedicated hardware device, no way to access it

Please verify this with @butonic ^

settermjd: ping @butonic

butonic: Since this is security related documentation half truths help no one. I learned that any security related documentation that does not define threat models is suspicious. So what are we protecting against?
* normal master (or user) key based encryption: files are stored on an external storage service. keys reside in the owncloud server, the storage admin only sees garbage. This scenario actually makes sense if storage and ownCloud services are managed by different people.

* using hsmdaemon (with softhsm2) to store the keys instead of ownCloud itself. Keys are stored with a different owner, preventing apache from directly reading the key material. This might help when someone gains access to the apache user and can read files. It won't help if he can manipulate php files and run arbitrary code. Still ... an improvement. But I find it hard to justify.

* using hsmdaemon with an actual HSM: keys are stored outside the machine, data cannot be accessed when someone steals the complete machine and boots it outside of the corporate network: the hsm is not accessible: files cannot be read. This makes sense.

* You can actually run hsmdaemon with softhsm2 on a different machine in the network, but IIRC traffic is not yet encrypted, so you need to setup your own secure sh tunnel before doing that. This would be the poor mans network attached HSM ... actually this might be a cool blog post. Or we invest in encrypting hsmdaemon traffic. However, I would prefer moving it to reva instead.

Now, regarding the above sentences: there is no difference between softhsm2 and any other HSM. ownCloud encrypts all files with a symmetric file individual key. To read a file that key has to be decrypted using our hsmdaemon. The hsmdeamon uses pkcs11 to talk to whatever hsm driver (module) is configured. Inside the HSM the master key has been generated and It should not leave the HSM, although there might be process reasons to allow that. The real question is what is the threat model and related to that: does the driver / module store the key on the same machine?

Back to the original phrasing:
```
At least one PKCS11 library is necessary, which is typically provided by an HSM vendor. 
However, if none is available, you can xref:configuration/server/security/hsmdaemon/install-softhsm2.adoc[install a software HSM (SoftHSM2)].
SoftHSM2 is the most simple approach to employ storage encryption with a master key in an HSM. 
This is because it hands off the master key material to a user other than the web server user.
```

The first sentence is there to stress that it is not our responsibility to come up with a pkcs11 driver for an HSM. That is the responsibility of the system admin.
The second sentence is intended to point to an example. SoftHSM2 can be used because documentation is freely available. Other HSM vendors tend to make their docs hard to find.
The third and fourth sentence give a reason for using SoftHSM2 over the core encryption, because it is relatively easy to set up and already increases security.

All of it falls under `==== Install Using a Preconfigured PKCS11 Module`

Maybe we should just make a note out of the last three sentences, leaving `At least one PKCS11 library is necessary, which is typically provided by an HSM vendor.` as the main point. Softhsm2 actually only serves an as example...

settermjd:
@butonic, have I understood you correctly with https://github.com/owncloud/docs/pull/802/commits/ffda10b3e8841af310623dcadf7570e62ab493b4?

modules/admin_manual/pages/configuration/server/security/hsmdaemon/softhsm2.adoc
```diff
@@ -1,7 +1,21 @@
-= Install SoftHSM2
+= SoftHSM2
```
mmattel: missing: `:toc: right` directly after the page title

butonic:
@settermjd one more thing:
> When using hsmdaemon with an HSM, the keys are not stored on the same physical machine as ownCloud.
When using hsmdaemon with an HSM, the keys can be stored on a different physical machine as ownCloud.

Sorry for nitpicking: the keys may actually land on the same machine. The HSM might only be used to decrypt them. I think rephrasing with **can** is a good way to describe the situation.

settermjd:
@jfd, with respect to https://github.com/owncloud/docs/pull/802#issuecomment-490005072, please do nit-pick. I'm all for it to ensure 100% accuracy and quality of the documentation.

settermjd:
@butonic, can you review 605db58 pls?

butonic:
@settermjd hm that sounds confusing ... why point out that there is still insecurity left ... 

**When using hsmdaemon with an HSM, the keys can be stored on a different physical machine as ownCloud.**

Sounds more positive, doesn't it? I feel that it conveys the idea of moving the keys elsewhere. If it really happens depends on the implementation details of the HSM. We shouldn't bother going into that, otherwise we might have to go into the details of individual HSM modules and how and where they store key material. I think it still makes sense to give SoftHSM2 as an example later, together with the corresponding threat model.

settermjd:
Okay. That sounds fair to me.

settermjd:
@butonic, change made.

modules/admin_manual/pages/configuration/server/security/hsmdaemon/index.adoc
```diff
@@ -19,9 +19,7 @@ The hsmdaemon is a daemon, provided by ownCloud, to delegate encryption to an {h
 This can be necessary, as PHP cannot, directly, interface with {pkcs11-url}[a PKCS11 stack]; neither with an API wrapper, because one does not exist, nor via the OpenSSL bindings.
 Because of this, a separate process is needed to decrypt anything with the private key stored in an HSM.
 
-When using hsmdaemon with an HSM, the keys are not stored on the same physical machine as ownCloud. 
-As a result, if someone steals the machine that stores your ownCloud instance and boots it outside your network, your data cannot be accessed. 
-As the HSM is not accessible, the keys cannot be read, so they cannot be used to decrypt your data.
+NOTE: When using hsmdaemon with an HSM, the keys _can_ still be stored on the same physical machine as ownCloud.
```
butonic: Maybe I have a wrong understanding of the meaning `can` implies. To me, `can` carries a tone of *hey this cool thing CAN be done*, a positve connotation. In this case, keys being staying on the same machine is a bad thing IMO. Which is why I would like to express more of a *there might still be a problem left*. Something like:
`NOTE: When using hsmdaemon with an HSM, the keys _might_ still be stored on the same physical machine as ownCloud, depending on the HSM implementation.`

But it depends on what we are trying to say in general:
- These are the threat models you can cover with this config, or
- This is how you configure it, but be aware that the following threat models are not covered.

The latter would mean coming up with an endless list of uncovered things. The former means stating a finite number of threat models, maybe ordered by degree of security and providing a config for that. Listing the same config over and over seems redundant, so maybe providing the config steps that allow the admin to reach the minimum degree of security (internal encryption) with added/changed config steps needed to reach the next higher level of security would be a cool thing. Like *congratulations, you are now protected against x*, or in gamification terms *bronze/silver/gold/platinum/... encryption unlocked* ...

Issue #828, opened 2019-03-26
Keep upgraded shipped apps on upgrade (hotfix/patch releases)


PVince81:
Whenever a patch/hotfix release of the core or enterprise bundle is provided, an admin should upgrade to it.

The expectation is that such patch/hotfix release will cause minimal downtime.

The following situation could occur, on offline instances:

1. Admin installed owncloud-enterprise-complete 10.1.0
1. Admin updated non-shipped app "activity" app from 2.4.1 to 2.4.2 using marketplace download
1. Admin updated shipped app "firewall" from 0.7.0 to 0.8.0 using marketplace
1. Admin sets instance to be offline (no more marketplace contact)
1. A hotfix release comes out, owncloud-enterprise-complete 10.1.1. As it's a hotfix, it only provides the same app versions as 10.1.0 did. So firewall is 0.7.0 and activity is 2.4.1.
1. Admin does a [manual upgrade according to docs](https://doc.owncloud.com/server/admin_manual/maintenance/manual_upgrade.html#the-power-user-upgrade). This brings back the old versions of activity and firewall.
1. At this point the database states that we're on activity 2.4.2 and firewall 0.8.0 but as we're offline, there is no way to upgrade. The system is in an inconsistent state.
1. According to the [copy apps steps in doc](https://doc.owncloud.com/server/admin_manual/maintenance/manual_upgrade.html#copy-old-apps), the admin should copy third party apps back into the apps folder.

Expected: the last step should also mention any other upgraded apps.
Actual: the last step only mentions third party apps.

Additionally, it will be difficult for an admin to distinguish what apps were upgraded or not.
Ideally we should move to always recommend the separation of **shipped apps** from **upgraded** and **third party** apps by using [custom app directories](https://doc.owncloud.com/server/10.1/admin_manual/installation/apps_management_installation.html#using-custom-app-directories).

For now I suggest we adjust the documentation to be clear about the need to keep a backup of the updated apps repo. Maybe tell the admin to run `occ apps:list` to find out about app version discrepancies. Sadly the list doesn't help to distinguish between updated apps and not updated ones. Might need some FS scraping...

This needs to be written in a way that leaves **no room for surprises** as it's easy to get confused.

@pmaier1 @individual-it @mmattel @settermjd 

@voroyam FYI

PVince81:
- [ ] tool/command to detect upgraded shipped apps: https://github.com/owncloud/core/issues/32048

PVince81:
[...]

PR #829, opened 2019-03-26
Extend update instructions to be able to preserve updated apps when installing bundles

PVince81:
Fixes https://github.com/owncloud/docs/issues/828

See the ticket for context.

This is to prevent enterprise-complete bundles to overwrite already upgraded versions with older versions, as it could happen with the upcoming 10.1.1 hotfix release.

### DIFF
--- a/modules/admin_manual/pages/maintenance/manual_upgrade.adoc
+++ b/modules/admin_manual/pages/maintenance/manual_upgrade.adoc
@@ -87,26 +87,8 @@ NOTE: Regardless of which approach that you take, they will both assume that you
 `/var/www/owncloud`.
 
 [[the-standard-upgrade]]
-=== The Standard Upgrade
-
-Delete all files and folders in your existing ownCloud directory
-(`/var/www/owncloud`) — *_except_* `data` and `config`.
-
-NOTE: Don’t keep the `apps` directory.
-
-With those files and folders deleted, extract the archive of the latest
-ownCloud server, over the top of your existing installation.
-
-....
-# Extract the .tar.bz2 archive
-tar -jxf owncloud-10.0.3.tar.bz2 -C /var/www/
-
-# Extract the zip archive
-unzip -q owncloud-10.0.3.zip -d /var/www/
-....
-
 [[the-power-user-upgrade]]
-=== The Power User Upgrade
+=== The Standard Upgrade
 
 Rename your current ownCloud directory, for example, from `owncloud` to
 `owncloud-old`. Extract the unpacked ownCloud server directory and its
--- a/modules/admin_manual/pages/maintenance/manual_upgrade.adoc
+++ b/modules/admin_manual/pages/maintenance/manual_upgrade.adoc
@@ -22,7 +22,7 @@ cp -rv /var/www/owncloud /opt/backup/owncloud && mysqldump <db_name> > /opt/back
 == Review Installed Apps
 
 Take note of apps that are currently installed and their versions by
-running `occ apps:list` and saving it for future reference.
+running `occ apps:list` and save them for future reference.
 
 [[review-third-party-apps]]
 == Review Third-Party Apps
@@ -137,12 +137,12 @@ upgrade is finished.
 [[copy-old-updated-apps]]
 == Copy Old Updated Apps
 
-If your setup has a read-only `apps/` folder and is using a separate
-writable folder for updated apps like `apps-external/` or another location,
-please skip this section as well.
+You can skip this section if:
 
-If you are installing the standard owncloud community bundle, usually
-called `owncloud-$version.tar.bz2`, you can skip this section.
+- Your setup has a read-only `apps/` folder and is using a separate
+writable folder for updated apps like `apps-external/` or another location
+- You are installing the standard ownCloud community bundle, usually
+called `owncloud-$version.tar.bz2`
 
 If you are installing an enterprise bundle like `owncloud-enterprise-complete-$version.tar.bz2`,
 please verify that the xref:#review-installed-apps[apps you formerly had installed]
@@ -157,7 +157,7 @@ compared to the formerly installed version, please delete the one that
 came with the bundle and replace it with the one from your old `apps/`
 folder. This will make sure that there are no version discrepancies.
 
-Note that this procedure does not apply for third party apps which you are
+NOTE: this procedure does not apply for third party apps which you are
 recommended to xref:#copy-old-third-party-apps[copy back only after the update].
 
 [[start-the-upgrade]]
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -146,6 +134,32 @@ code) and the Market app is enabled, but there is no available internet
 connection, then these apps will need to be manually updated once the
 upgrade is finished.
 
+[[copy-old-updated-apps]]
+== Copy Old Updated Apps
+
+If your setup has a read-only `apps/` folder and is using a separate
```
mmattel: Why not writing:
```
You can skip this section if:
- Your setup has ...
- You are installing
```

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -146,6 +134,32 @@ code) and the Market app is enabled, but there is no available internet
...
+This could happen for example when installing a patch or hotfix bundle in which
+only minimal changes are contained without recent app updates that
+exist on the marketplace.
+
+For every app that has an older version in the bundle
+compared to the formerly installed version, please delete the one that
+came with the bundle and replace it with the one from your old `apps/`
+folder. This will make sure that there are no version discrepancies.
+
+Note that this procedure does not apply for third party apps which you are
```
mmattel: Better:
```
NOTE: This procedure does 
```
Which renders to a note icon + text

PVince81:
@mmattel updated

mmattel:
There is no commit nor change I can see?

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -18,6 +18,12 @@ First, xref:maintenance/backup.adoc[backup] the following items:
 cp -rv /var/www/owncloud /opt/backup/owncloud && mysqldump <db_name> > /opt/backup/backup-file.sql
 ....
 
+[[review-installed-apps]]
```
settermjd: There's no need to explicitly add a header as one, exactly the same, will be auto-generated.

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -18,6 +18,12 @@ First, xref:maintenance/backup.adoc[backup] the following items:
...
+== Review Installed Apps
+
+Take note of apps that are currently installed and their versions by
+running `occ apps:list` and saving it for future reference.
```
settermjd: s/saving/save/g

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -146,6 +134,32 @@ code) and the Market app is enabled, but there is no available internet
 connection, then these apps will need to be manually updated once the
 upgrade is finished.
 
+[[copy-old-updated-apps]]
```
settermjd: No need to explicitly create an anchor.

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -146,6 +134,32 @@ code) and the Market app is enabled, but there is no available internet
...
+== Copy Old Updated Apps
+
+If your setup has a read-only `apps/` folder and is using a separate
+writable folder for updated apps like `apps-external/` or another location,
+please skip this section as well.
+
+If you are installing the standard owncloud community bundle, usually
```
settermjd: s/owncloud/ownCloud/g

settermjd: Plus, it's likely good to draw attention to this and the paragraph above by using admonitions.

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -162,7 +176,8 @@ The upgrade operation can take anywhere from a few minutes to a few hours, depen
 When it is finished you will see either a success message, or an error message which indicates why the process did not complete successfully.
 
 [[copy-old-apps]]
-== Copy Old Apps
+[[copy-old-third-party-apps]]
```
settermjd: An explicit anchor isn't required.

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
PVince81: I know... for some reason this file already had those so I thought it was an expected pattern

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
PVince81: @settermjd not sure how to do the admonition part. Add an addition NOTE: block at the beginning of the section ? What to write to draw attention ?

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
PVince81: also I think it's nice to keep these even when the title changes to make sure that older links still work. or is there another way ?

PVince81:
I've fixed only the "owncloud" one. I'd leave the anchors items as they were there before.
If we want to discard this pattern then we should go through the whole doc and remove all redundant anchors.

Regarding the extra admonition, please add it yourself @settermjd. Feel free to reword the other sections if needed.

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
settermjd: Some do have them, though it's my beliefe that they should be removed. I believe a significant number of them were auto-generated during the conversion from sphinx-doc with Pandoc. There's nothing wrong with keeping them. It just seemed superfluous (in this instance) if one that's exactly the same would be auto-generated.

PVince81:
@mmattel seems I missed pushing.

@settermjd please check the latter PR

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
settermjd: [Here's the docs for admonitions](https://asciidoctor.org/docs/user-manual/#admonition).

mmattel: or here in a condensed version :smile:
https://github.com/owncloud/docs/blob/master/docs/best-practices.md#admonition

PVince81:
this PR originated to the fact that we intended to bundle old apps with 10.1.1 enterprise complete tarball to avoid breaking semver.

we'll be reworking the bundling concept and what they mean, as the above approach is too complicated for admins.

closing as obsolete

PR #866, opened 2019-04-02
Small script improvement of post-installation-steps.sh

mmattel:
This is a small script update improving overall usage experience.

### DIFF
--- a/modules/admin_manual/examples/installation/post-installation-steps.sh
+++ b/modules/admin_manual/examples/installation/post-installation-steps.sh
@@ -67,20 +67,34 @@ if [ "$do_upgrade" = "y" ]; then
   (echo "$answer" | grep -iq "^y") && upgrdcfg="y" || upgrdcfg="n"
 fi
 
-
 # check if upgrading an existing installation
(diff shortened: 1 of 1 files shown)


modules/admin_manual/examples/installation/post-installation-steps.sh
```diff
@@ -67,20 +67,34 @@ if [ "$do_upgrade" = "y" ]; then
...
-  read -p "Please specify the file to extract with full path: " -r -e answer
+  read -p "Is the instance in maintenance mode? (y/N)? " -r -e answer
+  (echo "$answer" | grep -iq "^y") && mmode="y" || mmode="n"
+  if [ "$mmode" = "n" ]; then
+    echo "Please enable maintenance mode first: sudo -uwww-data ./occ maintenance:mode --on"
+    echo
+    exit
+  fi
+
+  read -p "Please specify the tar file to extract with full path: " -r -e answer
```
settermjd: I appreciate that `answer` is often a common convention for naming the variable that is initialised with the input from the user. However, it seems to make more sense to name it to something like `tarFile`, given the content that it stores, and what it's being used for.

PR #872, opened 2019-04-03
TOC in admin manual - installation

mmattel:
This is the first of a series of PR´s which addresses the use of auto-generated Table of Contents (ToC) for much better readability of the documentation. It also reduces the need of manually creating and maintaining xrefs to do the same.

This first shot covers all documents in administration - installation.

I will do consecutive PR´s (not commits!) because the list of addressed files is big and I want to avoid blockings or merge conflicts.

I have checked each page changed by building the documentation locally.

### DIFF
--- a/modules/admin_manual/pages/installation/apps_management_installation.adoc
+++ b/modules/admin_manual/pages/installation/apps_management_installation.adoc
@@ -1,8 +1,9 @@
 = Installing and Managing Apps
-:toc:
+:toc: right
 
-After installing ownCloud, you may provide added functionality by
-installing applications.
+== Introduction
+
+After installing ownCloud, you may provide added functionality by installing applications.
 
 [[supported-apps]]
 == Supported Apps
--- a/modules/admin_manual/pages/installation/apps_management_installation.adoc
+++ b/modules/admin_manual/pages/installation/apps_management_installation.adoc
@@ -1,4 +1,5 @@
 = Installing and Managing Apps
+:toc:
 
 After installing ownCloud, you may provide added functionality by
 installing applications.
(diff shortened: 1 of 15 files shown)


mmattel:
You raise an interesting point.
Let´s sort that out.
- On the left we have a `site overview`, means which topics do we address in general
- On the right with toc, we have a `page overview`, means which topics are addressed in that particular page

If you clone the branch, let the build run and browse thru the pages, you see a big  positive difference.
The reader gets an overview what the page he is looking at is about and which topics are addressed. Which you can now easily click due to the toc. This is imho a huge advantage compared to the state before which was quite often confusing, especially on bigger pages.

settermjd:
The auto-generated TOC is an advantage, but not when it's so close to the navigation menu, imho.

mmattel:
If we would move the page overview into the site overview, site overview becomes unreadable.
And I doubt that a page toc would be possible in (site) nav.adoc.

settermjd:
Here's an example of what I mean, from [the Mulesoft documentation](https://docs.mulesoft.com/general/).

mulesoft right-hand side table of contents

mmattel:
I got your point.
The screen size for the text to read is already small, adding a toc on the right like in the example above will reduce the reading  size even more. But we can try with a `:toc: right` for those pages where it fits (not suitable when we use a toc as a replacement for an xref-list).
The bigger challenge is to set the toc on the pages and check levels and rendering. If we place in page or right is to be discussed.

settermjd:
Exactly! So this needs a UI update to properly implement.

mmattel:
Will change all `:toc:` --> `:toc: right` to be prepared. When docs-ui allows a right placement, positioning will be done automatically

PR #887, opened 2019-04-06
Fix code errors in MyStorage.php

### LINKED ISSUE #758: Invalid code in MyStorage.php

settermjd:
This fixes #758.


settermjd:
@phil-davis, the docs build pipeline doesn't do it at the moment, but I'd suggest the best thing to do is to run a linter over all the example code files.

modules/developer_manual/examples/app/storage-backend/OCA/MyStorageApp/Storage/MyStorage.php
```diff
@@ -145,12 +145,12 @@ public function fopen($path, $mode) {
...
 				// so we use a temporary file first
-				$ext = pathinfo($filename, PATHINFO_EXTENSION);
+				$ext = pathinfo($path, PATHINFO_EXTENSION);
 				$tmpFile = \OC::$server->getTempManager()->getTemporaryFile($ext);
 
 				// this wrapper will call the callback whenever fclose() was called on the file,
 				// after which we send the file to the library
-				$result = CallbackWrapper::wrap(
+				return CallbackWrapper::wrap(
 					$source,
```
settermjd: @phil-davis, I still have to figure out where this is coming from.

settermjd:
@phil-davis, there are make targets for linting XML, PHP, Yaml, Kotlin, Go, and JSON files, all starting with `validate-`.

PR #891, opened 2019-04-06
Fix omissions in the manual upgrade documentation

### LINKED ISSUE #112: Manual Upgrade Instructions
There are a few problems that can be fixed in these instructions

https://doc.owncloud.org/server/10.0/admin_manual/maintenance/manual_upgrade.html

Under "Enable Maintenance Mode", the admin is instructed to stop the cron service. At no point in the page is the admin instructed when or how to restart the cron service, leaving cron totally disabled.

Under "The Power User Upgrade", the file extraction step is so vague that it never mentions file ownership or permissions. If the admin has extracted or copied the files as any user other than www-data, the remaining instructions on this page will cause file access errors. A good improvement here would be a link to https://doc.owncloud.org/server/10.0/admin_manual/maintenance/update.html#set-updating-permissions-label

Under "Copy Old Apps", the instructions are unnecessarily vague about "3rd party applications". Which apps are 3rd party? Which are not? For example, an earlier section explicitly shows how to disable the activity, files_pdfviewer, files_texteditor, and gallery apps. However, following all the instructions on the page results in a new /owncloud/ directory where these apps are missing, and it is not clear why these apps should be manually copied from /owncloud-old/. I think this just needs to be explained better.

Crosss posted from https://github.com/owncloud/core/issues/32783

settermjd:
This commit fixes #112 by documenting that:

- The Cron service needs to be re-enabled after an upgrade
- The web server user and group needs to own the downloaded archive

### DIFF
--- a/modules/admin_manual/pages/maintenance/manual_upgrade.adoc
+++ b/modules/admin_manual/pages/maintenance/manual_upgrade.adoc
@@ -21,10 +21,16 @@ cp -rv /var/www/owncloud /opt/backup/owncloud && mysqldump <db_name> > /opt/back
 [[review-third-party-apps]]
 == Review Third-Party Apps
 
-Review any installed third-party apps for compatibility with the new
-ownCloud release. Ensure that they are all disabled before beginning the
-upgrade. After the upgrade is complete re-enable any which are
-compatible with the new release.
+Review any installed third-party apps for compatibility with the new ownCloud release. 
+You can find a list of all third-party apps by running the following command, from the root directory of your installation.
+
+[source,console]
+----
+sudo -u www-data ./occ app:list --shipped=false
+----
+
+Ensure that they are all disabled before beginning the upgrade. 
+After the upgrade is complete, re-enable any apps that are compatible with the new release.
 
 WARNING: Install unsupported apps at your own risk.
 
@@ -114,6 +120,10 @@ unzip -q owncloud-10.0.3.zip -d /var/www/
 Rename your current ownCloud directory, for example, from `owncloud` to `owncloud-old`. 
 Extract the unpacked ownCloud server directory and its contents to the location of your original ownCloud installation.
 
+IMPORTANT: These steps must be run as your web server user. 
+Otherwise,  you will likely encounter errors during the upgrade process, and ownCloud will likely not function correctly. 
+For more information, please refer to xref:installation/manual_installation.adoc#set-strong-directory-permissions[the details on setting strong directory permissions].
+
 ....
 # Assumes that the new release was unpacked into /tmp/
 mv /tmp/owncloud /var/www/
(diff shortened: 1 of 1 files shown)


settermjd:
@miqrogroove, please review the changes and provide your feedback?

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -21,10 +21,16 @@ cp -rv /var/www/owncloud /opt/backup/owncloud && mysqldump <db_name> > /opt/back
...
+Review any installed third-party apps for compatibility with the new ownCloud release. 
+You can find a list of all third-party apps by running the following command, from the root directory of your installation.
+
+[source,console]
+----
+sudo -u www-data ./occ app:list --shipped=false
+----
+
+Ensure that they are all disabled before beginning the upgrade. 
+After the upgrade is complete re-enable any which are compatible with the new release.
```
miqrogroove:
```suggestion
After the upgrade is complete, re-enable any apps that are compatible with the new release.
```

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -108,9 +117,12 @@ unzip -q owncloud-10.0.3.zip -d /var/www/
 [[the-power-user-upgrade]]
 === The Power User Upgrade
 
-Rename your current ownCloud directory, for example, from `owncloud` to
-`owncloud-old`. Extract the unpacked ownCloud server directory and its
-contents to the location of your original ownCloud installation. :
+Rename your current ownCloud directory, for example, from `owncloud` to `owncloud-old`. 
+Extract the unpacked ownCloud server directory and its contents to the location of your original ownCloud installation.
+
+IMPORTANT: These steps must either be run as your web server user.
```
miqrogroove: "Either" without "or", likely a typo.

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -164,19 +173,21 @@ When it is finished you will see either a success message, or an error message w
...
-simplest way is by using occ from the command line.
+Assuming your upgrade succeeded, next disable maintenance mode and restart the Cron service. 
+The following example shows how to do both.
 
 ....
+# Disable maintenance mode using the occ command.
 sudo -u www-data php occ maintenance:mode --off
+
+# Restart the Cron service
+sudo service cron stop
```
miqrogroove: Restart or stop?

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -21,10 +21,16 @@ cp -rv /var/www/owncloud /opt/backup/owncloud && mysqldump <db_name> > /opt/back
...
-Review any installed third-party apps for compatibility with the new
-ownCloud release. Ensure that they are all disabled before beginning the
-upgrade. After the upgrade is complete re-enable any which are
-compatible with the new release.
+Review any installed third-party apps for compatibility with the new ownCloud release. 
+You can find a list of all third-party apps by running the following command, from the root directory of your installation.
+
+[source,console]
+----
+sudo -u www-data ./occ app:list --shipped=false
```
mmattel: This is just a info:
`--shipped` does not do what we think is does. Shipped means that apps can/cant be disabled. It does not mean that apps are shipped by the tar file. There is a core pr (https://github.com/owncloud/core/pull/34563) in progress that addresses this issue. When merged, a new PR identifying tared apps needs to be added to `app:list`. Then we have to adopt that command here. For the time being, I would let this as written, better than nothing...

settermjd: Thanks for the clarification, @mmattel.

PR #909, opened 2019-04-08
Update system requirements

settermjd:
This change removes PHP 5.x from the list of supported and recommended PHP versions.

💁‍♂️ - **Only backport to 10.2**

### DIFF
--- a/modules/admin_manual/pages/installation/system_requirements.adoc
+++ b/modules/admin_manual/pages/installation/system_requirements.adoc
@@ -38,7 +38,7 @@ we officially recommend and support:
 | * Apache 2.4 with xref:installation/manual_installation.adoc#multi-processing-module-mpm[`prefork` and `mod_php`]
 
 | PHP Runtime
-| * 7.1, and 7.2
+| * 7.0, 7.1, and 7.2
 |===
 
 [NOTE]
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/installation/system_requirements.adoc
```diff
@@ -38,14 +38,14 @@ we officially recommend and support:
 | * Apache 2.4 with xref:installation/manual_installation.adoc#multi-processing-module-mpm[`prefork` and `mod_php`]
 
 | PHP Runtime
-| * 5.6, 7.0, 7.1, and 7.2
```
pmaier1: 7.0 is still supported.

phil-davis: core 10.2 release drops PHP 5.6 but still supports PHP 7.0

settermjd:
Thanks for catching that oversight.

settermjd:
Is PHP 7.0 still supported? Because https://github.com/owncloud/core/blob/master/composer.json#L35 shows that the minimum PHP version allowed is 7.1. Am I looking at the wrong file?

pmaier1:
> Is PHP 7.0 still supported?

It is. Probably you need to look at stable10 branch.

settermjd:
Fair enough. Just checked and agree.

settermjd:
@pmaier1, 7.0 has been re-added.

PR #908, opened 2019-04-08
Improving how links can be written

mmattel:
This PR enhances the description in docs/best practices how links can be written/used.

PR #933, opened 2019-04-09
[10.0] [PR 908] Improving how links can be written

mmattel:
Backport of PR #908

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -66,6 +66,19 @@ Please see [Troubleshooting Complex URLs](https://asciidoctor.org/docs/user-manu
 This is an example of an URL containing problematic characters which needs special treatment:
 `https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)`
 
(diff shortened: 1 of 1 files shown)


docs/best-practices.md
```diff
@@ -66,6 +66,19 @@ Please see [Troubleshooting Complex URLs](https://asciidoctor.org/docs/user-manu
...
+```
+= The Page Header
+:internal-link-name-url: https://example.com/content/link_can_be_very_long
+
+Text {internal-link-name-url}[highlighted text] text.
+```
+
+It is important that `:internal-link-name:` is placed directly below the page header.
+Any number of these link directives can be added. Without being mandatory, it has turned
+out as a matter of good practice to end your link name with `-url`.
```
settermjd: it would be good, in just a few words, to say why it's turned out to be a good thing.

mmattel: OK will do, but in a new PR, because this is the backport. Pls approve so we can merge.

PR #938, opened 2019-04-09
Revert "[10.1] [PR 909] Update system requirements"

phil-davis:
Reverts owncloud/docs#927

Because PHP 5.6 is still supported by `10.1.*`

Note: the ppa=>PPA text edit can be done as a separate backport if somebody cares.


settermjd:
According to stable10, the minimum version is 7.0.8.

phil-davis:
If you checkout the `10.1` tag, you will see that `10.1` supports PHP 5.6.

Dropping PHP 5.6 only happened on `stable10` branch since `10.1` was released. So dropping PHP 5.6 only applies to the next release from `stable10` = `10.2`

settermjd:
Fair enough. Thanks for the clarification. Will update in just a moment.

PR #953, opened 2019-04-10
[10.1] [PR 877] TOC in admin manual - configuration

mmattel:
Backport of PR #877

PR #952, opened 2019-04-10
[10.0] [PR 877] TOC in admin manual - configuration

mmattel:
Backport of PR #877

PR #954, opened 2019-04-10
fix nav

mmattel:
Fixing a path element in database navigation
No backport needed, clean in 10.0/10.1 via #952 and #953


mmattel:
something bad happened, closing

PR #961, opened 2019-04-10
Copy drone config from other branches to avoid HTML uploads

tboerger:


tboerger:
Issues solved, merging right away.

mmattel:
For better understanding, a copy/paste from Talk:
...
it's still a bad idea to copy the drone config from master branch to release branches. only master branch should upload the generated html at all. the drone configs within the other branches are only there to verify that the docs are getting built... not uploaded
...
Means - **NO FORWARD PORT** to master !

PR #974, opened 2019-04-11
Update user_roles.adoc

mmattel:
Moving text in lists into new lines for better identification of topic and description/details


mmattel:
This screenshot shows the difference between before (left) / after (right) based on two examples.
In my option, this has a much better readability and overview because of splitting the item line into 
major info and additional content.

image

settermjd:
I disagree, as the splits are forced.

mmattel:
> ...as the splits are forced.

Yes they are forced, else you cant do it - do you have a different approach ?
Tell me how you would do it.
The current readbility has room and the need for improvement.
If there is no other suggestion, I would merge this one.
Better this than that.

settermjd:
Sorry, I disagree. I don't see how the new style adds sufficient value. Before we make a change like this, let's discuss it fully, because merging this PR necessitates applying this style across the remainder of the documentation.

PR #876, opened 2019-04-04
[WIP] Link the appliance top-level nav item to the what is it page

settermjd:
This fixes #800. It ensures that the user can find out what the Appliance is as quickly as possible.

PR #994, opened 2019-04-16
Important fix for a wrong link

mmattel:
This is a small but important fix.
Related to merged PR #876 
All open PRs are failing because of missing this fix.
It is necessary to merge this PR and backports immediate.
Then, the failing PRs need to be rebased to be merge ready.

**Note** Drone fails because this issue occurs also in 10.0 and 10.1 (10.2 ?)
Means that at least one of the upcoming backports will fail too.


settermjd:
This will need to be further changed, as that xref can't be backported to other branches, as it's specifically referring to the master branch. Well, it can be backported, but it won't refer to the content in those branches.

settermjd:
Ah, I see that it wasn't backported originally. I'll do it properly this time.

PR #1004, opened 2019-04-17
Move "General Topics" physically to the correct location under configuration

mmattel:
This PR has no content change beside the correction of the headline from `Issues` to `General Topics` as the navigation tells.

It is just about moving all files to the correct location under `configuration` and name the path correctly. Before it was one level up. `General Topics` is part of configuration therefore the old location of the files were wrong.

All corresponding xrefs have been corrected too.

This cleanup process was necessary to have all configuration stuff under one correct directory.


mmattel:
We have already corrected a couple of file locations, file names, but this is the first time you address SEO...

settermjd:
@mmattel, re https://github.com/owncloud/docs/pull/1004#issuecomment-484818014, it's likely the first time that it's occurred to me. It's as good a time as any to start.

mmattel:
Added page aliases for SEO

voroyam:
@settermjd you requested the changes, you should be the one who approves it in the end.

PR #1005, opened 2019-04-17
Fixup --database-pass parameter of acceptance-tests install example

phil-davis:
The --database-pass parameter of acceptance-tests install example does not have a nice format.
I am not sure why it was written like this, maybe just an accident?


mmattel:
I guess backporting to
10.0
10.1
10.2
or?

phil-davis:
Backports:
10.2 #1006 
10.1 #1007 
10.0 #1008

PR #1019, opened 2019-04-18
Update branching workflow to reflect new current version (10.2)

settermjd:


mmattel:
wouldnt it be good at the svg image to remove on the right at the green and yellow circle the last arrow pointing back to master? technically it is correct but we do not forward port and it would highlight much better that the branches have different content.

settermjd:
@mmattel, I see what you mean. I'll update the image accordingly.

PR #1031, opened 2019-04-18
Best Practices - experimental attribute

mmattel:
Small addition that the use of some macros need the `:experimental:` attribute.


tboerger:
It's a directive and not attribute.

mmattel:
@tboerger pls see the antora documenation, they name it attribute, see https://asciidoctor.org/docs/user-manual/#user-interface-macros

PR #1037, opened 2019-04-19
Prevent automatic linking of an URL

mmattel:
Adds the description how a URI will not be rendered as link but as text.
Helpful when creating example URI´s
Prevents false positives for broken link checking...


docs/best-practices.md
```diff
@@ -79,6 +79,18 @@ It is important that `:internal-link-name-url:` is placed directly below the pag
 Any number of these link directives can be added. Without being mandatory, it has turned
 out as a matter of good practice to end your link name with `-url`.
 
+**NOTE** If you want to prevent automatic linking of an URL, prepend it with a backslash (\\).
```
settermjd: s/an URL/a URL/g

docs/best-practices.md
```diff
@@ -79,6 +79,18 @@ It is important that `:internal-link-name-url:` is placed directly below the pag
...
+This will create text but not a clickable link and can be used for example URLs.
+This method helps preventing false positives for broken link checking.
```
settermjd: s/helps preventing/helps prevent/g

docs/best-practices.md
```diff
@@ -79,6 +79,18 @@ It is important that `:internal-link-name-url:` is placed directly below the pag
...
+
+Example:
+
+```
+This is an example web address: \http://example.com
+```
+The above example renders to: `This is an example web address: http://example.com`
```
settermjd: s/renders to/renders as/g

PR #1040, opened 2019-04-20
TOC in user manual - all

mmattel:
Follow up to #872 (TOC in admin manual - installation)

Some text, level and table fixes included
Fixes of header levels where necessary (because of beeing wrong).

This PR TOC´s the complete user manual


mmattel:
Intentional not to backport to 10.0 ?

PR #1065, opened 2019-04-23
updated selinux configuration to reflect apps-external change with 10.2

patrickjahns:
With 10.2 the `apps-external` folder has been made default. 
This pr now changes the selinux configuration to allow http/apache to write to the apps-external folder ( instead of the old apps folder )

### DIFF
--- a/modules/admin_manual/pages/installation/selinux_configuration.adoc
+++ b/modules/admin_manual/pages/installation/selinux_configuration.adoc
@@ -21,7 +21,7 @@ adjust the filepaths in these examples for your installation
 ....
 semanage fcontext -a -t httpd_sys_rw_content_t '/var/www/html/owncloud/data(/.*)?'
 semanage fcontext -a -t httpd_sys_rw_content_t '/var/www/html/owncloud/config(/.*)?'
-semanage fcontext -a -t httpd_sys_rw_content_t '/var/www/html/owncloud/apps(/.*)?'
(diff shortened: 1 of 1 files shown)


mmattel:
@patrickjahns question, shouldn´t it be added instead replaced?
`apps` is still, as far I understood, also for selinux a valid location

settermjd:
@mmattel, your comment in https://github.com/owncloud/docs/pull/1065#issuecomment-486557915 is exactly what I was thinking. @patrickjahns?

patrickjahns:
@mmattel 
Good question - from my understanding no, as the `apps` folder is read only for ownCloud. 
Modifications should only be made to `apps-external` and thus only there the apache/httpd process would need rw permissions

Thinking a bit more of this now, there seems to be then one exception: `templateeditor` - this will try to save the mail templates within the current active theme. Would need to double check which folder it saves the templates, when the 'default' theme is active to ensure template editor works

mmattel:
Pls keep in mind that `apps-external` is automatically set for new installations only. Not when doing an upgrade. Therefore I propose to keep `apps`.

patrickjahns:
Works for me - maybe add a note about this? As you are correct, old installations will not have this folder by default

mmattel:
My very personal view, I would keep `apps` and add `apps-external`. ownCloud defines what using when and we avoid writing additional notes or description text to clarify things. If it would help, I can do a commit for you.

mmattel:
**NOTE:** Backport only to 10.2, not below...

PR #1104, opened 2019-04-30
Improve makefile phony clean comment to make it clear that the pdf path is meant

mmattel:
This PR fixes the path for command `make clean` in makefile: `build/` --> `public/`

**Update:**
I reset the path to the original state according the comment of @tboerger below,
and updated the command description to make it more clear

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ setup: ## Install Antora's command tools locally
 	yarn install
 
 .PHONY: clean
-clean: ## Remove build artifacts from output dir
+clean: ## Remove build artifacts from PDF output dir
 	-rm -rf build/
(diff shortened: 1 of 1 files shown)


tboerger:
Build is the pdf output, so this previous path is totally fine.

mmattel:
@tboerger many thanks for this. I set the path back and improved the comment.

tboerger:
IMHO not needed at all... The Makefile is anyway only used for building the PDFs.

mmattel:
For the sake of public knowledge, we should add that.
You know, I know now, @settermjd may now
but all the others do not know.

It is a comment improvement making things clear

PR #1166, opened 2019-05-08
Globally enable the experimental attribute

### LINKED ISSUE #975: Add :experimental: to the default Antora page settings
For the sake of simplicity, as discussed, I propose that the Antora directive `:experimental:` is added to the defaults.

## Why Should This Change Be Made? (Optional)
This enables the use of macros like `menue` or `btn` or `kbd` 
Else we have to write that directive on each page using that macros

- [x] New Content Addition
- [ ] Old Content Deprecation

- [ ] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [x] Other

settermjd:
This fixes #975. This is an attribute that we've been testing for some time to see if it enhances readability. We believe that it is ready to be used throughout the documentation. So, the attribute's being enabled globally with this change, and any per/page references to it have been
removed.  

💁‍♂ For more information, please see [Antora's experimental attribute documentation].

[Antora's experimental attribute documentation]: https://docs.antora.org/antora/2.0/asciidoc/ui-macros/#set-the-experimental-attribute

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -347,8 +347,6 @@ Please see the reference link for more details.
 Reference: [`Keyboard Shortcuts`](https://asciidoctor.org/docs/user-manual/#keyboard-shortcuts)
 Reference: [`UI buttonss`](https://asciidoctor.org/docs/user-manual/#ui-buttons)
 
-**IMPORTANT** You must set the `:experimental:` attribute to enable the UI macros.
-
 You can create a button styled text like you want a user to press specific keyboard button(s) or browser text buttons.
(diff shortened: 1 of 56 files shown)


mmattel:
How is the impact on ios/android/client/branded?
Do we need to do something there too, eg, deleting all individual page experimental attributes?

If merged, needs an update in best practices guide

settermjd:
I don't know, exactly. But regardless, it would require a PR in those repos similar to this one.

PR #1192, opened 2019-05-10
Update user_roles.adoc

mmattel:
Small fix of an intend of one item


modules/admin_manual/pages/configuration/user/user_roles.adoc
```diff
@@ -11,10 +11,9 @@ high-level overview of each type.
 
 * Is not a regular user.
 * Has access to specific content made available via public links.
-** Can be password-protected +
+** Can be password-protected
 (optional, enforced, policy-enforced).
-+
-- Can have an expiration date +
+** Can have an expiration date +
```
settermjd: Can you please remove the trailing `+`?

mmattel: tbh, no mathew,
you are just looking to a trailing `+` but not to the result. pls see the screenshot below.
all items here are rendered like that scheme and this is a big improvement for readabilitry.
image

settermjd: Actually, I'm asking for the `+` to be removed, as I don't believe that the reflow is natural nor aids comprehension. Additionally, this needs to be properly explored and discussed, as it has implications for the remainder of the documentation.

PR #1217, opened 2019-05-16
deprecate old linux platforms for 10.2 and onwards

jnweiger:
Suggest to remove debian 7, centos 6.9, fedora 27 from the official server list.
Suggest to add CentOS 6.10 and fedora 30.

Not sure if CentOS 6.10 could actually be dropped? It was already removed from packaging and support of CentOS6 needs much help from EPEL repos anyway.


jnweiger:
Maybe also add apache 2.6 as supported? Needs checking, if things would differ from 2.4

pmaier1:
Are you sure that none of them comes with PHP > 7.2?

> Maybe also add apache 2.6 as supported? Needs checking, if things would differ from 2.4

Yeah, should be tested first.

settermjd:
@jnweiger, when you wrote: 

> Not sure if CentOS 6.10 could actually be dropped? It was already removed from packaging and support of CentOS6 needs much help from EPEL repos anyway.

You were referring to CentOS 6.9, yes?

jnweiger:
CentOS6 was dropped from packaging. The SUSE build service cannot distinguish betwenn 6.9 and 6,10 -- we cannot tell which one it actually would support.

settermjd:
@jnweiger, sorry to come back to this so late; is it still right to merge?

jnweiger:
yes, still valid.Better late than never.

PR #1220, opened 2019-05-16
Update ldap documentation

settermjd:


voroyam:
Is it possible to have the toc only for level 1 and 2 headers?

PR #1211, opened 2019-05-15
Document auto-accepting incoming federated shares from trusted servers

settermjd:
This PR updates the Configuring Federation Sharing documentation to add a section documenting how to configure ownCloud to auto-accept incoming federated shares from trusted servers. It fixes #542 and relates to https://github.com/owncloud/core/pull/33980.

PR #1226, opened 2019-05-20
[10.1] [PR 1211] Document auto-accepting incoming federated shares from trusted servers

settermjd:
Backport of PR #1211


settermjd:
@phil-davis, @PVince81 [asked for a backport to 10.1.1+](https://github.com/owncloud/core/pull/33980#pullrequestreview-194673091). Based on that, I inferred that 10.1 was a valid branch to backport to.

phil-davis:
IMO `10.1.1+` means whatever is the next release after `10.1.1`.
Such references are necessary, because at the time it is not known if the next release will be `10.1.2` or `10.2.0`. At the time of merging a PR it is not possible to exactly define what will be the core version that it will be released in.

As far as I can see, this was in the `10.2` changelog testing, so I think it is released in `10.2`

settermjd:
Fair enough. Thanks, again, for spotting my oversight.

PR #1239, opened 2019-05-23
Use source block delimiter consistently throughout the documentation

settermjd:
Perhaps because we were still learning, perhaps because of harmless mistakes, a mixture of [literal and source block delimiters](https://asciidoctor.org/docs/user-manual/#built-in-blocks-summary) have been used in the documentation. This change standardises all of them to use the source delimiter, where applicable.


settermjd:
Thanks, @phil-davis. Will do this as quickly as possible, alongside other requirements/tasks.

settermjd:
I'm going to close and recreate this PR as the time taken to do that will be smaller than that required to fix the conflicts.

PR #1240, opened 2019-05-23
Correct source block attribute substitution configuration

settermjd:
After reading [a tweet](https://twitter.com/mojavelinux/status/1131503689057746944) by @mojavelinux, I see that I made a mistake in 55cf61e, because I didn't add `+` after `attributes`. Consequently, this will only allow attributes to be substituted, but nothing else. This PR rectifies that oversight.


mojavelinux:
You got it!

mmattel:
We should add/update a note to the BPG...

mmattel:
backporting?

settermjd:
Thanks, @mojavelinux!

PR #1242, opened 2019-05-23
updated FAQ section

dercorn:
We had a workshop, where we discussed the current state of the FAQ section and gave it a major overhaul.

### DIFF
--- a/modules/admin_manual/pages/faq/index.adoc
+++ b/modules/admin_manual/pages/faq/index.adoc
@@ -1,30 +1,65 @@
 = Frequently Asked Questions
 :toc: right
 
+[[cannot-see-new-ldap-accounts]]
+== I cannot see new LDAP accounts
+
+Set up a new cronjob for your Apache user to run occ with the user:sync command regularly.
+
+Example for Ubuntu:
+
+....
+crontab -u www-data -e
+....
+
+WARNING: Replace "www-data" with the correct Apache user for your Linux distribution.
+
+Insert
+
+....
+*/6 * * * * /usr/bin/php /var/www/owncloud/occ user:sync -n -m disable -r "OCA\User_LDAP\User_Proxy"
+....
+
+Save and exit. The command will now be executed every ten minutes.
+
 [[i-want-to-upgrade-from-community-version-to-enterprise-version.-what-are-the-changes]]
 == I want to upgrade from Community Version to Enterprise Version. What are the changes?
 
-In ownCloud Enterprise you will get access to new apps and features, mainly targeted towards enterprises; apps and features that ensure security, for example.
+Please contact sales@owncloud.com.
 
-You can upgrade to the Enterprise version without concern, as your existing files, shares, and users remain as they are.
+For a comparison between the editions, see: https://owncloud.com/standard-or-enterprise/
 
-[[how-do-i-transfer-files-from-one-user-to-another]]
-== How do I transfer files from one user to another?
+[[code-integrity-warning]]
+== Code integrity warning
 
-See xref:configuration/files/file_sharing_configuration.adoc#transferring-files-to-another-user[transferring files to another user].
+Please see https://doc.owncloud.com/server/admin_manual/configuration/general_topics/code_signing.html#fixing-invalid-code-integrity-messages
 
-[[how-do-i-deal-with-problems-caused-by-using-self-signed-ssl-certificates]]
-== How do I deal with problems caused by using self-signed SSL certificates?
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/faq/index.adoc
```diff
@@ -1,30 +1,65 @@
 = Frequently Asked Questions
 :toc: right
 
+[[cannot-see-new-ldap-accounts]]
+== I cannot see new LDAP accounts
+
+Set up a new cronjob for your Apache user to run occ with the user:sync command regularly.
```
settermjd: @dercorn, would it be better to say "_web server user_" instead of "_Apache user_"? I appreciate that Apache is the main web server supported by ownCloud; but I'd suggest that it would be better to be vendor-agnostic, as users can use whatever they're prepared to support.

modules/admin_manual/pages/faq/index.adoc
```diff
@@ -1,30 +1,65 @@
...
+
+Insert
+
+....
+*/6 * * * * /usr/bin/php /var/www/owncloud/occ user:sync -n -m disable -r "OCA\User_LDAP\User_Proxy"
+....
+
+Save and exit. The command will now be executed every ten minutes.
+
 [[i-want-to-upgrade-from-community-version-to-enterprise-version.-what-are-the-changes]]
```
settermjd: No need to add an explicit id if the text matches the text in the header. One exactly the same will be automatically generated.

modules/admin_manual/pages/faq/index.adoc
```diff
@@ -1,30 +1,65 @@
...
+[[warnings-in-admin-settings]]
+== Warnings in Admin settings
 
-See xref:configuration/server/occ_command.adoc#security[the security section of the OCC command].
+Please see https://doc.owncloud.com/server/10.0/admin_manual/configuration/server/security_setup_warnings.html
+
+[[why-do-my-uploads-fail]]
+== Why do my uploads fail?
+
+Usually, php may time out when uploading large files. This issue has been resolved with ownCloud 10.0.10. From this version onwards, the web interface also supports https://doc.owncloud.com/server/10.1/admin_manual/release_notes.html#improved-reliability-for-uploads-via-web-interface-on-unreliable-connections[chunked uploads].
```
settermjd: s/php/PHP/g. Secondly, please put each sentence on a separate line.

modules/admin_manual/pages/faq/index.adoc
```diff
@@ -1,30 +1,65 @@
...
-You can upgrade to the Enterprise version without concern, as your existing files, shares, and users remain as they are.
+For a comparison between the editions, see: https://owncloud.com/standard-or-enterprise/
 
-[[how-do-i-transfer-files-from-one-user-to-another]]
-== How do I transfer files from one user to another?
+[[code-integrity-warning]]
+== Code integrity warning
 
-See xref:configuration/files/file_sharing_configuration.adoc#transferring-files-to-another-user[transferring files to another user].
+Please see https://doc.owncloud.com/server/admin_manual/configuration/general_topics/code_signing.html#fixing-invalid-code-integrity-messages
```
mmattel: do not create doc-html links
use xref to adoc

modules/admin_manual/pages/faq/index.adoc
mmattel: xref to adoc

modules/admin_manual/pages/faq/index.adoc
dercorn: @settermjd during recent discussions we decided that we want to narrow down the scope of mentioned alternatives. Yes, it is possible to run ownCloud with nginx and perhaps even lighttpd. But, we want to keep the majority of users on track with a single option, as we simply cannot effectively support all available options out there. Somebody has to actually answer all those support tickets.
Users of nginx and other web server software should know how to handle their stuff, can find answers  at central or should just go with Apache.

modules/admin_manual/pages/faq/index.adoc
dercorn: I see. Honestly, I just took what was already there and modified it. Will try to keep that in mind for the upcoming PRs.

modules/admin_manual/pages/faq/index.adoc
dercorn: Understood.

settermjd:
@dercorn, your branch is based off of 10.1. Can this be ported to master and 10.2?

PR #1263, opened 2019-05-27
Finish deprecating 10.0 branch in 10.2

settermjd:


site.yml
```diff
@@ -10,8 +10,8 @@ content:
   - url: https://github.com/owncloud/docs.git
     branches:
     - master
+    - '10.2'
```
tboerger: That line got to be dropped, 10.2 is already part of the content definition above.

settermjd: Ah, of course. Thanks.

settermjd: @tboerger, done.

PR #1284, opened 2019-05-28
added action descriptions

voroyam:
added action descriptions

log and delete


voroyam:
Those were good changes, as always :)

phil-davis:
IMO this antivirus option works the same when the app is installed on a core 10.0 10.1 or 10.2 server, so this PR should be backported to all 3 doc branches.

PR #1285, opened 2019-05-29
Revert private links details param 10.0

phil-davis:
The `privateLinksDetailsParam` was only added in 10.2
Revert parts of the changes that were backported to `10.0` in PR #1280

PR #1286, opened 2019-05-29
Revert private links details param 10.1

phil-davis:
The `privateLinksDetailsParam` was only added in 10.2
Revert parts of the changes that were backported to `10.1` in PR #1281


phil-davis:
See PR #1285 for this revert in `10.0`

PR #1289, opened 2019-05-29
new picture for mid-sized-deployments

voroyam:
added redis server


voroyam:
> In the PNG, both Webservers are labelled "Webserver 1"
> Is that intended?

no. Good Catch

voroyam:
> reasonable

That's what I've been going for. Reasonable. :)

voroyam:
> LGTM
> Should something be written in the "Components" section about the Redis server?
> Or is it OK that "any reasonably modern hardware/VM will do the job?

We plan to rework the this whole page ... someday. This will be changed hopefully :)

PR #1291, opened 2019-05-29
removed upload gui option

voroyam:


voroyam:
I think in the switch from 9 to 10.

mmattel:
@voroyam when deleting a text block containing links to images, please check if these images are used somwhere else and in case they are not, remove the images too to avoid image orpahns.

Finding and fixing orphand images was a nightmare job when migrating from old to new docs...

voroyam:
Good Point. Will check

settermjd:
@voroyam, any chance you can link to the commit where the change was made?

voroyam:
> @voroyam, any chance you can link to the commit where the change was made?

what commit?

PR #1302, opened 2019-06-05
Update the Ubuntu 18.04 installation example

settermjd:
This change makes a number of changes with the view to making the guide less terse, _yet_ ensuring it remains short, sweet, and straight to the point. 
It:

- Reworked the heading structure
- Updated headers to follow the style guide
- Fixed spelling and grammatical errors
- Updated code examples to be that much simpler to read and to use syntax highlighting, where possible
- Extracted some code samples out into example files, making them easier to maintain and test/lint
- Updated code examples to use sudo, which is the convention on Ubuntu
- Linked to relevant information, both internal and external, to help users of varying skill and knowledge levels.
- Move the navigation entry to the correct position


modules/admin_manual/pages/installation/ubuntu_18_04.adoc
```diff
@@ -1,233 +1,196 @@
...
 ....
 FILE="/usr/local/bin/occ"
 /bin/cat <<EOM >$FILE
 #! /bin/bash
-cd /var/www/owncloud
-sudo -u www-data /usr/bin/php /var/www/owncloud/occ "\$@"
-EOM
 
-chmod +x /usr/local/bin/occ
+cd {install-directory}
```
voroyam: I am not sure we need this as variables. The webserver user will always be www-data as we decided to make the docs "ubuntu only". I am not sure I understand the meaning behind setting a variable for the install directory. Can you explain?

modules/admin_manual/pages/installation/ubuntu_18_04.adoc
```diff
@@ -1,233 +1,196 @@
...
-a2enmod rewrite
-a2enmod headers
-a2enmod env
-a2enmod dir
-a2enmod mime
-a2enmod setenvif
-service apache2 restart
+echo "Enabling Apache Modules"
+
+sudo a2enmod dir env headers mime rewrite setenvif
```
voroyam: have you tested this? :)

voroyam:
looks good. thanks for the changes.

voroyam:
At the end we should add a notion that the installation is finished and ownCloud is accessible at localhost or your configured domain/IP.

voroyam:
Something like "Point your web browser to http://localhost"

modules/admin_manual/pages/installation/ubuntu_18_04.adoc
settermjd: yes.

modules/admin_manual/pages/installation/ubuntu_18_04.adoc
settermjd: The reason for doing so is to avoid writing the same thing multiple times, and potentially introducing bugs or inconsistencies.

settermjd:
@voroyam, added a final note at the end of the doc, as requested.

PR #1304, opened 2019-06-06
reordered installation points

voroyam:
moved deployment recommendations and considerations higher up


settermjd:
@michaelstingl, are you okay with this change?

voroyam:
> @michaelstingl, are you okay with this change?

Michael is on vacation

michaelstingl:
> > @michaelstingl, are you okay with this change?

yes, makes sense

> Michael is on vacation

*was last week

PR #1327, opened 2019-06-18
Change default version to 10.2

### LINKED ISSUE #632: Change the default version of the documentation to come from 10.1
The default version of the documentation that’s loaded needs to come from the branch that’s tracking the latest release of ownCloud, which is currently `10.1`, not from `master`. As the default version of the documentation that users, currently see is master, which is a floating version, not tied to a specific ownCloud version. This is likely to cause confusion and to be misleading.

- [ ] New Content Addition
- [ ] Old Content Deprecation

- [ ] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

settermjd:
To change default version of the docs to 10.2, this PR:

- Removed the local working copy from the default playbook (site.yml)
- Removed the relevant references to master branch through the documentation
- Created a new playbook for testing local changes, and added two new yarn commands for using that playbook to build the docs and to also validate the local changes
- Updated the branching workflow to reflect the changes

💁‍♂ A comprehensive redirect for all requests to the master branch version of the docs to 10.2 is required, but outside the scope of this PR. That will need admin assistance to complete.

### Fixes

#632.


tboerger:
If you really want to remove the master part you should drop the whole branch and make the 10.2 the default branch instead of adding this local job stuff again.

settermjd:
I respectfully disagree. Having master as the ever-rolling branch and ultimate source of truth and branching off of that makes sense. This is because:

- The approach uses de facto conventions that are widely understood. 
- It's both clear and intuitive as to where each version of the documentation can be found, and, conversely, what each branch should contain. 
- Having two local jobs shouldn't add so much extra work as to be challenging for contributors. Plus, the target names are intuitive.

settermjd:
This PR requires https://gitea.owncloud.services/devops/websites/pulls/1 to be merged, before it can be considered fully complete.

tboerger:
But the docs won't be rendered anywhere, or it doesn't make sense to render them at all.

settermjd:
> But the docs won't be rendered anywhere, or it doesn't make sense to render them at all.

I don't understand what you're saying. Can you clarify, pls?

tboerger:
The master branch won't be accessible/published anywhere because of redirects within Nginx. So IMHO the master branch doesn't make any sense to me anymore.

settermjd:
So, @tboerger, you'd be okay with merging this PR then?

settermjd:
Scrap what I just said.

settermjd:
Closing, as this is superseded by 10.3.

Issue #1328, opened 2019-06-19
Sharing API claims that permission cannot be changed for a public link share


phil-davis:
## What Needs to be Documented?
Remove the note in the sharing API docs:
```
Also, a permission cannot be changed for a public link share.
```

## Where Does This Need To Be Documented?
 modules/developer_manual/pages/core/ocs-share-api.adoc

## Why Should This Change Be Made? (Optional)
We have acceptance tests that demonstrate changing the permissions of a public link share. So it is possible. And in discussion with others, it should be possible (it is actually a requirement)
So the note seems to be an error from the past.

## What Type Of Content Change Is This? (Optional)
- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [x] Bug Fix to Existing Content

## Which Manual Does This Relate To? (Optional)
- [ ] Admin Manual
- [x] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

phil-davis:
PR to fix #1329

micbar:
@settermjd Please check If we need this in other docs too.

settermjd:
I've grepped the documentation and don't find any references to this string. If it is still there, in another form, please reopen this issue. Otherwise, closing.

PR #1329, opened 2019-06-19
Remove text about public link share permissions

phil-davis:
Remove note:
```
Also, a permission cannot be changed for a public link share.
```

Issue #1328 

The note is not true.


phil-davis:
This note has always been incorrect, so this fix could be backported to all published doc branches.

PR #1337, opened 2019-06-23
Misleading general owncloud.log path

tiagofreire-pt:
This cause fail2ban errors.


voroyam:
I would change ownCloud Server to ownCloud Appliance.

ownCloud Server is our docker image, and it has no such paths.

voroyam:
> > I would change ownCloud Server to ownCloud Appliance.
> > ownCloud Server is our docker image, and it has no such paths.
> 
> I agree, but the document starts stating that it assumes the user is running the Server.

Because this document is not exclusively for the appliance. You, however, want to make a change **exclusively** for the appliance. The default owncloud server has his log path in /owncloud/data, the univention appliance has the path you stated. That's why if you make such a change, or a change at all to get fail2ban working in the appliance, you have to state this fact.


I would change `owncloud_data` to `owncloud/data`.

Never understood why we are assuming that users change their data directory to this from the start.

modules/admin_manual/pages/configuration/server/harden_server.adoc
```diff
@@ -330,7 +330,8 @@ protocol = tcp
 filter = owncloud
 maxretry = 3
 bantime = 10800
-logpath = /var/owncloud_data/owncloud.log
+# If running 'Owncloud Server', try instead: /var/lib/univention-appcenter/apps/owncloud/data/files/owncloud.log
```
tiagofreire-pt: # If running 'Owncloud Appliance', try instead: /var/lib/univention-appcenter/apps/owncloud/data/files/owncloud.log

tiagofreire-pt: # If you're running 'Owncloud Appliance', try instead: /var/lib/univention-appcenter/apps/owncloud/data/files/owncloud.log

voroyam: yes. that looks appropriate.

voroyam:
Document still unchanged.

Will add the changes myself, since the author is kinda absent.

tiagofreire-pt:
> 
> 
> Document still unchanged.
> 
> Will add the changes myself, since the author is kinda absent.

Thanks for the extra hand on this! :)

PR #1357, opened 2019-06-27
Fix broken PDF generation and docs build process

settermjd:
This PR updates the docs so that the PDF generation works again, and in so doing, stops the build from breaking on the `docs-pdf` step.

The reason that the build is breaking is partially visible in the excerpt below:

```console
asciidoctor: ERROR: ownCloud_Admin_Manual.adoc: line 157: include file not found: /drone/src/modules/admin_manual/pages/issues/code_signing.adoc
asciidoctor: ERROR: ownCloud_Admin_Manual.adoc: line 158: include file not found: /drone/src/modules/admin_manual/pages/issues/general_troubleshooting.adoc
asciidoctor: ERROR: ownCloud_Admin_Manual.adoc: line 159: include file not found: /drone/src/modules/admin_manual/pages/issues/impersonate_users.adoc
asciidoctor: WARNING: image to embed not found or not readable: /drone/src/modules/admin_manual/assets/images/installation/deprecs-2.png
Prawn::Errors::CannotFit
  Use --trace for backtrace
Makefile:91: recipe for target 'pdf-admin' failed
make: *** [pdf-admin] Error 1
```

On closer inspection, the reason why the files are not able to be found is for two reasons:

1. Because the PDF configuration file was not kept in sync with the Antora navigation file, as content was restructured and renamed. 
2. [asciidoctor-pdf](https://asciidoctor.org/docs/asciidoctor-pdf/) only understands the older `partialsdir`, `examplesdir`, & `imagesdir` include syntax, not the newer, Antora, `partial$`, & `example$`, etc.

This PR refactors the build process for the PDF manuals, so that the content for the PDF manuals is generated from the Antora navigation file and reverts the documentation to use the older AsciiDoc include syntax.


books/ownCloud_Admin_Manual.adoc
```diff
@@ -8,25 +8,32 @@ The ownCloud Team <docs@owncloud.com>
 :toclevels: 2
 :icons: font
 :icon-set: octicon
+:example$: modules/admin_manual/examples
```
tboerger: This is already set within the makefile, so it should be dropped.

settermjd: It was an experiment that didn't work.

settermjd:
@tboerger, I've made a host of changes and have tested them as working locally. Can you have a look at them and add your review/feedback, please?

tboerger:
```
sed 's|xref:\(.*\)\[.*\]|include::\1[leveloffset=+3]|'
```

settermjd:
@tboerger, what about the following?

```bash
#!/bin/bash

filename="$1"

while read line 
do
    if [[ ${line} =~ \]$ ]]; then 
        level_offset=$(echo "$line" | awk -F"*" '{print NF-1}')
        revised_line=$(echo "$line" | sed 's/xref:/include::{module_base_path}/' | sed 's/\[.*\]//g' | sed -r 's/^\*{1,} //')
        echo "${revised_line}[leveloffset=+${level_offset}]"
    fi
done < "${filename}"
```

settermjd:
@tboerger, pls review https://github.com/owncloud/docs/pull/1357/commits/1afb1c95c78dc866e071556af6237f43888c254a, time permitting.

settermjd:
I need help with 1afb1c95c78dc866e071556af6237f43888c254a.

.gitignore
```diff
@@ -2,3 +2,4 @@ cache/
 node_modules/
 public/
 build/
+vendor/
```
tboerger: This is obsolete

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
+#!/bin/bash
```
tboerger:
```
#!/usr/bin/env bash
```

is best practice, especially for mac since the shipped bash is pretty old.

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+
+set -o noclobber
```
tboerger: What about `set -e`?

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+    if [[ -n $DRONE_TAG ]]; then
+        VERSION=${DRONE_TAG/v//}
+    else
+        [[ -n $DRONE_BRANCH ]] && VERSION=${DRONE_BRANCH/v//} || VERSION=master
+    fi
+fi
+
+function usage()
+{
+    echo "Usage: cmd [-c] [-b <admin|developer|user>] [-v <go|xml|json|php|kotlin|yaml>] -r <revision> -d <release date>"
```
tboerger: Since you anyway only print `cli` i would suggest to rename the script to `bin/cli`.

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+    echo "...build directory cleaned."
+}
+
+function validate_code_files()
+{
+    language="$1"
+
+    case $language in
+        go)
+            command -v golint &>/dev/null && \
```
tboerger: What about printing an error if the command is not available?

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+                echo "Validating go source files" && \
+                find . -type f -name "*.go" \
+                    ! -path "./node_modules/*" \
+                    ! -path "**/vendor/*" \
+                    ! -path "./.git/*" \
+                    -exec sh -c 'echo Linting {} && golint {} && echo' \;
+            ;;
+
+        json)
+            command -v jsonlint &>/dev/null && \
```
tboerger: What about printing an error if the command is not available?

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+                find . -type f -name "*.json" \
+                    ! -path "./node_modules/*" \
+                    ! -path "**/vendor/*" \
+                    ! -path "./.git/*" \
+                    -exec sh -c 'echo Linting {} && jsonlint -qp {} && echo' \;
+            ;;
+
+        kotlin)
+            command -v ktlint &>/dev/null && \
```
tboerger: What about printing an error if the command is not available?

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+                ktlint --reporter=plain "./modules/*_manual/**/*.kt" || true;
+            ;;
+
+        php)
+            command -v php &>/dev/null && \
```
tboerger: What about printing an error if the command is not available?

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+                find ./modules/*_manual/examples -type f -name "*.php" -exec php -l {} \;
+            ;;
+
+        xml)
+            command -v xmllint &>/dev/null && \
```
tboerger: What about printing an error if the command is not available?

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+                find ./modules/*_manual/examples -type f -name "*.xml" -exec xmllint --noout {} \;
+            ;;
+
+        yaml)
+            command -v yamllint &>/dev/null && \
```
tboerger: What about printing an error if the command is not available?

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+            ;;
+    esac
+}
+
+function convert_antora_nav_to_asciidoc_list()
+{
+    filename="$1"
+
+    while read line 
+    do
```
tboerger: You put `if` on a single line but not this loop?

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+function build_pdf_manual()
+{
+    manual="$1"
+    release_date="$2"
+    revision="$3"
+    build_directory="$(pwd)/build/server/${revision}/${manual}_manual/"
+    manual_infix="$(tr '[:lower:]' '[:upper:]' <<< ${manual:0:1})${manual:1}"
+    book_file="books/ownCloud_${manual_infix}_Manual.adoc" 
+    nav_file="modules/${manual}_manual/nav.adoc"
+    input_file="books/catfile.$$.adoc"
```
tboerger: Why `catfile` and not `ownCloud_${manual_infix}_Manual`?

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+
+    if [[ $manual != "admin" && $manual != "developer" && $manual != "user" ]]; then
+        echo "That is not a valid manual" 
+        exit $ERR_UNSUPPORTED_MANUAL
+    fi
+
+    echo "Generating version '${revision}' of the ${manual} manual, dated: ${release_date}"
+
+    [[ ! -e "$build_directory" ]] && mkdir -p "$build_directory"
```
tboerger: For `mkdir -p` you can drop the condition.

bin/cli.sh
```diff
@@ -0,0 +1,177 @@
...
+function clean_build_dir()
+{
+    echo "Cleaning build directory..."
+    rm -rvf "build/"
+    echo "...build directory cleaned."
+}
+
+function validate_code_files()
+{
+    language="$1"
```
tboerger: Variables within functions should be defined as `local`.

.gitignore
settermjd: Thanks for spotting that. Something I overlooked when removing the PHP code.

bin/cli.sh
settermjd: Thanks for the recommendation.

bin/cli.sh
settermjd: Thanks.

bin/cli.sh
settermjd: I didn't go in to too much detail, given time was of the essence. However, fine by me to add error output.

bin/cli.sh
settermjd: If it's a better approach to do that here, fine by me.

bin/cli.sh
settermjd: Good question.

bin/cli.sh
settermjd: Would it not make sense to only create the directory if it didn't already exist?

bin/cli.sh
settermjd: That's what I was forgetting. Thanks.

bin/cli.sh
tboerger: `mkdir -p` only creates it if it doesn't exist and is simply a no-op if it already exists.

tboerger:
The file is still named `bin/cli.sh` instead of `bin/cli` :)

settermjd:
I didn't think [changing the name](https://github.com/owncloud/docs/pull/1357#issuecomment-508017396) was mandatory, but have done that in the most recent, revised, commit.

tboerger:
Now you have dropped the whole shell script :(

settermjd:
I was a touch hasty and didn't add the renamed file after the `git rm`. It's there now.

bin/cli.sh
settermjd: Thanks for the clarification.

settermjd:
The final problem with the PDF is in the admin manual. Not sure which image is causing the problem, yet. And the feedback from the community's been helpful, but not conclusive.

tboerger:
I was able to pin down the issue for failing PDF builds to a malicious table definition which you can find at https://github.com/owncloud/docs/blob/fix-broken-pdf-generation/modules/admin_manual/pages/configuration/server/antivirus_configuration.adoc#available-configuration-settings, if I drop the section it builds fine.

settermjd:
@tboerger, I found the same thing. Turns out, that the cause of the error wasn't an image, which I and some in the community originally thought that it might be, but a table that had column widths defined for _some_ — but not all — columns. The error and stacktrace were NOT helpful in debugging the problem. This isn't an issue with asciidoctor-pdf, but with Prawn, which asciidoctor-pdf is built on top of. So, it's an indirect asciidoctor-pdf problem.

settermjd:
And the build PASSES!

PR #1362, opened 2019-06-27
Update windows-network-drive_configuration.adoc

voroyam:
reworked the document

### DIFF
--- a/modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
+++ b/modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
@@ -1,16 +1,21 @@
 = Installing and Configuring the External Storage: Windows Network Drives App
-:toc: right
+:toc: macro
+:toclevels: 1
+:anacron-examples: http://www.thegeekstuff.com/2011/05/anacron-examples
+:flock-docs-url: https://linux.die.net/man/2/flock
+:shell-flock-intro: https://linuxaria.com/howto/linux-shell-introduction-to-flock
 :msft-security-bulletin-ms17-010-url: https://docs.microsoft.com/en-us/security-updates/SecurityBulletins/2017/ms17-010
 :samba-478-url: https://www.samba.org/samba/history/samba-4.7.8.html
-:samba-481-url: https://www.samba.org/samba/history/samba-4.8.1.html 
+:samba-481-url: https://www.samba.org/samba/history/samba-4.8.1.html
 :samba-url: https://www.samba.org/
 :smb2-url: https://en.wikipedia.org/wiki/Server_Message_Block#SMB_2.0
+:smbclient-manpage-url: https://www.samba.org/samba/docs/man/manpages-3/smbclient.1.html
 :wannacry-ransomware-attack-url: https://en.wikipedia.org/wiki/WannaCry_ransomware_attack
 
 == Introduction
 
 The https://marketplace.owncloud.com/apps/windows_network_drive[External Storage: Windows Network Drives app]
-creates a control panel in your Admin page for seamlessly integrating Windows and Samba/CIFS shared network 
+creates a control panel in your Admin page for seamlessly integrating Windows and Samba/CIFS shared network
 drives as external storages.
 
 Any Windows file share and Samba servers on Linux and other Unix-type
@@ -29,7 +34,7 @@ the configuration.
 
 image:enterprise/external_storage/wnd-1.png[Windows Network Drive share on your Files page]
 
-TIP: For more information on SMB/CIFS in ownCloud, refer to xref:configuration/files/external_storage/smb.adoc[the Samba file server configuration documentation]. 
+TIP: For more information on SMB/CIFS in ownCloud, refer to xref:configuration/files/external_storage/smb.adoc[the Samba file server configuration documentation].
 
 Files are synchronized bi-directionally, and you can create, upload, and
 delete files and folders. ownCloud server admins can create Windows
@@ -47,13 +52,13 @@ which adds security.
 == Installation
 
 Install the https://marketplace.owncloud.com/apps/windows_network_drive[External Storage: Windows Network Drives app]
-from the ownCloud Market App or ownCloud Marketplace. 
+from the ownCloud Market App or ownCloud Marketplace.
 To make it work, a few  dependencies have to be installed.
 
-* A Samba client. This is included in all Linux distributions. 
+* A Samba client. This is included in all Linux distributions.
 On Debian, Ubuntu, and other Debian derivatives it is called `smbclient`.
 On SUSE, Red Hat, CentOS, and other Red Hat derivatives it is `samba-client`.
-* `php-smbclient` (version 0.8.0+). It should be included in most Linux distributions. 
+* `php-smbclient` (version 0.8.0+). It should be included in most Linux distributions.
 You can use https://github.com/eduardok/libsmbclient-php[eduardok/libsmbclient-php],
 if your distribution does not provide it.
 * `which` and `stdbuf`. These should be included in most Linux distributions.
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
```diff
@@ -1,16 +1,17 @@
 = Installing and Configuring the External Storage: Windows Network Drives App
-:toc: right
+:toc: macro
```
settermjd: Why change the style from `right` to `macro`? With this change, the table of contents no longer renders in the page.

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
```diff
@@ -348,14 +288,179 @@ processed.
...
+[Unit]
+Description=ownCloud WND Listener for SERVER SHARE
+After=syslog.target
+After=network.target
+Requires=apache2.service
+[Service]
+User=www-data
+Group=www-data
+WorkingDirectory=/var/www/owncloud
+ExecStart=/usr/bin/php ./occ wnd:listen -vvv SERVER SHARE USER PASSWORD
```
jvillafanez: There are better ways to set the password for the wnd:listen. The one shown here is a quick and dirty way to set the password, but I don't think we should recommend this

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
```diff
@@ -348,14 +288,179 @@ processed.
 As said, you can wrap that command in a Cron job so it’s run every 5
 minutes for example.
 
+=== WND Listener Configuration
+
+Create a service following the instructions below that checks the share for changes:
```
jvillafanez: I think we should remark this is for systemd. This is recommended (if systemd is being used) but technically not needed.

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
```diff
@@ -348,14 +288,179 @@ processed.
...
+sudo systemctl enable owncloud-wnd-listen-SERVER-SHARE.service
+sudo systemctl start  owncloud-wnd-listen-SERVER-SHARE.service
+----
+
+=== WND Process Queue Configuration
+
+Create or add a `crontab` file in `/etc/cron.d/oc-wnd-process-queue`.
+
+* Make a `crontab` entry to run a script iterating over all `SERVER SHARE` pairs with
+an appropriate `occ wnd:process-queue` command. The commands must be **strictly sequential**.
```
jvillafanez: Just to clarify, if the process-queue overlaps with another execution of the same process-queue, this might be problematic due to possible race condition. However, you can run multiple process-queue as long as the host and share are different in all of them.
Basically, you should have one process-queue running per host and share.

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
```diff
@@ -348,14 +288,179 @@ processed.
...
+This can be done by using `flock -n` and tuning the `-c` parameter of `occ wnd:process-queue`
+
+....
+* * * * *     sudo -u www-data /usr/bin/php /var/www/owncloud/occ wnd:process-queue <HOST> <SHARE>
+....
+
+
+=== Serializer
```
jvillafanez: I was confused with this header. Maybe "execution serialization" fits better.

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
```diff
@@ -348,14 +288,179 @@ processed.
...
+
+Parallel runs of `wnd:process-queue` might lead to a user lockout.
```
jvillafanez: I'd skip this paragrapth. If the process-queue is set with a cron job, it's extremely weird that a race condition in the specific point that would trigger this problem will happen. It's weird it happens if 2 processes start at the same time, so it's even more weird if it happens if the second process starts with more than 1 minute of delay.

The main reason to serialize the execution is to ensure a predictable order of the notifications.

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
```diff
@@ -348,14 +288,179 @@ processed.
...
+
+Take the example of attempting to connect to the share named MyData
+using `occ wnd:listen`. Running the following command would work:
+
+[source,console,subs="attributes+"]
+....
+{occ-command-example-prefix} wnd:listen MyHost MyData svc_owncloud password
+....
+
+However, running this command would not:
```
jvillafanez: it might be better to remark that the command is case sensitive, and that it must match the information from the mount point configuration

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
voroyam: I got it from the occ command, did not rly know why.

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
voroyam: remarked

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
voroyam: Changed

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
voroyam: from where to where do you want the text to be omitted?

voroyam: @jvillafanez

jvillafanez:
```
Parallel runs of `wnd:process-queue` might lead to a user lockout.
The reason for this is that several `wnd:process-queue` might use the same wrong password because it hasn't been updated by the time they fetch it.
```

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
voroyam: Can you then tell me the better ways?

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
voroyam: done

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
voroyam: removed

modules/admin_manual/pages/enterprise/external_storage/windows-network-drive_configuration.adoc
settermjd: @jvillafanez, do you have further feedback?

voroyam: I suspect a variable has to be defined.

jvillafanez: https://github.com/owncloud/windows_network_drive/pull/112 should contain the relevant information. https://github.com/owncloud/windows_network_drive/pull/129 also contains additional options related to password handling.

jvillafanez:
```
occ wnd:listen host share username --password-file=/tmp/mypass --password-trim
```
probably is the easiest alternative. Just change the file location and make sure that file contains the password.

voroyam: added

voroyam:
@jvillafanez if everything is correct now, I would like to merge this

PR #1370, opened 2019-07-02
added app:enable encryption

voroyam:
as a redundancy I added the command  app:enable encryption for users who only look for the code block.

If the app is enabled, the command does no harm.

If the app is not enabled, the enabling of the app is important for the following commands.


settermjd:
I rebased master on this branch to fix the breaking build.

PR #1371, opened 2019-07-02
Link to the new iOS Mobile App on the docs root page

settermjd:
This follows up a conversation with @michaelstingl, earlier today.


settermjd:
@michaelstingl, that's a question for @tboerger. AFAIK, no. thanks for the 👍

settermjd:
FWIW, we should just about have fixed the PDF build failure.

PR #1392, opened 2019-07-05
Update repo link for owncloud-files

enbrnz:
The link is pointing to an old repository which would install version 10.1.0 and is confusing as it shows '10.0' in the link.
So users assume that they could use '10.2' in there to update and install version 10.2 of owncloud-files, but that repository doesn't work.


settermjd:
Thanks for reporting this, @enbrnz!

PR #1411, opened 2019-07-09
Fixed the upgrade docs

voroyam:
The docs were "refined" after a knowledge sharing session but something(tm) went horribly wrong.

The new content was simply copied above the old content, creating a loop upgrade like: "first back up your files, then enable maintenance mode, .... first back up your files, then enable maintenance mode, ..."

This document is huge, so it took a lot of work to fix this, because some parts of the old document were hasty copy pasted, missing the proper context. Like if you have 2 sentences, and the second one references the subject of the first one with "it" and you copy just the second sentence - it does not make any sense.

Also the work from @settermjd was applied only to the old document, not to the new one, especially the reformatting of the code blocks, so I had to pull them from the old document in the new one.

I did took this effort because I work with the docs, and need them to be correct. I could easily clone the docs and keep my own copy updated, would be less work.

I hope that in the future, we could all take more care in maintaining the docs, so that we improve them and not make more work for other people.

### DIFF
--- a/modules/admin_manual/pages/maintenance/manual_upgrade.adoc
+++ b/modules/admin_manual/pages/maintenance/manual_upgrade.adoc
@@ -1,11 +1,12 @@
 = Manual ownCloud Upgrade
 :toc: right
+:toclevels: 1
 
-== Introduction
+== Preparation
 
 This section describes how to manually upgrade your ownCloud installation.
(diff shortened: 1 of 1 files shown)


voroyam:
@settermjd can you back port this?

PR #1427, opened 2019-07-10
Update finalise-install.sh example

enbrnz:
The example wasn't using the scripts parameters in a crucial step. Making it unnecessarily complicated to adopt the script with different variables, as script-users would have to adjust the webserver-user at several positions in the script even though we define a webserver-user at the beginning of the script.


phil-davis:
@settermjd will need forward port to `master` (and maybe backport to `10.01` etc?)

voroyam:
In my view we don't need the variables. Too much confusion. We would need them if we would have a docs with multiple distributions. But our docs is ubuntu/apache/mariadb only.

PR #1437, opened 2019-07-11
redis address correction

voroyam:
the ip address in the example was machine specific. replaced it with the loop back address.

### DIFF
--- a/modules/admin_manual/pages/configuration/server/occ_command.adoc
+++ b/modules/admin_manual/pages/configuration/server/occ_command.adoc
@@ -510,10 +510,10 @@ Adding Redis to the configuration:
 ----
 {occ-command-example-prefix} config:system:set \
    redis \
-   --value '{"host": "{oc-examples-server-url}", "port": "{std-port-redis}"}' \
+   --value '{"host": "{oc-examples-server-ip}", "port": "6379"}' \
    --type json
 
-System config value redis set to json {"host": "{oc-examples-server-url}", "port": "{std-port-redis}"}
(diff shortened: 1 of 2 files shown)


settermjd:
@voroyam, are you happy with the additional changes that I've made?

voroyam:
> Looks like a good time to use two of the site-wide attributes: `oc-examples-server-url` and `std-port-redis` (in site.yml), instead of hard-coded values.

would you mind linking to the site.yml or at least writing here what the values of those variables are so people don't have to do it themselves?

If you are changing the page, you are doing that anyway, right?

The site.yml is a long page with small text, so reading through the entire bottom half with long variable names is not a pleasure that I want to repeat.

voroyam:
> @voroyam, are you happy with the additional changes that I've made?

Not really. The redis port won't change anytime soon, so I don't see the point behind replacing 4 didgets with a longer variable name.

The host has to be the IP-address, not the server URL. At least this is my understanding.

settermjd:
> > @voroyam, are you happy with the additional changes that I've made?
> 
> Not really. The redis port won't change anytime soon, so I don't see the point behind replacing 4 didgets with a longer variable name.
> 
> The host has to be the IP-address, not the server URL. At least this is my understanding.

Thanks for the clarification.

voroyam:
LGTM

voroyam:
@settermjd can you review this and merge?

settermjd:
Just created PRs to backport it to 10.2, 10.1, & 10.0. All should be merged in the next 30 mins.

PR #1457, opened 2019-07-15
Fix self-referencing page alias

settermjd:
This PR fixes a self-referencing page alias.


settermjd:
Created unnecessarily.

PR #1468, opened 2019-07-16
 Standardize 'public link share' words

phil-davis:
IMO we has decided to call these "public link shares", so make the documentation consistent.

- change any "public share link" to "public link share"
- change any "public share" to "public link share"

Preparation for/sort of related to issue #1328 

May as well get this consistent first, then touch some place(s) that need content clarified.


phil-davis:
I rebased and it automatically dismisses the review! @settermjd please review again.

phil-davis:
Backports:
`10.2` #1518 
`10.1` #1519 
`10.0` #1520

PR #1470, opened 2019-07-16
removed non apache2 instructions

voroyam:
Since we (the ownCloud Team) decided that we want to keep the docs Apache / maria db / Ubuntu exclusive  I removed the instructions for cent OS or RHEL or SUSE Linux from the manual installation guide.

The Full Guide shall be transferred to central, where it can be maintained by the community.

This PR also fixes the code block formatting error with the docs and the minuses.


voroyam:
> Perhaps, instead of removing the information entirely, it could be moved to a "supplemental" section.

After a long conversation in the ownCloud team, we came to the conclusion that it would make sense to have the official docs limited to what we support/test on. If we keep extra stuff somewhere, users / customers can complain / refer to those pages and make us responsible for them.

Also it was not removed, it was moved here: 

https://central.owncloud.org/t/doc-manual-installation/20999/2

PR #1489, opened 2019-07-18
add decryption example

voroyam:
User Key decryption

added variable for password

added command to use that variable in decrypting files

added commands to remove the entries of encryption from database and the encryption keys from storage

### DIFF
--- a/modules/admin_manual/pages/configuration/files/encryption/encryption_configuration_quick_guide.adoc
+++ b/modules/admin_manual/pages/configuration/files/encryption/encryption_configuration_quick_guide.adoc
@@ -102,6 +102,9 @@ They need to:
 
 === Decrypt Encrypted Files
 
+If you have an instance with a few users, you can use this example to decrypt the files.
+Note that you have to enter the password for each user manually.
+
 [source,console,subs="attributes+"]
 ----
 {occ-command-example-prefix} maintenance:singleuser --on
@@ -109,6 +112,18 @@ They need to:
 #enter **Recovery Key** for **each user**
 
 # Recovery Key is a password set by the admin
+{occ-command-example-prefix} maintenance:singleuser --off
+----
+
+If you have a large instance with many users, use this to decrypt the files:
+
+* Set the variable as export `OC_RECOVERY_PASSWORD=1111`, then run this set of commands:
+(Replace "1111" with your actual Recovery Key)
+
+[source,console,subs="attributes+"]
+----
+{occ-command-example-prefix} maintenance:singleuser --on
+sudo -E -u www-data php occ encryption:decrypt-all -m recovery -c yes
 {occ-command-example-prefix} maintenance:singleuser --off
 ----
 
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/configuration/files/encryption/encryption_configuration_quick_guide.adoc
```diff
@@ -112,6 +115,18 @@ They need to:
 {occ-command-example-prefix} maintenance:singleuser --off
 ----
 
+If you have a large instance with many users, use this to decrypt the files:
+
+* Set the variable as export `OC_RECOVERY_PASSWORD=1111`, then run this set of commands:
```
pako81: Example on how to export it? Like `export OC_RECOVERY_PASSWORD=1111 `

modules/admin_manual/pages/configuration/files/encryption/encryption_configuration_quick_guide.adoc
```diff
@@ -121,3 +136,24 @@ They need to:
 # ignore the "already disabled" message
 {occ-command-example-prefix} app:disable encryption
 ----
+
+=== Cleanup your database
+
+Access your database, select ownCloud and remove the remaining entries
```
pako81: "access your ownCloud database" sounds better

modules/admin_manual/pages/configuration/files/encryption/encryption_configuration_quick_guide.adoc
```diff
@@ -102,6 +102,9 @@ They need to:
 
 === Decrypt Encrypted Files
 
+If you have an instance with a few users, you can use this example to decrypt the files.
+Note that you have to enter the password for each user manually.
```
pako81: I would add "the ownCloud admin must be sure all users already enabled the recovery password option in their personal settings page"

modules/admin_manual/pages/configuration/files/encryption/encryption_configuration_quick_guide.adoc
```diff
@@ -102,6 +102,9 @@ They need to:
...
+
```
pako81: I would specify that when running `encryption:decrypt-all` the recovery-key option must be chosen and not the user password one.

modules/admin_manual/pages/configuration/files/encryption/encryption_configuration_quick_guide.adoc
voroyam: added

modules/admin_manual/pages/configuration/files/encryption/encryption_configuration_quick_guide.adoc
voroyam: specified

modules/admin_manual/pages/configuration/files/encryption/encryption_configuration_quick_guide.adoc
voroyam: example added

modules/admin_manual/pages/configuration/files/encryption/encryption_configuration_quick_guide.adoc
voroyam: reframed

voroyam:
I agree on 80% on your suggestions @pako81 

The titles describe what the following commands do.

After
`Decrypt Encrypted Files`
the command
`occ encryption:decrypt-all`
literally decrypts the encrypted files.

After
`Deactivate User-Specific Key-based Encryption`
the commands
`occ encryption:disable`
`occ app:disable encryption`
Deactivate the encryption and the encryption app.

In my view these are accurate titles for what is in the section. Also the titles are the same for both encryption modules, so the titles are representing a structure.
Like Masterkey - enable - decrypt - disable.
Then User key - enable - decrypt - disable.

Can you explain to me what the benefit in your view would be in changing the titles?

pako81:
yeah, you are probably right, titles can stay like this. Then let's merge it :)

Issue #250, opened 2018-11-06
knowledgebaseenabled parameter description


mrlinux1:
https://doc.owncloud.org/server/10.0/admin_manual/configuration/server/knowledgebase_configuration.html

This option seems to work only for community edition. And it seems to be a legacy thing.

Then there is a problem to write " you have to use it to disable ..." 
but showing "knowledgebaseenabled" => true.

And please change "knowledgebaseenabled" to 'knowledgebaseenabled'.

voroyam:
https://github.com/owncloud/docs/pull/1536

legacy content removed.

PR #1536, opened 2019-07-23
removed legacy content

voroyam:
removed legacy content

inspired by:

https://github.com/owncloud/docs/issues/250


tboerger:
This option have been removed?

voroyam:
> This option have been removed?

Apparently this never worked in enterprise and sometimes in community, but since no one we know has this option, make sense to remove it from the docs.

tboerger:
I guess we got it within the base docker image, need to check that

voroyam:
> I guess we got it within the base docker image, need to check that

Have you ever used that? :)

tboerger:
We offer the option via env variable if it's documented within the sample config on core repository :)

PR #1558, opened 2019-07-25
Add a GitHub settings yaml file to the project

settermjd:
This change aims to simplify and better manage the labels in the docs repo. It fixes #1241.

PR #1594, opened 2019-07-31
[10.0] [PR 1558] Add a GitHub settings yaml file to the project

settermjd:
Backport of PR #1558


tboerger:
Why do you backport that at all? It doesn't make sense to put that into other branches than master.

settermjd:
Good point.

PR #1596, opened 2019-07-31
Fix TOC title in admin PDF manual

### LINKED ISSUE #1568: TOC in PDF manuals broken
The Table of Content (TOC) in the PDF manuals are broken. In particular in the navigation pane of the PDF viewer.

- There is text that doesn't belong in the TOC
- All entries are under the introduction
- The granularity of the entries is questionable (i.e. hundreds of pages between two TOC lines). That does not make it feasible to use the TOC to navigate a document

E.g. in https://doc.owncloud.com/server/10.2/admin_manual/ownCloud_Admin_Manual.pdf

image

settermjd:
This change removes the toc-title attribute from user_roles.adoc which, in a PDF context, overrides the manual's TOC title value, set in the book configuration file. This fixes #1568.

### DIFF
--- a/books/ownCloud_Admin_Manual.adoc
+++ b/books/ownCloud_Admin_Manual.adoc
@@ -9,3 +9,4 @@ The ownCloud Team <docs@owncloud.com>
 :icons: font
 :icon-set: octicon
 :module_base_path: modules/admin_manual/pages/
+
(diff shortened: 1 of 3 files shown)


mmattel:
Before you remove the `toc-title`, check the impact.
`toc-title` has always a corresponding `toc::[]`. especially on the occ command page, the toc is not on top but for valid reasons in the document. See https://doc.owncloud.com/server/10.2/admin_manual/configuration/server/occ_command.html

settermjd:
@mmattel, thanks for the suggestion. Unfortunately, it has to be removed as it adversely affects the generation of the PDF manual.

PR #1604, opened 2019-08-05
Add script to check for orphaned example files

settermjd:
This PR adds a small shell script that checks if any of the example files (scripts) is used anywhere throughout the documentation.

### DIFF
--- a/bin/find-orphaned-files.sh
+++ b/bin/find-orphaned-files.sh
@@ -8,6 +8,7 @@
 MANUALS=( admin developer user )
 
 echo "Checking for orphaned example files."
+echo "Please run this scrip from your docs root"
 echo
 
 for manual in "${MANUALS[@]}"; do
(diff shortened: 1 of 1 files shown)


bin/find-orphaned-files.sh
```diff
@@ -0,0 +1,24 @@
...
+
+echo "Checking for orphaned example files."
+echo
+
+for manual in "${MANUALS[@]}"; do
+    [[ ! -e "./modules/${manual}_manual/" ]] && echo "Cannot find ${manual} manual." && continue
+
+    echo "Checking the ${manual} manual"
+    find "./modules/${manual}_manual/examples/" \
+        \( -path "**/.DS_Store" -o -path "**/vendor" -o -path "**/.idea" -o -path "**/.gitkeep" \) -prune -o \
```
tboerger: `.idea` should never be part of some sub folder, so IMHO you could drop that.

mmattel: Generally correct, in this find statement, it is part of excluded directories to be searched in.
Whyever a directory named .idea is present, exclude it in the search.
Therefore it is ok to me.

tboerger: .idea folders are project folders by intellij, which would never be in a sub directory.

mmattel: fixed (removed)

mmattel:
@settermjd 
I have removed `.idea` according to @tboerger comment above.
Additionally I added some text to improve usability.

PR #1623, opened 2019-08-09
[10.0] [PR 1617] Improving Best Practices

mmattel:
Backport of PR #1617

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -240,6 +240,26 @@ include::example$installation/post-installation-steps.sh[]
 ----
 ```
 
+### OCC Examples
+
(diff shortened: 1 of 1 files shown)


docs/best-practices.md
```diff
@@ -240,6 +240,26 @@ include::example$installation/post-installation-steps.sh[]
 ----
 ```
 
+### OCC Examples
+
+When creating examples that show how to use occ, ensure that you use the `occ-command-example-prefix` attribute.
+Doing so will keep all examples of its use consistent throughout the documentation.
+
+**Note:** when used within a source code block, as in the following example, `subs="attributes"` has to be set, otherwise it won't render properly:
```
settermjd: `subs="attributes"` is incorrect, as the example below shows. It should be `subs="attributes+"`

PR #1639, opened 2019-08-12
add notion for guest users

voroyam:
Guest users have no home storage. But we don't have a notion written in the docs that an error message will appear during files:scan


voroyam:
@settermjd is the wording correct?

settermjd:
@voroyam, thanks for making this change! I just extended it ever so slightly.

settermjd:
@voroyam, is this change applicable to 10.0 or 10.1?

voroyam:
yes

PR #1650, opened 2019-08-14
Document new options for adding users via provisioning API and occ

### LINKED ISSUE #1643: Document new options to add users via API and occ

- Server 10.3 will offer new ways to invite users by mail via provisioning API and occ. 
- https://github.com/owncloud/enterprise/issues/2831

- Provisioning API: Dev docs
- OCC: Admin docs

- [x] New Content Addition
- [ ] Old Content Deprecation

- [x] Admin Manual
- [x] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

settermjd:
This change, updates the occ and provisioning API documentation to cover that when creating users, either one or both of a username and password can be supplied. If one of these options is supplied, it makes the other optional. More specifically; if an email address is supplied and a password isn't, that a password will be automatically generated and the user will be emailed a link to change their password.

This fixes #1643.

### DIFF
--- a/modules/admin_manual/pages/configuration/user/user_provisioning_api.adoc
+++ b/modules/admin_manual/pages/configuration/user/user_provisioning_api.adoc
@@ -30,18 +30,37 @@ sending a basic HTTP authentication header.
 ==== Syntax
 
 [cols=",,",options="header",]
-|==============================================
+|===
 | Request Path | Method | Content Type
 | `ocs/v1.php/cloud/users` | `POST` | `text/plain`
-|==============================================
+|===
 
-[cols=",,",options="header",]
-|========================================================
-| Argument | Type | Description
-| userid | string | The required username for the new user
-| password | string | The required password for the new user
-| groups | array | Groups to add the user to [optional]
-|========================================================
+[cols="15%,15%,70%",options="header",]
+|===
+| Argument 
+| Type 
+| Description
+
+| `email` 
+| string 
+a| The email address for the user (_optional_).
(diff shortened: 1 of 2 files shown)


modules/admin_manual/pages/configuration/user/user_provisioning_api.adoc
```diff
@@ -51,18 +70,41 @@ sending a basic HTTP authentication header.
...
+[source,console]
+----
+curl -X POST http://admin:secret@example.com/ocs/v1.php/cloud/users \
+   -d userid="Frank" \
+   -d email="frank@example.com" \
+   -d password="frankspassword" \
+   -d groups[]="finance" -d groups[]="management"
+----
+
+.Create the user "Frank" with email address "frank@example.com" and add him to the "finance" and "management" groups. A temporary password will be created and the user will be emailed with a link to change the password.
```
sharidas: Doesn't the previous example and this one look same? At least from the behaviour wise. If you agree with me then just remove the `-d password` line, then it would add more info. It would help users know that without providing password, the user can be created and with the help of email user can set the password.

settermjd: Thanks for drawing my attention to this. I'll check it out.

settermjd:
@sharidas, just removed the extra reference.

settermjd:
If you're happy with it, please give it a 👍

PR #1651, opened 2019-08-14
document migration step commands

voroyam:
If users / customer missed one or several migration steps and are not facing issues, they can manually run those steps with commands documented here.

### DIFF
--- a/modules/admin_manual/pages/configuration/server/occ_command.adoc
+++ b/modules/admin_manual/pages/configuration/server/occ_command.adoc
@@ -3319,7 +3319,33 @@ Example:
 {occ-command-example-prefix} notifications:generate -g Office "Emergency Alert" "Rebooting in 5min"
 ----
 
-== ownCloud Maintained Application Commands
+=== Migration Steps Command
+
+You can run migration steps with the `migrations` command.
+
+[source,console,subs="attributes+"]
+----
+{occ-command-example-prefix} migrations:execute <app> <version>
+----
+
+===== Arguments
+
+[width="100%",cols="20%,70%",]
+|===
+| `app`  | Name of the app this migration command shall work on.
+| `version`  | The version to execute.
+|===
+
+===== Example
+
+This example executes the migration step for the core app:
+
+[source,console,subs="attributes+"]
+----
+{occ-command-example-prefix} migrations:execute core 20181220085457
+----
+
+== Apps Commands
 
 This command reference covers the ownCloud maintained apps commands.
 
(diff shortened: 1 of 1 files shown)


pako81:
yes, of course, there is. Just compare https://github.com/owncloud/core/tree/master/core/Migrations (for the required oC version) with the content of the `oc_migrations` table and you will get the missing ones (if any).

Those above are only the ones for core but there are other migrations related to i.e. `activity`, `dav`, `files_sharing` etc.

pako81:
But to be honest, I am not sure if we want to document this. How this is documented right now it doesn't anyway tell that much since you would need to know what that specific migration is doing.

Which means retrieve all possible migrations, which are spread in code across core and several apps, and compare them with the content of your `oc_migrations` table. And I do not expect a "normal" user to be able/willing to do this.

voroyam:
> But to be honest, I am not sure if we want to document this. How this is documented right now it doesn't anyway tell that much since you would need to know what that specific migration is doing.
> 
> Which means retrieve all possible migrations, which are spread in code across core and several apps, and compare them with the content of your `oc_migrations` table. And I do not expect a "normal" user to be able/willing to do this.

Well, my motivation for the documenting this was so that support people to know the command. Starting a VM and trying something with `occ migra..` until you get the right command takes longer than opening up docs and searching for it in the docs page for occ commands.

I know this is far from perfect but it's definitely a first step. And a step towards a better state of things is better than staying in the same place imo.

Also we should not keep secret commands, but document all that we have, as these commands are available for people to see, but if we don't tell them what they are for = secret knowledge.

modules/admin_manual/pages/configuration/server/occ_command.adoc
```diff
@@ -3319,7 +3319,33 @@ Example:
...
+===== Example
+
+This example executes the migration step for the core app:
+
+[source,console,subs="attributes+"]
+----
+{occ-command-example-prefix} migrations:execute core 20181220085457
+----
+
+== Apps Commands
```
settermjd: Is this header still necessary?

voroyam: why not?

settermjd: How's it different to https://github.com/owncloud/docs/pull/1651/files#diff-2818bba1033f9f6297d7e1437dd4370eR153?

voroyam: why not just write the words? It's just an "s".

Instead of this, you do https://en.wikipedia.org/wiki/Cut,_copy,_and_paste and make people https://en.wikipedia.org/wiki/Reading and https://en.wikipedia.org/wiki/Research

But to the matter at hand: App commands sounds right to me, because what follows are the commands ( many of them ) to an app ( one at the time )

Apps commands sounds chaotic at first, and then it would also assume that what follows are commands for multiple apps.

App commands sounds clean. That's why I changed it. If you feel otherwise please explain.

PR #1674, opened 2019-08-20
Document recommendations for using NFS storage

### LINKED ISSUE #256: Document recommendations for using NFS storage
There can be different scenarios where ownCloud's storage is located on a NFS mount (primary/secondary). Also in some scenarios multiple application servers can use the same NFS mount.

We should document some basic examples/recommendations of mounting NFS to the underlying OS to be used with oC. Probably this is basic NFS mounting but I think it makes a lot of sense to give admins some confidence that they don't have to think of additional options or similar.

Furthermore we should give some recommendations for advanced setups (e.g. several instances connecting to the same NFS mount) when there are special options to set (e.g. noac).

settermjd:
This creates an initial set of deployment recommendations for using NFS with ownCloud. It fixes #256.


modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
```diff
@@ -0,0 +1,146 @@
...
+=== _netdev
+
+Use this option to ensure that the NFS storage is available before ownCloud starts.
+It is essential when MySQL’s data files are located on the NFS storage, as MySQL could error if the connection to the storage is not ready when ownCloud starts.
+
+TIP: Alternatively, you can use {autofs-url}[autofs] to ensure that mounts are always available.
+
+=== soft
+
+We recommend this option to better handle storage failures.
```
cdamken: Why do we recommend that?, An Administrator has to know how the system should be configured.

IMHO we should say: Soft mounts are easier to administrate and Hard, if you don't have enough experience this could be your option.

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
```diff
@@ -0,0 +1,146 @@
...
+If `hard` is used, turning off the ownCloud server is likely to be problematic, and you need a boot and shutdown sequence in the servers.
+
+=== retrans
+
+Use this option to set the number of minor timeouts and retransmissions before major timeouts happen.
+Major timeouts either abort file operations or display the "server not responding" message.
+We recommend that you set it to ?.
```
cdamken: To "?" ?

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+=== timeo
+
+Use this option, set in tenths of a second, to set the time before an initial retransmission after a timeout.
+
+According to {nfs-man-page-url}[the NFS man page], better overall performance may be achieved by increasing the timeout when mounting on a busy network, to a slow server, or through several routers or gateways. 
+We recommend that you set it to ?.
```
cdamken: To "?" ?

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
```diff
@@ -0,0 +1,146 @@
...
+The smaller the size, the greater the number of packets need to be sent to send or receive a file.
+Conversely, the larger the size, the fewer the number of packets need to be sent to send or receive a file.
+
+You can find this information by looking at the `/proc/mounts`, as in the example below.
+
+[source,console]
+----
+cat /proc/mounts | grep -i owncloud
+
+192.168.0.103:/data /mnt/owncloud nfs rw,vers=3,rsize=32768,wsize=32768,hard,proto=tcp,timeo=600,retrans=2,sec=sys,addr=192.168.0.103 0 0
```
cdamken: About the rsize and wsize  you have to mount first the storage withi a big number and test with:
(as example:)
```
#root@snoopy3:~# mount | grep /mnt/owncloud | egrep -o rsize=[0-9]*
rsize=65536
#root@snoopy3:~# mount | grep /mnt/owncloud | egrep -o wsize=[0-9]*
wsize=65536
```
The value that is shown there is the maximun transfer rate.

About the sixth field here is  0, it should be 2 because if the value is set to zero, the device or partition will be excluded from fsck check and if it is nonzero the fsck check will be run in the order in which the value is set. The mount partition is not a root partition

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+Once you've determined the best sizes, set them by passing the (`rsize` and `wsize`) options when mounting the share or in the share's mount configuration.
+
+.Specifying the read and write block sizes when calling mount
+[source, console]
+----
+mount 192.168.0.104:/data  /mnt -o rsize=65536,wsize=65536
+----
+
+=== Tune the MTU (Maximum Transmission Unit) Size
```
cdamken: Playing with MTU could generate no connection to the other computer, and depends from the devices between them, 1500 is the normal or Jumbo Packages, but IMHO this can be modified if you really know what you are doing. otherwise, you end with a lot of lost packages.

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
settermjd: Thanks for spotting that.

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
settermjd: I'll add a note to this section.

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
```diff
@@ -0,0 +1,146 @@
...
+* *Improved Security:* It {nfs-strong-security-architecture-url}[mandates a strong security architecture]. It does not require {rpc-statd-url}[rpc.statd] or {lockd-url}[lockd]. As a result, it only uses port 2049.
+* *Improved Reliability:* Uses TCP by default instead of UDP.
+* *Improved Performance:* It uses Multi-Component Messages, which reduce network traffic. It is capable of using a 32KB page size, compared to the default, 1024 bytes.
+* Use of {nfs-read-write-delegations-url}[Read/Write Delegations].
+
+== Tune the Mount Parameters
+
+=== _netdev
+
+Use this option to ensure that the NFS storage is available before ownCloud starts.
```
mmattel: Better: `Use this option to ensure that the network is up and running before NFS tries to connect.`
ownCloud does not start. It is reactive...
Maybe you want to write: `File and database services need to be up and running before accessing ownCloud.`

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
```diff
@@ -0,0 +1,146 @@
...
+2: wlan0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
+----
+
+.Retrieve interface MTU size with cat
+[source,console]
+----
+cat /sys/class/net/<interface>/mtu
+----
+
+To change the MTU size, under Ubuntu, specify the MTU size, per interface, in `/etc/network/interfaces`, as in the following example, and then restart networking.
```
mmattel: Starting with Ubuntu 18.04 this file/path is not longer used.
The correct file for configuring interfaces is now: `/etc/netplan/50-cloud-init.yaml`
Be adviced, that the content is now in yaml format and needs careful intend formatting !!
You will find examples by searching the internet.

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
settermjd: Actually, https://www.networkworld.com/article/2224654/mtu-size-issues.html makes for an excellent discussion on the topic of MTU sizes.

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
settermjd: I just re-read the setting definition, and I see that you're correct, @mmattel.

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
mmattel: First, the connection between server and storage should be non-routed and wired/switched Gigabit.
Second, add a note to check that packages do not get fragmented between server and storage by using the dont fragment option pinging the other side and adopting the mtu value in case of fragments.
Third, any optimisation for a mtu value like 9000 (Jumbo Frames) requires a storage backend being capable saturating gigabit which only high end enterprise storage systems can provide. (technical background on request).

mmattel: additional note regarding jumbo frames: why do you try to fix an issue on the ys level (network) while your storage runs on a ms level (mechanical, if running hdd)... the gain on high end and optimized storage systems is less then 8%

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
settermjd: Thanks. I wasn't aware of the change.

settermjd: From further reading of the NetworkManager and Netplan documentation, it doesn't appear that setting the MTU is an altogether straight-forward process.

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
settermjd: I've linked to that article in the changes that I'm pushing shortly.

settermjd: @mmattel, that seems like overkill here. I've linked to the NetworkWorld article which gives an excellent discussion about potential MTU issues.

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
settermjd: Is there a recommended `timeo` value?

modules/admin_manual/pages/installation/deployment_recommendations/nfs.adoc
settermjd: I believe that that was in the notes in #256.

settermjd: I based it on the following, [which you wrote](https://github.com/owncloud/docs/issues/256#issuecomment-411718997).

> Soft-mount is recommended in case that a storage failure happens. otherwise turning off the owncloud server is more problematic with a hard mount and you need a boot and shutdown sequence in the servers.

mmattel:
@settermjd I have some ideas for this document but can start to add a proposal earliest tomorrow afternoon or friday

settermjd:
@cdamken, can you add further feedback to my questions, pls?

mmattel:
To answer the outstanding questions above, see http://manpages.ubuntu.com/manpages/trusty/man5/nfs.5.html

**soft**: default hard, no change needed (*), option bg is important in conjunction of hard 
**retrans**: default 3, no change needed, important for soft
**timeo**: default 600, no change needed

- Please see also the example of the mount options in the document
- (*) Please also see linked document https://www.netapp.com/us/media/tr-4067.pdf page 112
Quote: For business-critical NFS exports, NetApp recommends using hard mounts. NetApp strongly
discourages the use of soft mounts. (They are THE nfs storage provider)

Option **bg** is important to overcome "hickups" when rebooting the storage backend system or temporarily loss of network connection while continuing server operation without manual intervention when things are up again.

I have updated the document accordingly and cleaned up the nfs mount options section and squashed it.

settermjd:
Thanks for the changes and for the extra information. However, please **never** force push when multiple people are collaborating on a branch. Please only use, [force-with-lease](https://estl.tech/a-gentler-force-push-on-git-force-with-lease-fb15701218df?gi=d2a9b4a988a4), if required.

settermjd:
> `/drone/src/fonts/mplus1mn-regular-ascii-conums.ttf not found` ??

I started fixing that yesterday. Aiming to fix it today.

PR #1675, opened 2019-08-20
Update file_sharing_configuration.adoc

jnweiger:
make it clear that `Restrict users to only share with users in their groups`  does not restict sharing with groups at all.

Hope that this is correct :-) 
Is this set of options defined in the feature files?


phil-davis:
e.g. https://github.com/owncloud/core/blob/master/tests/acceptance/features/webUIRestrictSharing/restrictSharing.feature

```
  Scenario: Restrict users to only share with users in their groups
    Given the setting "Restrict users to only share with users in their groups" in the section "Sharing" has been enabled
```
Yes, the user can still share with any group. But when they want to share with an individual user/users then they can only choose other users that are members of groups that the user is a member of.

```
  Scenario: Restrict users to only share with groups they are member of
    Given the setting "Restrict users to only share with groups they are member of" in the section "Sharing" has been enabled
```
Also correct, this setting stops users from sharing with any group. They can only share with groups that they are a member of.

modules/admin_manual/pages/configuration/files/file_sharing_configuration.adoc
```diff
@@ -119,6 +119,8 @@ e.g. read-only to avoid creating shares with too many permissions unintentionall
 ==== Restrict users to only share with users in their groups
 
 Check this option to confine sharing within group memberships.
```
phil-davis: Maybe this sentence could also be revised to be similar style to the one in the next section.
```suggestion
When this option is enabled, users can only share with other users that are members of groups that the user is a member of.
```

jnweiger:
TODO: We also need to simplify the strings in the admin interface.

pmaier1:
> TODO: We also need to simplify the strings in the admin interface.

I thought about the wording of these two options thoroughly. IMO, if you read closely, it is very clear. The option itself is a little complex but I don't think this will be improved by wording changes.

modules/admin_manual/pages/configuration/files/file_sharing_configuration.adoc
```diff
@@ -118,7 +118,9 @@ e.g. read-only to avoid creating shares with too many permissions unintentionall
 
 ==== Restrict users to only share with users in their groups
 
-Check this option to confine sharing within group memberships.
+When this option is enabled, users can only share with other users that are members of groups that the user is a member of.
+They can still share with all groups of the instance but not with users that do not share a group.
```
settermjd: Given the setting's wording, this option seems counter-intuitive, i.e., either they can **_only_** share with users in their groups, or they can do that, and share with other groups.

pmaier1: The difference here is sharing with **single** users or with **groups of users**. There are use cases where admins want to allow users to share with all single users in their instance but only with groups they are a member of. Or they should be restricted to share only with users in their groups but are allowed to share with all groups (e.g., without knowing exactly who is a member of such group). This is why this option has been split up some time ago.
If you want to restrict users to _really_ only be able to share within their groups, you need to tick both options.

phil-davis: @settermjd it is like @pmaier1 said. The challenge is how to describe it in English that is correct, understandable and brief - maybe all 3 cannot be achieved together. Suggestions are most welcome.

settermjd: Thanks @pmaier1 @phil-davis. You're right, @phil-davis, wording is a little tricky. I'll see if I can come up with some meaningful suggestions.

settermjd: Here's a mockup of a small rework of the two checkboxes.
The two checkboxes are now listed under an applicable sub-heading, inside a "Restrictions" group.

new user and group restrictions wording

The intent here is to:

1. Reduce the number of required words
2. More intuitively show what the settings are about; and
3. Leave the door open to further restrictions in the future

Yes/no?

settermjd: @phil-davis @jnweiger any feedback?

settermjd:
@phil-davis, can you add further feedback, pls?

modules/admin_manual/pages/configuration/files/file_sharing_configuration.adoc
jnweiger: "Users can only share with users in their group" is much better! Thank you!
But I still see some room for misinterpretation. 
All of short, precise, and still correct and easy english is hard.

How about mentioning group shares in the positive form, like in e.g. 
"Users can only share with users in their groups or with groups"

That makes the english a bit harder and longer, but (hopefully) adds precision.

I'd leave the final balancing to you. :-)

settermjd: Thanks for the feedback, @jnweiger. I agree, conciseness, correctness, and precision in any language is hard (English is no friend in that respect). That said, I'll get it done.

settermjd: Here's one final suggestion: "_Users can only share with the users in the groups which they are a member of_". I've been contemplating this for a while now, this morning, and I can't think of anything better. @jnweiger, would you be happy with it?

settermjd:
ping @phil-davis @jnweiger

pmaier1:
The changes to file_sharing_configuration.adoc look good :+1: 
I don't understand your intention for the changes in manual_upgrade.adoc.

settermjd:
> The changes to file_sharing_configuration.adoc look good 👍
> I don't understand your intention for the changes in manual_upgrade.adoc.

As the commit mentioned, it was attempting to get a list of core/supported apps. Should it be removed?

pmaier1:
> As the commit mentioned, it was attempting to get a list of core/supported apps. Should it be removed?

Ok, looks strange to me tbh, would remove it.

settermjd:
> > As the commit mentioned, it was attempting to get a list of core/supported apps. Should it be removed?
> 
> Ok, looks strange to me tbh, would remove it.

There was a lot of discussion around it, as to whether it was the right approach. I'll remove it shortly.

modules/admin_manual/pages/configuration/files/file_sharing_configuration.adoc
```diff
@@ -118,7 +118,9 @@ e.g. read-only to avoid creating shares with too many permissions unintentionall
 
 ==== Restrict users to only share with users in their groups
 
-Check this option to confine sharing within group memberships.
+When this option is enabled, users can only share with other users that are members of groups that the user is a member of.
```
phil-davis:
```suggestion
When this option is enabled, users can only share with the users in the groups which they are a member of.
```
As suggested by @settermjd in the thread below.
IMO this reads better.

modules/admin_manual/pages/configuration/files/file_sharing_configuration.adoc
phil-davis: I am happy with @settermjd suggestion. I made a comment on the exact line where it would apply.

modules/admin_manual/pages/maintenance/manual_upgrade.adoc
```diff
@@ -60,6 +60,20 @@ Review any installed third-party apps for compatibility with the new ownCloud re
 Ensure that they are all disabled before beginning the upgrade.
 Third party apps are all apps that are not distributed by https://marketplace.owncloud.com/publishers/owncloud[ownCloud]
 
+[TIP]
```
phil-davis: What happened here?
This commit has nothing to do with the topic of this PR.

settermjd: Odd. I'll get it removed.

settermjd: On reviewing it further, I have no idea why I committed that with this PR.

phil-davis:
IMO this should be ported forward to 10.3 and master branches)
(and maybe back to 10.1 ?)

PR #1701, opened 2019-08-22
Create app_settings.adoc

voroyam:
we now have app settings in the appliance which configure options so users don't have to acess the command line.


voroyam:
@settermjd can you backport this to 10.3?

PR #1717, opened 2019-08-27
Make the title more intuitive

settermjd:
Make the title more intuitive, because not everyone may understand or have heard of the FAQ acronym, so we're giving the full term, just to cover all bases.


settermjd:
Was only for testing purposes.

PR #1730, opened 2019-09-02
Guests app documentation

voroyam:
Added documentation here for the guests app as we had not have anything before that except one paragraph in user roles. But since our users / customer use the guests app intensively we should have a better documentation about it. This should be a start, we can add things to it later on as we discover possible use cases and issues.


voroyam:
looks great. will add screen shots later on

settermjd:
Thanks! I've revised the text, its flow, and embedded the YouTube video correctly, using the `video` macro. I've also created a docs-ui PR to better style the `videoblock` element, so that it renders better, as well.

PR #1731, opened 2019-09-02
[WIP] Update big file upload documentation

### LINKED ISSUE #1707: Big File upload docs need to be updated
First the document I'm talking about:
[Big File Upload Configuration](https://doc.owncloud.com/server/admin_manual/configuration/files/big_file_upload_configuration.html)

Following [a discussion in central](https://central.owncloud.org/t/htaccess-doesnt-work-with-php7-and-oc-10-2/21510/11), I noticed that the information inside this documentation article is "a little" dated. Perhaps it would be a good idea to go through it piece by piece and overhaul it.

For example:

> The default maximum file size for uploads, in ownCloud, is 512MB.

This might be correct for unchunked uploads, but these only occur in very specific edge cases nowadays. So the default is now that there is no maximum file size as even uploads through the WebUI are chunked.

> 32Bit OS-architecture

I think we can assume now that most OSs are 64bit

> Make sure that the latest version of PHP (at least 5.6) is installed

This is just wrong, as the latest version is PHP 7.4 and we only start supporting PHP 7.3 with the upcoming release of oC 10.3. On top of that, the mentioned PHP 5.6 is no longer supported since a few releases.

Aside from fixing the existing content I would like to submit an addition:
For very long running uploads (>1hr) into public folders (no chunking), 'filelocking.ttl' should be set to something very large (maximum file size / slowest assumed connection = time in seconds). 
I've written down a little bit more about that [in central](https://central.owncloud.org/t/htaccess-doesnt-work-with-php7-and-oc-10-2/21510/10?u=eneubauer), otherwise please ask me if you need more information.

- [x] New Content Addition
- [ ] Old Content Deprecation

- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

settermjd:
This updates the big file upload documentation to reflect changes in recent versions of ownCloud. It also fixes #1707.

### DIFF
--- a/modules/admin_manual/pages/configuration/files/big_file_upload_configuration.adoc
+++ b/modules/admin_manual/pages/configuration/files/big_file_upload_configuration.adoc
@@ -1,51 +1,59 @@
 = Big File Upload Configuration (> 512MB)
 :toc: right
+:mod_reqtimeout-url: https://httpd.apache.org/docs/current/mod/mod_reqtimeout.html
+:limitrequestbody-url: https://httpd.apache.org/docs/current/en/mod/core.html#limitrequestbody
+:sslrenegbuffersize-url: https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslrenegbuffersize
+:fcgidmaxrequestinmem-url: https://httpd.apache.org/mod_fcgid/mod/mod_fcgid.html#fcgidmaxrequestinmem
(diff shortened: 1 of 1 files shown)


settermjd:
@JKawohl, is this okay to merge?

PR #1745, opened 2019-09-03
Update playbook configuration for the 10.3 release

settermjd:
This change removes the master branch, replacing it with the 10.3 branch. Assuming it's correct, then only branches _10.1_, _10.2_, & _10.3_ will be built (and deployed). This requires #1744 to work properly.

💁‍♂ **Waiting on the 10.3 release to be ready. Will be deployed at that time.**


tboerger:
I'm still against dropping master build.

settermjd:
> I'm still against dropping master build.

no problem. reverted that aspect.

Issue #1751, opened 2019-09-04
Upgrading steps examples


ho4ho:
## What Needs to be Documented?
Improvement examples https://github.com/owncloud/docs/pull/1728 fixing https://github.com/owncloud/docs/issues/1710 and by added https://github.com/owncloud/docs/pull/1657 / https://github.com/owncloud/docs/issues/1551

Some examples posted by https://github.com/owncloud/docs/pull/1728#discussion_r320337662 how upgrading should be done:

> https://owncloud.org/news/owncloud-10-0-3-release/ said:
>
>> It is now possible to directly upgrade from 8.2.11 to 10.0.3 in a single upgrade process
>
> and https://github.com/owncloud/core/blob/v10.2.1/version.php#L33:
>
>> $OC_VersionCanBeUpgradedFrom = [[8, 2, 11],[9, 0, 9],[9, 1]];
>
>So example if understanding correctly:
>
> 8.2.11 -> Directly 10.2.1
> 8.2.10 -> Need 8.2.11, then 10.2.1
> 9.0.9 -> Directly 10.2.1
> 9.0.8 -> Need 9.0.9, then 10.2.1
> 9.1.8 -> Directly 10.2.1
> 9.1.0 -> Directly 10.2.1
> 
> and older:
> 
> 7.0.15 -> Need later 8.0.x, later 8.1.x, 8.2.11 then directly 10.2.1
> 7.0.10 -> Need 7.0.15, later 8.0.x, later 8.1.x, 8.2.11 then directly 10.2.1

## Where Does This Need To Be Documented?
https://github.com/owncloud/docs/blob/master/modules/admin_manual/pages/_partials/maintenance/major_release_note.adoc

## Why Should This Change Be Made? (Optional)
Better help for upgrading older versions.

## What Type Of Content Change Is This? (Optional)
- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## Which Manual Does This Relate To? (Optional)
- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

ho4ho:
See https://github.com/owncloud/core/issues/36160 still people people doing upgrade wrong. Maybe because by lacking clear steps?

PR #1753, opened 2019-09-05
add examples

voroyam:
fixes https://github.com/owncloud/docs/issues/1751


settermjd:
Here's an example of how the revised formatting renders.

table-example

voroyam:
Absolutely. Thank you for the help.

voroyam:
Does this has to be backported?

settermjd:
> Absolutely. Thank you for the help.

yw.

settermjd:
> Does this has to be backported?

I believe so, yes. I'll take care of that.

PR #1758, opened 2019-09-09
Correct the LDAP Password Change with Encryption

voroyam:
After feedback from users, we decided to adjust the wording on this topic and provide a clear instruction how to change the password of a LDAP user with user key encrypted ownCloud


settermjd:
Thanks for creating this, @voroyam.

PR #1770, opened 2019-09-10

voroyam:
- clarify that overwrite.cli.url is a config.php parameter
- describe why we are resetting the federation job


voroyam:
The Config PHP key has to be written explicitly in my opinion so that people can copy paste it.

The other changes are fine. Thanks

settermjd:
People shouldn't copy and paste code. Secondly, they'd have to change the example anyway.

voroyam:
> People shouldn't copy and paste code. Secondly, they'd have to change the example anyway.

In an ideal world I would agree with you. In an ideal world everyone would study the software they are using and reading the documentation and know what settings are set how. But unfortunately we don't live in that world.

Our goal, as far as I can see it, is to 
- get people as much information as possible before they skip over it.
- reduce support workload by providing easy to use documentation and how to manuals.

Of course your goal might be different than what I just wrote, that's why we have PRs and discussions.

PR #1792, opened 2019-09-12
remove nginx reference from docs

voroyam:
Since we decided that the official docs is ubuntu apache2 only, this removes any references of nginx from the documentation.

To Do: https://github.com/owncloud/docs/blob/master/modules/admin_manual/pages/configuration/files/big_file_upload_configuration.adoc

waiting for the rework PR of this document to remove the reference.


voroyam:
@settermjd can you backport this to 10.1 and 10.2?

PR #1793, opened 2019-09-12
Document oC 10.3 upgrade requirements for Amazon S3

settermjd:
This PR informs documents the fact that, when upgrading from versions of oC prior to 10.3, if they are using Amazon S3 as an external storage, admins need to install and enable [files_external_s3].

### Relates to 

- https://github.com/owncloud/files_external_s3/issues/235
- https://github.com/owncloud/files_external_s3/pull/237

[files_external_s3]: https://github.com/owncloud/files_external_s3


settermjd:
Thanks, @sharidas.

PR #1821, opened 2019-09-18
Remove manual section ids

settermjd:
While they may have started off as a good idea, over time they're creating maintenance nightmare. What's more, the vast majority aren't necessary, as the exact same values will be automatically generated. For these reasons, they're being removed.


voroyam:
I don't know what purpose they server. If you think we can remove them - so be it.

settermjd:
> I don't know what purpose they server. If you think we can remove them - so be it.

Manual section ids should only be created if, for example, there are two sections on the one page that have the same header, and you need to link to each of those sections from somewhere else in the documentation. 

If that's not the case, then it makes no sense, IMHO, to create them manually, as Antora generates section id links automatically.

mmattel:
shouldn´t this be backported?

settermjd:
You're right, it should. I'm not sure how this was overlooked.

PR #1854, opened 2019-09-25
Add package-lock.json

settermjd:


tboerger:
Why do you want to add that? This is a pure npm install thing, building the docs is documented with yarn which already provides a lock file.

Issue #1871, opened 2019-09-30
Document new public link endpoints


pmaier1:
## What Needs to be Documented?

ownCloud Server 10.3 comes with new endpoints to access public links via WebDAV and to send public link mail notifications.

- https://github.com/owncloud/core/pull/35932
- https://github.com/owncloud/core/pull/36063

To make use of both an admin needs to set `'dav.enable.tech_preview' => true,` in config.php.

## Where Does This Need To Be Documented?

Developer documentation

## Why Should This Change Be Made? (Optional)

New backend functionality should be documented for developers.

## What Type Of Content Change Is This? (Optional)
- [x] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## Which Manual Does This Relate To? (Optional)
- [ ] Admin Manual
- [x] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

settermjd:
Closed as has been implemented.

PR #1879, opened 2019-10-02
Document the public-files WebDAV API endpoint

settermjd:
This relates to owncloud/core/pull/35932 and #1871.


settermjd:
@DeepDiver1975, can you review, pls?

settermjd:
ping, @DeepDiver1975.

settermjd:
>     * request with password is missing - basic auth header with public:${share-password}

Thanks for calling that out. Will fix it.

settermjd:
>     * request with password is missing - basic auth header with public:${share-password}

Which example are you referring to? Both the curl and PHP request examples submit auth headers.

modules/developer_manual/examples/core/scripts/curl/dav/public_files/view_public_link.sh
```diff
@@ -0,0 +1,29 @@
...
+</d:propfind>
+EOF
+)
+
+curl "'$SERVER_URI/$API_PATH" \
+    -H 'Content-Type: application/xml; charset=UTF-8' \
+    -H 'Depth: 1' \
+    -X PROPFIND \
+    --data-binary "${REQUEST_BODY}" \
+    --user "${USERNAME}:${PASSWORD}" | xml_pp
```
DeepDiver1975: this is just wrong - we are talking about public link access - there is no user and password.

in case the public link is protected with a password: use 'public' as user name and the share link password as password

modules/developer_manual/examples/core/scripts/php/dav/public_files/view_public_link.php
```diff
@@ -0,0 +1,27 @@
...
+$client = new Client([
+    'base_uri' => '{oc-examples-server-url}/remote.php/dav/',
+]);
+
+$share_token = '<share_token>';
+
+try {
+    $response = $client->request('PROPFIND', "public-files/${share_token}", [
+        'auth' => [
+            '{oc-examples-username}',
```
DeepDiver1975: same as above

settermjd: Thanks for the clarification on both points.

settermjd:
@DeepDiver1975, both points of feedback have been addressed.

modules/developer_manual/examples/core/scripts/curl/dav/public_files/view_public_link.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+##
+## Variable Declaration
+##
+USERNAME={oc-examples-username}
```
DeepDiver1975: not needed

modules/developer_manual/examples/core/scripts/curl/dav/public_files/view_public_link.sh
```diff
@@ -0,0 +1,28 @@
...
+PASSWORD={oc-examples-password}
```
DeepDiver1975: not needed

settermjd:
Latest changes implemented.

PR #1891, opened 2019-10-08
Fix (potential) asciidoctor shell script bug

settermjd:
While I've never seen it before on macOS and haven't seen the bug
reported in Drone, when running bin/cli on Linux (Mint 19.2) today, I
received the following error:

```
asciidoctor: FAILED: input path /dev/fd/63 is a fifo, not a file
```

According to the asciidoctor man page, the asciidoctor synopsis is: 
"_asciidoctor [OPTION]... FILE..._".
Secondly: "_If FILE is - then the AsciiDoc source is read from standard input._"

From that, I can only determine that it should, on all platforms, have
problems with anonymous pipes. Strange why it didn't turn up sooner.

Making the change in this commit correct the issue (on Linux) and also
turned up a few other bugs in the documentation (that are being
corrected in other PRs).

I'd honestly love to know why this didn't show up sooner.


tboerger:
Not related to this PR, but there is an error within the PDF build: https://drone.owncloud.com/owncloud/docs/5819/1/6

settermjd:
@tboerger, does this require backporting / just to 10.3?

PR #1938, opened 2019-10-17
Integrate go.php redirects directly into the docs

### LINKED ISSUE #1937: Replace go.php links with page aliases
Currently we are maintaining a static list of redirect rules within the documentation deployment with a high effort. For every release we got to extend the list of redirects. To avoid that and to keep the maintainence more simple we should simply use page aliases.

You can see an example at [admin_manual/pages/appliance/configuration/clamav.adoc](https://raw.githubusercontent.com/owncloud/docs/master/modules/admin_manual/pages/appliance/configuration/clamav.adoc).

For the naming I would suggest aliases in the form of `go/alias-name`, e.g. `https://doc.owncloud.com/server/10.4/go/admin-install.html` which will redirect to `https://doc.owncloud.com/server/10.4/admin_manual/installation/`.

settermjd:
This fixes #1937, by integrating the NGINX redirects directly into the
docs, as best as possible, by using Antora redirects. Redirection to
anchors is not, currently, supported. However, as best as possible, the
redirects have been integrated.

### DIFF
--- a/modules/admin_manual/pages/configuration/database/db_conversion.adoc
+++ b/modules/admin_manual/pages/configuration/database/db_conversion.adoc
@@ -1,5 +1,6 @@
 = Converting Database Type
 :toc: right
+:page-aliases: go/admin-db-conversion.adoc
 
 == Introduction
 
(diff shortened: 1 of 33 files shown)


settermjd:
Thanks for the rapid approval, @tboerger. Are you going to update the NGINX redirects for this new approach, or shall I?

tboerger:
I don't know a good way to integrate the redirectly file automatically into the deployment, so please keep the builtin redirects for now.

PR #1947, opened 2019-10-19
release notes - occ app:disable can disable one app a time

adamgolebiowski:


mmattel:
The process described is correct, needed to do so on my own instance upgrade.
Thanks for updating.

PR #1983, opened 2019-10-23
Document the --progress/-p options for system:cron

settermjd:
This fixes #1955 by documenting the `--progress` and `-p` options available to `system:cron` in v10.3.1. It also adds in help output, so that the command is fully understood.

**Only** backport to 10.3 - **not to any earlier branch**.

PR #1989, opened 2019-10-24
Run cronjob in quiet mode

xelan:
Avoids output of progress in the cronjob, which is subsequently sent during every run if cron mailing is enabled.

The `--` separator is necessary, as otherwise the `-q` option would be passed to the PHP executable instead of the `occ` script.

### DIFF
--- a/modules/admin_manual/pages/_partials/configuration/server/occ_command/system_command.adoc
+++ b/modules/admin_manual/pages/_partials/configuration/server/occ_command/system_command.adoc
@@ -23,7 +23,7 @@ If you’re automating background jobs via Cron (or plan to), update the relevan
 /usr/bin/php -f /path/to/your/owncloud/cron.php
 
 # Use the following one instead
-/usr/bin/php -f /path/to/your/owncloud/occ system:cron
+/usr/bin/php -f /path/to/your/owncloud/occ -- -q system:cron
 ----
 
 [NOTE]


xelan:
#1983 seems related :smile:

settermjd:
@xelan, thanks for the contribution. I'll check it out and give a :+1: afterward, all being well.

xelan:
If occ is run directly without `php -f` (i.e. just `php occ` as in your example) the separator is not needed, and would be passed to the occ script literally.

Please try the full command from the PR.
```bash
php -f /path/to/your/owncloud/occ -- -q system:cron
```

xelan:
Apparently both ways work, either `php -f /path/to/occ` with the separator or `php /path/to/occ` without it.

settermjd:
I hear what you're saying, but that approach isn't the recommended way used throughout the docs to call `occ`. Why do you call it using the `-f` switch?

settermjd:
Thanks for clarifying that it was already documented that way. A large cleanup of the examples in the documentation was conducted about a month ago, and this must have been missed. I'll use that example to check for any others.

xelan:
Ah ok, thank you very much!

settermjd:
Closing, as the changes in #2004 negate the changes in this PR. Thanks for contributing the PR, though, @xelan.

xelan:
Thank you very much for addressing this issue so promptly and your fixes, @settermjd!

PR #1995, opened 2019-10-25
Add missing cron job

enbrnz:
For a complete setup you also need to set up `occ dav:cleanup-chunks` cron job as per documentation:
https://doc.owncloud.com/server/10.1/admin_manual/configuration/server/background_jobs_configuration.html#cleanupchunks
- I have set the cronjob to once a day after midnight
- By default the command deletes 2 day old chunks
- It is possible to configure another number of day(s) old chunks to delete as an argument to the command


voroyam:
@enbrnz this has to be Backported to every other branch, correct?

enbrnz:
> @enbrnz this has to be Backported to every other branch, correct?

Yes, this is correct

phil-davis:
@settermjd I don;t see a backport to 10.3 branch?

PR #1234, opened 2019-05-21
Document the system:cron command

settermjd:
This change documents the `system:cron` command, and fixes #281.

**only backport to 10.3!**

PR #1998, opened 2019-10-28
Remove references to php -f for occ command

settermjd:
This PR removes all relevant (lingering) references to calling occ using `php -f`, which were pointed out #1989 by @xelan.

### DIFF
--- a/modules/admin_manual/pages/release_notes.adoc
+++ b/modules/admin_manual/pages/release_notes.adoc
@@ -61,7 +61,7 @@ For code cleanup reasons, the execution of background jobs (e.g., for public lin
 
 * If you're using https://doc.owncloud.com/server/admin_manual/configuration/server/background_jobs_configuration.html#cron[System cron] to trigger background job execution, there is a new `occ` command (`occ system:cron`) which executes the background jobs. 
   To make use of it, you have to change the entry in `crontab`. 
-  Instead of executing `cron.php` (e.g., `/usr/bin/php -f /path/to/your/owncloud/cron.php`), cron should use `occ system:cron` (e.g., `/usr/bin/php -f /path/to/your/owncloud/occ system:cron`). 
+  Instead of executing `cron.php` (e.g., `/usr/bin/php -f /path/to/your/owncloud/cron.php`), cron should use `occ system:cron` (e.g., `{occ-command-example-prefix} system:cron`). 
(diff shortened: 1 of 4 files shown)


modules/admin_manual/pages/release_notes.adoc
```diff
@@ -61,7 +61,7 @@ For code cleanup reasons, the execution of background jobs (e.g., for public lin
 
 * If you're using https://doc.owncloud.com/server/admin_manual/configuration/server/background_jobs_configuration.html#cron[System cron] to trigger background job execution, there is a new `occ` command (`occ system:cron`) which executes the background jobs. 
   To make use of it, you have to change the entry in `crontab`. 
-  Instead of executing `cron.php` (e.g., `/usr/bin/php -f /path/to/your/owncloud/cron.php`), cron should use `occ system:cron` (e.g., `/usr/bin/php -f /path/to/your/owncloud/occ system:cron`). 
+  Instead of executing `cron.php` (e.g., `/usr/bin/php -f /path/to/your/owncloud/cron.php`), cron should use `occ system:cron` (e.g., `{occ-command-example-prefix} system:cron`).
```
mmattel: Shouldn´t the {occ...} also be set in the example?

mmattel:
You may want to check the crontab entry in #1234 to switch to {occ...}

modules/admin_manual/pages/release_notes.adoc
settermjd: Sory, but I don't follow what you're asking.

settermjd:
> You may want to check the crontab entry in #1234 to switch to {occ...}

I've updated the example in system_command.adoc on line 13 to use the site-wide variable. Thanks for pointing out that earlier oversight.

modules/admin_manual/pages/release_notes.adoc
mmattel: Sorry, you are right. Have overseen that cron.php was not called by occ... 🤦‍♂

voroyam:
This PR also removes the full path to php and occ. 

Either enter the full path or remove the note below the example code block that says "We added the full path on purpose".

I am for entering the full path, that way we can do less things wrong.

I am also for people reading the complete document and then making a change. But with the variables, as I mentioned multiple times, it's very hard to read.

@settermjd

Issue #292, opened 2018-05-22
[dev] Extra logging fields


PVince81:
Since 10.0.8 it is possible for developers to add additional JSON fields when creating log messages.

Example:
```
		$this->logger->info('Entity {entityName} has changed state to {newState}', [
			'app' => 'myapp',
			// these are used for text substitution in the above message
			'entityName' => 'X',
			'newState' => 'state2',
			// these will appear as additional JSON fields in the log entries
			'extraFields' => [
				'entityName' => 'X',
				'oldState' => 'state1',
				'newState' => 'state2',
				'entityOwner' => 'someuser',
			],
		]);
```

settermjd:
Backport to 10.1-3

PR #2003, opened 2019-10-28
[292] Document extra logging fields and string interpolation in logging

settermjd:
As outlined in #292, this change adds two logging examples that document
the ability to log plain text strings, as well as strings that are
interpolated, and the ability to create additional JSON fields when
logging.


settermjd:
@PVince81, please provide your feedback, time permitting.

PR #2017, opened 2019-10-31
Remove note about mtime

phil-davis:
`NOTE: ownCloud does not preserve the mtime (modification time) of directories...`

This note is appearing "randomly" in the middle of sections about the various file sharing options.

Maybe it was somewhere else in old docs? Where should it go?

### DIFF
--- a/modules/admin_manual/pages/configuration/files/file_sharing_configuration.adoc
+++ b/modules/admin_manual/pages/configuration/files/file_sharing_configuration.adoc
@@ -154,9 +154,6 @@ Check this option to enable auto-completion of ownCloud usernames.
 Check this option to restrict auto-completion of ownCloud usernames to only those users who are members of
 the same group(s) that the user is in.
 
-NOTE: ownCloud does not preserve the mtime (modification time) of directories, though it does update
-the mtimes on files.
-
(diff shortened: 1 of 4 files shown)


phil-davis:
@settermjd I'm sure its not completely random :)
Possibly it used to be a note under some other section/paragraph. And so it needs to be moved "back" to wherever it makes more sense.

settermjd:
@phil-d

> @settermjd I'm sure its not completely random :)
> Possibly it used to be a note under some other section/paragraph. And so it needs to be moved "back" to wherever it makes more sense.

Just wanted to be thorough in following this up.

settermjd:
Consequently, I'm not sure where to put it at the moment, but I'll grok the docs for a good spot for it.

settermjd:
One good place to put it might be after the first paragraph in [the retention section of the enterprise files tagging docs](https://github.com/owncloud/docs/blob/master/modules/admin_manual/pages/enterprise/file_management/files_tagging.adoc#retention). Another good place might be in the developer docs for [creating custom storage backends.](https://github.com/owncloud/docs/blob/master/modules/developer_manual/pages/app/advanced/storage-backend.adoc#required-methods), and for creating custom cache backends.

PR #2021, opened 2019-11-04
Correct developer documentation mistakes

### LINKED ISSUE #168: AppInfo Path Confusion

- [ ] Developer Manual

There is some confusion between the Routing doc and the DI Container doc.  For Routing, the given path is `appinfo/application.php` but then in the DI Container the path for apparently the same file is `lib/AppInfo/Application.php` and it is unclear what would be the difference, if any?

https://doc.owncloud.org/server/10.0/developer_manual/app/fundamentals/routes.html
https://doc.owncloud.org/server/10.0/developer_manual/app/fundamentals/container.html

settermjd:
A few mistakes in the developer documentation were pointed out in #168,
which have been causing confusion for some users (understandably). This
change corrects the mistakes, hopefully catching all instances of the
problem. It fixes #168.


settermjd:
@phil-davis, hope you don't mind the request for review.

PR #2036, opened 2019-11-06
Extract occ config command into partial

settermjd:
While not strictly necessary, my intent is that, over time, the occ command documentation will become easier to maintain, as each command can be edited more quickly as it is included in its own file.


mmattel:
👍 I like the idea !

Only a small thing, proposal, I suggest to rename the target document from `config_command.adoc` --> `config_commands.adoc` (plural) as the headline used is also plural.

settermjd:
> 👍 I like the idea !
> 
> Only a small thing, proposal, I suggest to rename the target document from
> `config_command.adoc` --> `config_commands.adoc` (plural) as the headline used is also plural.

I like what you're saying. However, on thinking about it further, I'm strongly leaning toward changing it to just `config.adoc`, so that it's `occ_commands/config.adoc`.

settermjd:
@mmattel, if you're happy with the most recent change that I've made, can you approve the PR so that it can be merged, pls?

PR #2057, opened 2019-11-11
Remove reference to full path for Cron to PHP in admonition

### LINKED ISSUE #2040: Fix Cron Background Job document
The PR removed the full path of the php script and the occ script. The note about the full path was left behind.

https://github.com/owncloud/docs/pull/1998#issuecomment-550267202

settermjd:
This fixes #2040. @voroyam, I'm not completely sure if the changed text in this PR is the text that you referred to in #2040, based on the PR's description.


voroyam:
Closing this PR.

The goal of my issue was to point out that the text was not noticed when changing the command and to revert the changes to the command i.e. adding full paths to occ and php. Also in 10.1 and 10.2 there is no need for the occ command.

PR #2079, opened 2019-11-14
Fix broken link to admin_manual/upgrading/marketplace_apps.html

settermjd:
Graylog reported that `admin_manual/upgrading/marketplace_apps.html` was 404'ing. This is happening because the file was renamed in 5a0bc56, but a page alias was not added to handle the updated page location. This change adds the required page-alias.


mmattel:
Backporting?

settermjd:
> Backporting?

Will be backporting to 10.3 at least. I haven't seen any log entries for older versions.

PR #2082, opened 2019-11-14
fixed the 10.1 manual upgrade document

voroyam:
after changing it in our knowledge sharing session this document got 2 upgrade sections.

This PR fixes it and brings it to the same state as the latest document in the master branch


settermjd:
@voroyam, thanks for spotting and fixing this.

PR #2092, opened 2019-11-18
Document ability to disable WND Notify functionality

settermjd:
This relates to https://github.com/owncloud/windows_network_drive/pull/248/files and https://github.com/owncloud/windows_network_drive/releases/tag/v1.0.1RC1.


settermjd:
@DeepDiver1975, it'd be great if you could give me feedback on this PR, pls.

settermjd:
@jvillafanez, can you provide feedback on this, pls?

jvillafanez:
This is fine. We could add that the password will be reset for the next requests regardless of the flag in order to remark that the lockout prevention will still be in place (although there could be cases not fully covered)

jvillafanez:
> The WND in-memory notifier for password changes provides the ability to notify all WND storages to reset their passwords.

I guess it's implicit that it will notify to all **affected** storages. I mean, if you have a connection to a different host, that connection won't have the password reset. The same if you use a different account to access to the storage.

settermjd:
Thanks for the feedback, @jvillafanez.

settermjd:
@jvillafanez, I've updated the text based on your feedback. I'm not sure that I've done it correctly, so please review and let me know.

PR #2135, opened 2019-12-04
Add go.php redirects for core#36315

VicDeo:
Add new aliases for https://github.com/owncloud/core/pull/36315
/go.php?to=use-https
/go.php?to=enable-http-strict-transport-security

should redirect to
/admin_manual/configuration/server/harden_server.html


tboerger:
Please also create a pr to the devops/websites repo on gitea.

PR #2161, opened 2019-12-18
Rename the document to distinguish it

voroyam:
When searching for "index.php" you will get documents which are named enable index.php-less URLs

There is no way to distinguish at that moment whether it's the appliance docs or the server docs.

Changing Enable to Configure should help in this situation. Until we can improve our search results... which may never come.

This needs to be backported to all the relevant branches to avoid confusion.


settermjd:
Sorry, but how does one word distinguish the information better than it was before?

voroyam:
Try typing in the search bar at the docs index.php.

What do you see right now? A bunch of pages named "Enable index...." in the various server versions / doc versions that are completely useless. You can't see if it's in the appliance manual or the server manual. 

Renaming it makes it different. Server page= Enable / Appliance page = Configure. 

So if you type in the search bar "Index.php" you get enable and configure.

Same thing we have with the WND docs. We have it in the appliance doc and in the server doc. But the documents are named differently, that's a quick way to distinguish them.

settermjd:
I see what you mean, however "completely useless" is a little bit of an overstatement (to me at least). I agree completely, though, that the search results need to be improved.

PR #2177, opened 2020-01-02
Update the background jobs configuration documentation

settermjd:
The docs had a duplicate copy of the "_Cron Jobs_" section. This was identified in #2040. This change removes the duplicate and also fixes up the formatting of the documentation so that it follows the style guide, making it easier to maintain.

PR #2180, opened 2020-01-03
[10.3] [PR 2177] Update the background jobs configuration documentation

mmattel:
Backport of PR #2177


mmattel:
np to me, but then you have to forward port these small changes to master, else they will get lost when phasing out 10.3...

PR #2182, opened 2020-01-03
Add missing file which caused PDF manual generation to throw errors

settermjd:
Add missing file to the developer manual which caused PDF manual generation to throw errors


mmattel:
Q: does this need forward porting to master or is this already in?

settermjd:
> Q: does this need forward porting to master or is this already in?

The changes are already available in master (_written ever so slightly differently_, yet having the same end result).

PR #2193, opened 2020-01-10
Update docs for 10.4 release

settermjd:
This PR will update the 10.4 branch for the 10.4 release, changing the version numbers and other, relevant, information.


antora.yml
```diff
@@ -1,6 +1,6 @@
 name: server
 title: ownCloud Server
-version: master
+version: 10.4
```
tboerger: No.... This is the master branch, so it's the master version.

modules/ROOT/pages/index.adoc
```diff
@@ -78,6 +78,33 @@ include::{partialsdir}maintenance-release-schedule-link.adoc[]
 These are the older ownCloud releases.
 Users of these releases are *strongly encouraged* to upgrade to the latest production release.
 
+=== ownCloud 10.3
```
tboerger: Why do we link all the versions here at all? There is the version switcher on the left side...

site.yml
```diff
@@ -9,9 +9,9 @@ content:
     - HEAD
   - url: https://github.com/owncloud/docs.git
     branches:
+    - '10.4'
     - '10.3'
     - '10.2'
-    - '10.1'
```
tboerger: Same here, keep it as part of the list.

mmattel: As far I remember (and I may be wrong) that only that latest three releases should be kept?

tboerger: These versions will receive backports, this doesn't mean that it will be removed from the version listing

mmattel: ok, thanks for the clarification.
Q: is there any statement how many releases are kept and when rolling starts? eg keep all 10.x releases until 11. And in case, which of the 10 release would get phased out.

modules/ROOT/pages/index.adoc
settermjd: It was standard practice for the longest time, to link to outdated versions.

site.yml
settermjd: @tboerger, customer support made the decision to only backport changes to the previous version. The previous _three_ versions, however, do get updates if they're to correct severe bugs. Otherwise, it's just the last version that's updated.

tboerger: Why should they be dropped from the version list at all?

settermjd: Because the decision was also made to only link the current version and the two previous versions in the documentation. I'm just going along with it.

settermjd: 10.1 has been re-added to the list of available content sources.

modules/ROOT/pages/index.adoc
tboerger: Back in that time there haven't be any version selector like we got within the antora navigation.

settermjd: @pako81 @cdamken @voroyam should the links to older versions be removed?

voroyam: Honestly - I use the version switcher. I haven't seen this site since I was looking for an outdated version to test stuff, like oC 9.1.

I think we should only link the versions that are no longer in the version switcher.

settermjd: latest three versions have been removed from the previous releases list. Can this be approved now?

cdamken: IMHO we should have only the last 2 available (3 could be ok too), but master should not appear (not even in the search)

cdamken: that means "latest" should be the lastest with number and not master

settermjd:
> The changes looks good.
> It is unclear to me, why the CI would report an antora error related to cdamken.
> Help appreciated here. This issue blocks the 10.4 release.

Looking at it now.

Issue #2192, opened 2020-01-10
Publish server/10.4/user_manual/


jnweiger:
This is needed for https://github.com/owncloud/enterprise/issues/3722, please.
The build of the tar requires access to https://doc.owncloud.com/server/10.4/user_manual/ownCloud_User_Manual.pdf

settermjd:
The 10.4 branch is now available, and there are three PRs that are being backported to it. Just need approval on those, and the branch is ready to be published. I think I need to also update the master branch wrt the 10.4 branch being published.

jnweiger:
I just saw an error in https://drone.owncloud.com/owncloud/docs/6622
```
Error: Duplicate nav: master@server:ROOT:nav.adoc
```

@settermjd Is that blocking the update?

settermjd:
It would have been, but is resolved now.

settermjd:
The 10.4 branch is currently available. It is not being automatically redirected to from latest, but that is coming after a current PR is reviewed and merged.

settermjd:
You can find it at https://doc.owncloud.com/server/10.4/.

settermjd:
The links to the PDFs are not, currently, working. Am working to fix that.

jnweiger:
The 10.4 overview page page https://doc.owncloud.com/server/10.4/ says "Latest Stable Release (version 10.2)" and links into the 10.2 tree. 

![image](https://user-images.githubusercontent.com/1108546/72511424-7089c080-384b-11ea-849f-1816f90af12f.png)

10.3 and 10.4 should be added to the overview.

settermjd:
> The 10.4 overview page https://doc.owncloud.com/server/10.4/ with the version switcher set to "10.4" says "Latest Stable Release (version 10.2)" and links into the 10.2 tree.

I thought that I submitted a PR for that. If not, I'll get right on it.

> 
> 10.3 and 10.4 should be added to the "10.4" switch, -- they are present though when I set the switch to "master". (But as noted above, the link to the 10.4 PDF is a 404)

Not quite sure what's going on there, yet. Will aim to have a PR created for it today, if not to have it resolved.

jnweiger:
Success!
https://doc.owncloud.com/server/10.4/user_manual/ownCloud_User_Manual.pdf is now there!

@settermjd should we wait for https://github.com/owncloud/docs/issues/2236 before progressing with https://github.com/owncloud/enterprise/issues/3722 ?

settermjd:
[...]

PR #2196, opened 2020-01-10
Remove 10.1 branch from and add 10.4 to the publishing list

settermjd:
This relates to #2192 and #2193. Unless I'm mistaken, this stops the 10.1 branch from being published and adds 10.4 to the published branches list.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -9,9 +9,9 @@ content:
     - HEAD
   - url: https://github.com/owncloud/docs.git
     branches:
+    - '10.4'
     - '10.3'
     - '10.2'
-    - '10.1'
   - url: https://github.com/owncloud/android.git
     branches:
     - master


settermjd:
This branch won't pass until #2193 has been merged.

site.yml
```diff
@@ -9,9 +9,9 @@ content:
     - HEAD
   - url: https://github.com/owncloud/docs.git
     branches:
+    - '10.4'
     - '10.3'
     - '10.2'
-    - '10.1'
```
tboerger: This should be kept, otherwise it also won't appear on the version list anymore and we got more reasons why we can't wipe old files from the deployment bucket.

settermjd: Fair enough.

PR #2201, opened 2020-01-11
Fix symbolic link creation

m8335la:
with the given example, it would create the symbolic link in the data directory, making it data/data. This should fix it.d


mmattel:
@m8335la thanks for the PR.
Looking over it, imho some changes are needed anyways as things are just wrong:

`If you need to move your ownCloud data directory from its current location to somewhere else, here is a manual process that you can take to make it happen.`
-->
`Here is a description for this manual process, if you need to move your ownCloud data directory from its current location to somewhere else without using a symbolic link.` 

Saying that, item 3 and the command with `ln -s` can be deleted.
`rsync` anyways copies the data from the old to the new location, so what is linking good for.
`ln -s` can't be successful as the data directory already exists and cant be linked anymore. 

Furthermore the current `rsync` command is wrong. It creates and copies the data folder on the target making the post steps wrong.
`rsync -avz /var/www/owncloud/data /mnt/owncloud`
-->
`rsync -avz /var/www/owncloud/data/ /mnt/owncloud` 
The corrected command only copies the content of the data directory without the directory itself.

I personally highly recommend using as target `/mnt/owncloud/data` because when you have more than one instance running ect, a proper path naming is quite helpful to remember what things are about. (then the rysnc command is correct, but the paths must be changed in the database commands below)

@settermjd

modules/admin_manual/pages/maintenance/manually-moving-data-folders.adoc
```diff
@@ -28,7 +28,7 @@ To save time, here’s the commands which you can copy and use:
 ----
 apachectl -k stop
 rsync -avz /var/www/owncloud/data /mnt/owncloud
-ln -s /mnt/owncloud /var/www/owncloud/data
+ln -s /mnt/owncloud/data /var/www/owncloud/data
```
settermjd: @mmattel was correct in [his comment](https://github.com/owncloud/docs/pull/2201#issuecomment-573400716). This will symlink `/mnt/owncloud/data` inside `/var/www/owncloud/data`. It should be changed to the following:

```console
ln -sF /mnt/owncloud/data /var/www/owncloud
```

Adding the `-F` option ensures that the target directory is removed so that the new symlink can be created without error.

mmattel:
@settermjd are you still sure that creating a symbolic link is necessary as the directory is migrated and not linked. pls see my comments above

settermjd:
> @m8335la thanks for the PR.
> 
> I personally highly recommend using as target `/mnt/owncloud/data` because when you have more than one instance running ect, a proper path naming is quite helpful to remember what things are about. (then the rysnc command is correct, but the paths must be changed in the database commands below)
> 

@mmattel, a symlink is worthwhile to avoid updating the ownCloud configuration, and that example (the data directory) is the standard ownCloud directory. Might just be a handy choice for people to make. That said, I guess it makes sense to update the configuration for the data directory, instead of symlinking it in place, as the directory is being moved. That, would be the most congruent. That said, either updating the ownCloud configuration or using a symlink are equally valid approaches.

modules/admin_manual/pages/maintenance/manually-moving-data-folders.adoc
settermjd: Pls, disregard the above. After a lot of thought, it makes most sense to me to replace the symlink instruction with an instruction showing how to update the ownCloud config to reference to the new directory location. Can you make that change, pls?

mmattel:
@settermjd reading the document title carefully (Manually Move a Data Directory) and the subesction below (Fix Hardcoded Database Path Variables), there is no text that says you can do either symlinking which does not need database hardcore changes or hard moving which does not need symlinking. Atm, the document with its current structure is a mix and not clear. This is what my comments with explanation above tell.

settermjd:
Okay. @m8335la, please update the PR to remove the symlink command and to update ownCloud's config to reference the new directory instead.

m8335la:
@settermjd, consider that done. My original intention was to preserve the option of symlinking as a valid approach for single-instance environments - at least it helped me :-). But as it's now it's also fine for me.

mmattel:
@m8335la 
you either have to write as path `/mnt/owncloud/data` (on all locations) and keep the rsync parameters or you change the rsync parameter to make it work. For details pls see my comments above.

settermjd:
> @settermjd, consider that done. My original intention was to preserve the option of symlinking as a valid approach for single-instance environments - at least it helped me :-). But as it's now it's also fine for me.

Thanks kindly.

m8335la:
@mmattel thanks for the hints, looking forward to your new feedback

modules/admin_manual/pages/maintenance/manually-moving-data-folders.adoc
```diff
@@ -12,23 +12,22 @@ to make it happen.
...
 1.  Stop Apache
 2.  Use rsync to sync the files from the current folder to the new one
-3.  Create a symbolic link from the new directory to the old one
-4.  Double-check xref:installation/installation_wizard.adoc#post-installation-steps[the directory permissions] on the new directory
+3.  Double-check xref:installation/installation_wizard.adoc#post-installation-steps[the directory permissions] on the new directory
+4.  Change ownCloud configuration to point to the new directory (see below)
 5.  Restart Apache
 
-To save time, here’s the commands which you can copy and use:
+To save time, here’s the commands which you can copy and use for Apache and rsync:
```
settermjd: Please change "here's" to "here are".

mmattel:
@m8335la many thanks for all the work!
Re-viewing  the whole document, I found more things that need improvement / correction:

Proposing a text improvement at the top:
`If you need to move your ownCloud data directory from its current location to somewhere else,...` -->
`If you need to move your ownCloud data directory from its current location to somewhere else without using symbolic links,...`

Before the ` NOTE:`  add a new section as the following text is not part of the introduction like: `== Prepare and Copy Data`

Fix needed:
```
UPDATE oc_accounts
SET
    home = REPLACE(home,
        '/var/www/owncloud/',
        '/mnt/owncloud/data/');
```
I guess that this was wrong before, but `/var/www/owncloud/` should be `/var/www/owncloud/data/`

Improvement needed:
```
UPDATE oc_jobs SET argument = '/your/new/data/path'
  WHERE id = <id of the incorrect record>;
```
You may want to replace for consitency `/your/new/data/path` with `/mnt/owncloud/data/`

mmattel:
Closing in favor of #2339

PR #2217, opened 2020-01-15
Split the occ commands out into separate files

settermjd:
This PR splits all of the occ core and app command documentation out from the one, monolithic occ_command.adoc file into a set of smaller, command-specific files. The intent is to make the command easier and quicker to maintain.


voroyam:
So, this document has been split, then merged, and now you want to split it again.

I agree with you that it's easier to maintain a smaller document, but from the support view - it's harder to find information when it's split up. Especially with the search being in the subpar state and no hope for it getting better.

settermjd:
This is a different kind of split. Before, it was the core and app commands, which were justifiably re-joined. However, this is just splitting out the commands into separate files, so that each one is easier to maintain. The information will still appear in the rendered documentation as before, well, in an alphabetical order, which should make it that much easier to scan, in the absence of proper search functionality.

settermjd:
@voroyam, any chance of a :+1: on this?

voroyam:
hm, my review was removed and I can't re review this. So thumbs up it is :)

PR #2220, opened 2020-01-15
remove unnecessary LDAP documentation structure

voroyam:
removed the LDAP folder and moved the ldap proxy cache server to the general configuration site

@cdamken


voroyam:
true, file removed.

PR #2224, opened 2020-01-15
remove ldap:update-group command

voroyam:
because it is deprecated and was removed a while ago from the occ commands


settermjd:
Can you create this off of #2217 pls?

settermjd:
Ideally, as the PR creator, that would be a job for you.

voroyam:
> Ideally, as the PR creator, that would be a job for you.

ofc I will do this. I just wanted to know what is the approach favored by you :)

voroyam:
okay, will open new PR

PR #2231, opened 2020-01-20
Use tabs only when rendering content in HTML format

settermjd:
This change is being made as there are build errors when generating PDF versions of the documentation because of the misunderstood tab formatting.

PR #2233, opened 2020-01-20
Update build tools to be format-aware

settermjd:
PR #2231 was merged without the relevant changes to the Drone configuration and PDF manual build script. This change makes the required changes so that the manuals build correctly and include the content with/without tabs, as appropriate.


settermjd:
Thanks, @tboerger.

PR #2242, opened 2020-01-22
Update index.adoc

jnweiger:
Contacts and calender are also not included in ownCloud 10.

Looking at demo.owncloud.org, (10.3) I have a hard time finding a 'menu:Apps[Productivity]' element anywhere. 
This snippet is included in the PDF, but not in the HTML version. Why do they differ?


settermjd:
@jnweiger, I did a little digging and found that the menu reference and surrounding instructions were relating to installing productivity apps from the Marketplace. I've updated them in fc39192 to better reflect what is required.

PR #2287, opened 2020-02-05
fix simple typo

enbrnz:
too many s


voroyam:
needs back port to all relevant branches

Issue #2140, opened 2019-12-05
New config-to-docs run neccessary


mmattel:
## What Needs to be Documented?
A new config-to-docs run for [10.4] is necessary because of:
https://github.com/owncloud/core/pull/36413 ([For 10.4] Set 599 HTTP code on error)
https://github.com/owncloud/core/pull/36526 (Add sample OIDC configuration to the sample app config file)

## Why Should This Change Be Made? (Optional)
Sync core and docs

## What Type Of Content Change Is This? (Optional)
- [x] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## Which Manual Does This Relate To? (Optional)
- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

@settermjd @micbar @jvillafanez @phil-davis

settermjd:
Would you mind doing this, @mmattel?

mmattel:
Yes, when 36526 is merged. the other one is already merged.

mmattel:
After a successful run, additional changes are made in core which need another run when merged.

- [x] https://github.com/owncloud/core/pull/36413
- [ ] https://github.com/owncloud/core/pull/36526
- [ ] https://github.com/owncloud/core/pull/36796
- [ ] https://github.com/owncloud/core/pull/36815

PR #2290, opened 2020-02-05
Update of config.sample files with latest changes from core

mmattel:
This is an update of config.sample files with the latest changes from core.
Fixes https://github.com/owncloud/docs/issues/2140

Backport to 10.4 / 10.3 necessary.


mmattel:
Proposal: lets backport to 10.4 only.

PR #2245, opened 2020-01-22
Move release notes and what's new pages to the ROOT module

settermjd:
The admin and user manuals both had "_What's New_" pages, and the admin manual contained the project's release notes, while the root module contained an outdated set of release notes. This PR moves the release notes from the admin manual to the root module, removes the What's New pages from the admin and user manuals and stores the most up to date copy of the file in the ROOT module.

This fixes #2240.

PR #2292, opened 2020-02-06
[10.4] [PR 2245] Update Release Notes Information

settermjd:
Backports #2245.


pmaier1:
Why do we need to change the location for release notes? They're linked on certain locations. Will the final URL be the same then?

settermjd:
It came up in discussion with @jnweiger previously. If this change isn't worthwhile, I'll revert the PR in master.

pmaier1:
I don't see the point in moving it out of the admin docs. Also we would have to change the links everywhere. Would not do it unless there's a good reason I don't see right now.

PR #2317, opened 2020-02-17
Update system_requirements.adoc

HanaGemela:


mmattel:
@HanaGemela 
Question, is this a backport of an existing PR? In case pls reference.
If not, why pushing to 10.3 instead of master?

HanaGemela:
@mmattel no, this is the first PR into server docs. I want this to be in the documentation for latest stable release. It can be backported to master and 10.4.
Or maybe we don't need this information in server docs at all as it is maintained in client docs?

mmattel:
FYI, we always file aginst master and the backport to 10.4 or 10.3 ect. Anything else makes the process not smooth and confusing. Only exception are fixes of existing stuff, but this is quite rare. It would be great if we could stick to this scheme.

HanaGemela:
@mmattel OK. I'll close this and create a new PR into master. Who is doing the backporting? Should I do it using backport.sh?

mmattel:
Great, awaiting your new PR :smile:
Reagrding BP, yes, use the script. You have to wait until master is merged as you need the merge reference. (But you also can ask @settermjd for support) It should create automagically the correct header text for GH which makes it easy to identify based on this header the root PR.

PR #2320, opened 2020-02-18
Update system_requirements.adoc

HanaGemela:
To match https://doc.owncloud.com/desktop/2.6/installing.html


mmattel:
Pls backport to 10.4
For 10.3 to be discussed, as 10.3 runs for a while.
If valid then BP, else not.

PR #2321, opened 2020-02-19
Changed recommendations to meet real life scenarios. And streamlined …

JKawohl:
Streamlined information. ownCloud doesn't work well/at all  with master-master replication.

### DIFF
--- a/modules/admin_manual/pages/installation/deployment_recommendations.adoc
+++ b/modules/admin_manual/pages/installation/deployment_recommendations.adoc
@@ -266,7 +266,7 @@ for information on configuring ownCloud to use LDAP and AD.
 Session management on the application server. PHP sessions are stored in
 a temporary filesystem, mounted at the operating system-specific session
 storage location. You can find out where that is by running
-`grep -R 'session.save_path' /etc/php5` and then add it to the
+`grep -R 'session.save_path' /etc/php*` and then add it to the
 `/etc/fstab` file, for example:
 
 [source,console]
@@ -464,4 +464,4 @@ See http://galeracluster.com/documentation-webpages/[the Galera Cluster WSREP do
 
 * http://www.severalnines.com/blog/become-mysql-dba-blog-series-database-high-availability[Database High Availability]
 * http://blog.bitnami.com/2014/06/performance-enhacements-for-apache-and.html[Performance enhancements for Apache and PHP]
-* https://ubuntuhowtoo.blogspot.com/2019/10/linux-ubuntu-1804-use-redis-for-cache.html[How to Set Up a Redis Server as a Session Handler for PHP on Ubuntu 14.04]
+* https://ubuntuhowtoo.blogspot.com/2019/10/linux-ubuntu-1804-use-redis-for-cache.html[How to Set Up a Redis Server as a Session Handler for PHP on Ubuntu 18.04]
--- a/modules/admin_manual/pages/installation/deployment_recommendations.adoc
+++ b/modules/admin_manual/pages/installation/deployment_recommendations.adoc
@@ -234,9 +234,10 @@ management on the application servers.
 ==== Database
 
 MySQL/MariaDB Galera cluster with
-https://mariadb.com/kb/en/mariadb/replication-cluster-multi-master/[master-master replication].
+https://mariadb.com/kb/en/setting-up-replication/[master-slave replication].
 InnoDB storage engine, MyISAM is not supported, see:
 xref:configuration/database/linux_database_configuration.adoc#mysql-mariadb-storage-engine[MySQL / MariaDB storage engine].
+For mariadb consider: https://mariadb.com/kb/en/mariadb-maxscale-22-automatic-failover-with-mariadb-monitor[MariaDB Monitor] to configure your setup for a failover Scenario.
 
 ==== Backup
 
@@ -279,9 +280,9 @@ for information on selecting and configuring a memory cache.
 
 ==== Storage
 
-For accessing a backend storage system via NFS, you can use a dedicated storage system like 
+For accessing a backend storage system via NFS, you can use a dedicated storage system like
 {netapp-hybrid-flash-array-url}[NetApp Hybrid Flash Storage Systems], or other systems like
-{ibm-elastic-storage-server-url}[IBM Elastic Storage] based on their Power8 servers or 
+{ibm-elastic-storage-server-url}[IBM Elastic Storage] based on their Power8 servers or
 {redhat-ceph-url}[RedHat Ceph] with their NFS-Ceph gateway.
 
 You may take a look on the {netapp-nfs-bpg-url}[NetApp NFS Best Practice and Implementation Guide] for best NFS configuring practices,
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/installation/deployment_recommendations.adoc
```diff
@@ -234,9 +234,10 @@ management on the application servers.
 ==== Database
 
 MySQL/MariaDB Galera cluster with
-https://mariadb.com/kb/en/mariadb/replication-cluster-multi-master/[master-master replication].
+https://mariadb.com/kb/en/setting-up-replication/[master-slave replication].
```
mmattel: Proposing the following:

**First**
Regarding URL´s as we do this in all the other documents, pls change them **all** to following scheme:

Put the URL you want to show on top into a macro (see examples close below `:toclevels: 1` like
`:setting-up-replication-url: https://mariadb.com/kb/en/setting-up-replication/`

Then use this in the line you want to print it:
`{setting-up-replication-url}[master-slave replication]`
This improves readability and link managability a lot.

**Second**
There is a section `== General Recommendations` recommending `PHP 5.6+`.
Same 2x (!) in `==== Session Management` with a console commands.

This should be changed to the new minimum of PHP 7.3
see
https://doc.owncloud.com/server/10.4/admin_manual/installation/system_requirements.html
and
https://www.php.net/supported-versions.php

**Third**
In section `== References` change the following
from
`https://www.digitalocean.com/community/tutorials/how-to-set-up-a-redis-server-as-a-session-handler-for-php-on-ubuntu-14-04`
to
`https://ubuntuhowtoo.blogspot.com/2019/10/linux-ubuntu-1804-use-redis-for-cache.html`
because our minimum supported Ubuntu release is 18.04...

modules/admin_manual/pages/installation/deployment_recommendations.adoc
```diff
@@ -150,7 +152,7 @@ for information on configuring ownCloud to use LDAP and AD.
 Local session management on the application server. PHP sessions are
 stored in a temporary filesystem, mounted at the operating
 system-specific session storage location. You can find out where that is
-by running `grep -R 'session.save_path' /etc/php5` and then add it to
+by running `grep -R 'session.save_path' /etc/php*` and then add it to
 the `/etc/fstab` file, for example:
 
 [source,console]
```
mmattel: pls check the command below, it is referencing to php5
Same in line about 264 and 274 (maybe shifted a bit but you will find it)

modules/admin_manual/pages/installation/deployment_recommendations.adoc
```diff
@@ -460,4 +464,4 @@ See http://galeracluster.com/documentation-webpages/[the Galera Cluster WSREP do
 
 * http://www.severalnines.com/blog/become-mysql-dba-blog-series-database-high-availability[Database High Availability]
 * http://blog.bitnami.com/2014/06/performance-enhacements-for-apache-and.html[Performance enhancements for Apache and PHP]
-* https://www.digitalocean.com/community/tutorials/how-to-set-up-a-redis-server-as-a-session-handler-for-php-on-ubuntu-14-04[How to Set Up a Redis Server as a Session Handler for PHP on Ubuntu 14.04]
+* https://ubuntuhowtoo.blogspot.com/2019/10/linux-ubuntu-1804-use-redis-for-cache.html[How to Set Up a Redis Server as a Session Handler for PHP on Ubuntu 14.04]
```
mmattel: `[How to Set Up a Redis Server as a Session Handler for PHP on Ubuntu 14.04]` 
-->
`[How to Set Up a Redis Server as a Session Handler for PHP on Ubuntu 18.04]`

modules/admin_manual/pages/installation/deployment_recommendations.adoc
mmattel: @JKawohl pls check the command
`echo "tmpfs /var/lib/php5/pool-www tmpfs defaults,noatime,mode=1777 0 0" >> /etc/fstab`
It is on two locations and contains a reference to php5 which should be at least php7.2 or php7.3

BTW, the path `/var/lib/php5/pool-www` is not valid for Ubuntu, this should be `/etc/php7.2/` but I do not know the subdirectory where `/pool-www` is located because I am running nginx with fpm and not apache.

As a solution, maybe to write it more neutral, do a grep in `/etc/php7.` to find where `pool-www` is located and suggest to use that found path in the command above.

modules/admin_manual/pages/installation/deployment_recommendations.adoc
mmattel: Some URL´s need to be fixed according the suggestion abobe, but this can be done in an other PR. The mentioned once are corrected.

mmattel:
Sorry for that many comments, but if there is a fix/improvement, it should be correct...

modules/admin_manual/pages/installation/deployment_recommendations.adoc
enbrnz: > @JKawohl pls check the commands
> `echo "tmpfs /var/lib/php5/pool-www tmpfs defaults,noatime,mode=1777 0 0" >> /etc/fstab`
> It is on two locations and contains a reference to php5 which should be at least php7.2 or php7.3

PHP-FPM should not be mentioned in the docs at all, as it is unsupported.
 
> BTW, the path `/var/lib/php5/pool-www` is not valid for Ubuntu (which we usually use for examples), this should be `/etc/php7.3/` but I do not know the subdirectory where `/pool-www` is located because I am running nginx with fpm which has a different sub location and not apache.
> 
> As a solution, maybe to write it more neutral, do a search in `/etc/php7.*` to find where `pool-www` is located like `find /etc/php7.* -type d | grep "pool-www"` and suggest to use the found path in the command above.

As far as I know the correct path for PHP configuration in Ubuntu version 18.04 and higher is `/etc/php/7.x/`

This whole document needs to be reworked, apparently. Perhaps we quickly apply the changes @JKawohl  suggested, so the really big mistakes are gone, and then we ask the doc-team to go over this document.

mmattel: > PHP-FPM should not be mentioned in the docs at all, as it is unsupported.

Not inteded to be mentioned. I just wrote about my environment and because I do not have Apache, I can't prove the location of `pool-www`

Apparrently the command paths are wrong. Ok to fix this in a separate PR.
Can you pls open a issue and add a link to this comment block, so it is exactly known what it is about.

JKawohl:
> I agree that we should remove the remarks about the primary-primary replication

Thank you, for a rewrite, we should make a decision if we want to move to primary-secondary language .

http://edition.cnn.com/2003/TECH/ptech/11/26/master.term.reut/
https://www.theregister.co.uk/2018/09/11/python_purges_master_and_slave_in_political_pogrom/

modules/admin_manual/pages/installation/deployment_recommendations.adoc
settermjd: Thanks for the discussion here, people. I'll get the relevant changes implemented.

PR #2116, opened 2019-11-27
Update system_requirements.adoc

voroyam:
removed old stuff, restructured

PR #2328, opened 2020-02-20
[10.3] [PR 2320] Update system_requirements.adoc

mmattel:
Backport of PR #2320

### DIFF
--- a/modules/admin_manual/pages/installation/system_requirements.adoc
+++ b/modules/admin_manual/pages/installation/system_requirements.adoc
@@ -96,8 +96,13 @@ recommend a minimum of 512MB.
 
 === Mobile OS
 
-* iOS 9.0+
-* Android 4.4+
+* Windows 7+
+* macOS X 10.10+ (*64-bit only*)
+* CentOS 6 & 7 & 8
+* Debian 8.0 & 9.0 & 10.0
+* Fedora 27 & 28 & 29 & 30 & 31
+* Ubuntu 18.04 & 19.04 & 19.10
+* openSUSE Leap 42.2 & 42.3 & 15.0 & 15.1 & 15.2
 
 === Web Browser
 
--- a/modules/admin_manual/pages/installation/system_requirements.adoc
+++ b/modules/admin_manual/pages/installation/system_requirements.adoc
@@ -96,13 +96,14 @@ recommend a minimum of 512MB.
 
 === Mobile OS
 
-* Windows 7+
+* Linux
+** CentOS 6 & 7 & 8
+** Debian 8.0 & 9.0 & 10.0
+** Fedora 27 & 28 & 29 & 30 & 31
+** Ubuntu 18.04 & 19.04 & 19.10
+** openSUSE Leap 42.2 & 42.3 & 15.0 & 15.1 & 15.2
 * macOS X 10.10+ (*64-bit only*)
-* CentOS 6 & 7 & 8
-* Debian 8.0 & 9.0 & 10.0
-* Fedora 27 & 28 & 29 & 30 & 31
-* Ubuntu 18.04 & 19.04 & 19.10
-* openSUSE Leap 42.2 & 42.3 & 15.0 & 15.1 & 15.2
+* Microsoft Windows 7+
 
 === Web Browser
 


modules/admin_manual/pages/installation/system_requirements.adoc
```diff
@@ -96,8 +96,13 @@ recommend a minimum of 512MB.
...
 
-* iOS 9.0+
-* Android 4.4+
+* Windows 7+
+* macOS X 10.10+ (*64-bit only*)
+* CentOS 6 & 7 & 8
+* Debian 8.0 & 9.0 & 10.0
+* Fedora 27 & 28 & 29 & 30 & 31
+* Ubuntu 18.04 & 19.04 & 19.10
+* openSUSE Leap 42.2 & 42.3 & 15.0 & 15.1 & 15.2
```
phil-davis: This should be in a `Desktop` section, not `Mobile OS`

mmattel: Can you find out which one is missing?
I have no clue...
These kind of things should not happen 😢

phil-davis: PR #2116 was applied to 10.3 branch 2019-11-28. It removed the whole "Desktop" section and added a "Client Operating System" section.

A change like that does not seems to have been made to `master` branch. So we now have 10.3 branch with quite a different `system_requirements` doc.

@voroyam do you have memory about why PR #2116 made that change, but was not applied to master?

voroyam: Not really. Maybe it was in that period of time after the discussion in the docs internal channel, where we + Felix decided to only make changes to one branch and don‘t backport stuff. I am not sure.

Why those kind of things happen - because of the messy branch selector at our online docs.

You can see only the 3, but in the search you get the master. 

None of the people I know that work on the docs, including me have a local copy of the branch, changes are made in the online docs via „Edit this page“ and this is messy. This is how these things happen.

mmattel: This is a good point you mention and editing via the docs web page is one of the many methods.
Because a PR is created, none of the creators gets hindered to do a back/forwardport of the changes made...
Except of a very specific fix in a particular release/branch, all changes should be made in master and ported, to exactly avoid those issues.

voroyam: I have bad experience with the backporting script.  Sometimes it messes up the name, not all info is there, you have to enter the description, comment, name of the PR manually.

I have tried to fix the manual upgrade page 10 times I think, because it was messed up. In the end I did a manual PR for each branch.

You are also required to have a local copy of the docs repo on your system to run this script. Not everyone does.

mmattel: It is the mess that must be avoided. How you achieve this is your personal decision.

voroyam: Well, there is no clear guide who back ports things how. Sometimes it's enough to write that it has to be back ported and what branches, sometimes you have to do it yourself.

mmattel: As a matter of good practice, the writer takes care that it happens and tracks/(re)triggers until it's done.

settermjd: In all the time that I've been using the backporting script, I've barely had any issues with it. Sure, it only fills out the title and description properly if the PR contains one commit, which is an odd limitation. But, the information is always available in the PR's initial commit list, if there are more than one commits available. 

Other than that, it's just a matter of calling the backport script and passing the relevant commit hash and branch to backport to. After a few attempts, I don't see why it would be challenging. I could be wrong. Perhaps. 

Additionally, backporting is something that has been discussed so much over the last several months, including which branches are backported to and why, which we documented after that discussion. It seems strange to me that this is even still a discussion being had.

settermjd: @mmattel, if you sort this list, it'd make them loads quicker to read.

mmattel: @settermjd I think becasue of missing back/forwardports (see comment from @phil-davis above), we should first fix that issue for all the relevant branches and then do a new PR in master with your changes and backport that one. Else the mess becomes imo even bigger.

settermjd: I'll work with this today to get it backported to all the relevant branches.

PR #2332, opened 2020-02-24
Forward port system requirements differences in 10.3

settermjd:

PR #2340, opened 2020-02-26
Backport system_requirements to 10.4

phil-davis:
This gets `system_requirements.adoc` to be the same is in docs `master`
(after PR #2332 to `master` applied various stuff from `10.3` forward to `master` and ...)

IMO we need to merge this so that we get the branches back in sync.

Then we can discuss what happened for the `Sync Clients and Mobile Apps` heading and the deletion of the `Mobile OS` section etc. I will make a PR to `master` suggesting what I think!


phil-davis:
I cherry-picked the commits from #2342 into here.
That gets things back into sync.

Issue #2343, opened 2020-02-26
PostgreSQL 10 support in core 10.4


phil-davis:
https://github.com/owncloud/core/issues/33187

Documentation update needed for 10.4

phil-davis:
Done in master PR #2344 and backported to 10.4 docs in PR #2345

PR #2344, opened 2020-02-26
PostgreSQL 10 is supported from core 10.4

phil-davis:
Issue #2343 

Backport to 10.4 only.

Note: `linux_database_configuration.adoc` has some detailed PostgreSQL setup information. There is some mention of PostgreSQL 8.4 there - that could be update/fixed. And should we do example setup for PostgreSQL 10?
(for any of that we could raise a separate issue...)


settermjd:
> Issue #2343
> 
> Backport to 10.4 only.
> 
> Note: `linux_database_configuration.adoc` has some detailed PostgreSQL setup information. There is some mention of PostgreSQL 8.4 there - that could be update/fixed. And should we do example setup for PostgreSQL 10?
> (for any of that we could raise a separate issue...)

PostgreSQL. Now you're talking! Great to hear about the support.

phil-davis:
Backport is in PR #2345

PR #2361, opened 2020-02-28
add locking to the doc of wopi app

voroyam:
To explain to our customer what the lock icon is for.

### DIFF
--- a/modules/admin_manual/pages/enterprise/collaboration/msoffice-wopi-integration.adoc
+++ b/modules/admin_manual/pages/enterprise/collaboration/msoffice-wopi-integration.adoc
@@ -59,3 +59,5 @@ The idea is to lock the file so other users, who have access, can't make changes
 In other words: the feature ensures that you are the "master editor", your changes will always be the "master state". 
 Other users can make changes, e.g., with the desktop client, but those will create conflict files for them which can be resolved afterwards.
 When you close the document, Microsoft Office Office Online also has a feature to unlock the file again so it can be edited by others again as you are done with it.
+
+You can always click on the lock icon next to your file name and unlock it manually using the button in the sidebar.
--- a/modules/admin_manual/pages/enterprise/collaboration/msoffice-wopi-integration.adoc
+++ b/modules/admin_manual/pages/enterprise/collaboration/msoffice-wopi-integration.adoc
@@ -51,13 +51,13 @@ In the example above, only users in the `admin` group would be able to access Mi
 
 NOTE: If the key is not defined, then all users have access to WOPI.
 
-== Locking the document
+== Locking the Document
 
-If you open a document with Microsoft Office Online in ownCloud, then it makes use of the WebDAV file locking functionality, which we have in the server. 
-The idea is to lock the file so other users, who have access, can't make changes to the document, while you're editing it. 
+If you open a document with Microsoft Office Online in ownCloud, it makes use of the WebDAV file locking functionality, available in ownCloud server. 
+The idea is to lock the file so other users, who have access, can't make changes to the document while you're editing it. 
 
 In other words: the feature ensures that you are the "master editor", your changes will always be the "master state". 
-Other users can make changes, e.g., with the desktop client, but those will create conflict files for them which can be resolved afterwards.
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/enterprise/collaboration/msoffice-wopi-integration.adoc
```diff
@@ -50,3 +50,14 @@ Microsoft Office Online access can be restricted to users in a specific group, b
...
 NOTE: If the key is not defined, then all users have access to WOPI.
+
+== Locking the document
+
+If you open a document with Microsoft Office Online in ownCloud, then it makes use of the WebDAV file locking functionality, which we have in the server. 
+The idea is to lock the file so other users, who have access, can't make changes to the document, while you're editing it. 
+
+In other words: the feature ensures that you are the "master editor", your changes will always be the "master state". 
+Other users can make changes, e.g., with the desktop client, but those will create conflict files for them which can be resolved afterwards.
+When you close the document, Microsoft Office Office Online also has a feature to unlock the file again so it can be edited by others again as you are done with it.
```
phil-davis:
```suggestion
When you close the document, Microsoft Office Office Online will unlock the file so it can be edited by others.
```
I suggest a shorter sentence here. It is supposed to "just work".

phil-davis: Is it "Microsoft Office Office Online" or just "Microsoft Office Online"?

voroyam: :D good catch :)

voroyam:
> What happens when a user is editing, has a lock, then "goes into the tunnel" "loses their internet connection" or... - is there a lock timeout that will free the lock if connectivity is lost for >n minutes?

this question can be answered by @DeepDiver1975

modules/admin_manual/pages/enterprise/collaboration/msoffice-wopi-integration.adoc
voroyam: fixed

DeepDiver1975:
> this question can be answered by @DeepDiver1975

wopi locks files for 30 minutes - https://wopi.readthedocs.io/projects/wopirest/en/latest/concepts.html#term-lock

settermjd:
> What happens when a user is editing, has a lock, then "goes into the tunnel" "loses their internet connection" or... - is there a lock timeout that will free the lock if connectivity is lost for >n minutes?

I'll update the document, incorporating this and @DeepDiver1975's feedback.

voroyam:
@DeepDiver1975 so wopi locks the document for 30 min regardless if the editor of the document closes it or not?

I think this is new information, because I have been told by people at oC that if you close the document the document shall become unlocked, and if this is not the case - this is a bug.

phil-davis:
@voroyam @DeepDiver1975 see my suggested edit in PR #2459 - maybe that will help?

voroyam:
yeah, we would have to wait until we fix the current app, so this will be true again :)

But thanks for the PR

phil-davis:
I updated my comment in the PR. IMO we should first make the documentation clearer about the way the feature should work. Then when we know that there is a bug under investigation, we can also put a notice in the documentation about that, with a link to the bug issue etc. (e.g. straight away make another PR with the bug issue notice)

DeepDiver1975:
> I think this is new information, because I have been told by people at oC that if you close the document the document shall become unlocked, and if this is not the case - this is a bug.

Closing a doc will trigger the unlock which will free the lock.
30 minutes is the max lock time. a follow up refresh lock call will extend the lock for another 30 minutes

voroyam:
> Closing a doc will trigger the unlock which will free the lock.

this is currently not working = bug. right?

DeepDiver1975:
> this is currently not working

I'm not aware that this is not working. Please point me to a bug report and I can have a look.

voroyam:
https://github.com/owncloud/enterprise/issues/3918

voroyam:
currently reproducing it on the QA testing instance. So far - so same.

PR #2369, opened 2020-03-02
Document the options for the occ files:checksums:verify command

### LINKED ISSUE #2318: Add arguments/options to files:checksums:verify
The arguments for files:checksums:verify are missing on this page:

https://doc.owncloud.com/server/admin_manual/configuration/server/occ_command.html#file-operations

I have no idea where the corresponding file is, I would have changed it myself. But now it's all includes of files that are located somewhere.

Also why do we call it arguments and in the occ script it's options? No idea.

Here are the arguments:

Options:
  -r, --repair          Repair filecache-entry with mismatched checksums.
  -u, --user=USER       Specific user to check
  -p, --path=PATH       Path to check relative to user folder, i.e, relative to /john/files. e.g tree/apple [default: ""]

settermjd:
This fixes #2318.


modules/admin_manual/pages/configuration/server/occ_command/file_commands.adoc
```diff
@@ -45,6 +45,20 @@ Mismatch for thumbnails/9/2048-2048-max.png:
...
+[width="100%",cols="20%,70%",]
+|===
+|`-r, --repair`         
+|Repair filecache-entry with mismatched checksums.
+
+|`-u, --user=USER`       
+|Specific user to check.
+
+|`-p, --path=PATH`       
+|Path to check relative to user folder, i.e, relative to /john/files. e.g tree/apple [default: ""].
```
mmattel: You may want to look to https://doc.owncloud.org/server/10.4/admin_manual/configuration/server/occ_command.html#the-path-option
for examples which have better readability

settermjd: Thank you for the suggestion.

settermjd: Just extended the description from the one provided by the command itself.

mmattel:
@settermjd 
Taking the current sentence:
the `--path` value was "tree/apple", the command would check the ownCloud directory `/john/files/tree/apple`.
This sentence can be read the following `--path="tree/apple"`which is wrong.
It MUST be `--path="john/files/tree/apple"`. There MUST NOT be a leading `/` at the username.

My porposal would be to say: The `--path` value consists of the username followed by the fixed string `/files/` and then the path. e.g. `--path="john/files/tree/apple"` You can see more details of the path option in command `files:scan`.

settermjd:
> @settermjd
> Taking the current sentence:
> the `--path` value was "tree/apple", the command would check the ownCloud directory `/john/files/tree/apple`.
> This sentence can be read the following `--path="tree/apple"`which is wrong.
> It MUST be `--path="john/files/tree/apple"`. There MUST NOT be a leading `/` at the username.
> 
> My porposal would be to say: The `--path` value consists of the username followed by the fixed string `/files/` and then the path. e.g. `--path="john/files/tree/apple"` You can see more details of the path option in command `files:scan`.

@mmattel, by my reading of [the relevant code](https://github.com/owncloud/core/blob/master/apps/files/lib/Command/VerifyChecksums.php#L110), what you have said is incorrect.. Also, the command's official description shows that the patth value is relative to the user's folder. Actually, that description is misleading. It should, then, be "Path to check relative to the user's files folder,"

PR #2372, opened 2020-03-03
Add scenario tags to acceptance test docs

phil-davis:
There are some other tags that we use in the acceptance test scenarios. Document them.


phil-davis:
Backports in #2373 and #2374

PR #2375, opened 2020-03-03
Fix typos in developer docs

phil-davis:
Should be backported to 10.4 and 10.3


phil-davis:
I  will make backports...

PR #2379, opened 2020-03-03
Fix Content-Type typo

phil-davis:
Are these `cOntenT-Type` really supposed to be like that in `MultipartStreamTest.php`?


settermjd:
I can only assume that they are typos, given the spec.

PR #2334, opened 2020-02-24
[2326] Update deployment recommendations example for finding the PHP session.save_path value

settermjd:
This change replaces the existing examples showing how to retrieve the `session.save_path` value with a small awk/bash one-liner that works for PHP 7.3. This fixes #2326. **Please note:** b34d1cd is a cleanup commit, nothing more.

PR #2381, opened 2020-03-03
[10.3] [PR 2334] [2326] Update deployment recommendations example for finding the PHP session.save_path value

settermjd:
Backports #2334.


mmattel:
We had the same thing in mind and did it at the same time with just some seconds in between... :smile:

PR #2401, opened 2020-03-05
mention quick eval option

jnweiger:
and/or to master?


mmattel:
If you do not Back(forward) Port this to master (and I recommend to do this to 10.3 too), it will be rotated out with upcoming releases as we only keep the last 2 active.
For the future, please, if not fixing a very particular thing for a single branch, make the changes to master and then BP to where ever you think it applies.

PR #2422, opened 2020-03-09
Remove Redis from the default list of recommended system requirements

### LINKED ISSUE #2356: Missing Redis server
https://doc.owncloud.com/server/10.4/admin_manual/installation/deployment_recommendations.html#recommended-system-requirements-2

The description says redis server, but there is none on the picture

settermjd:
This PR removes Redis from the default list of recommended system requirements & fixes #2356.

### DIFF
--- a/modules/admin_manual/pages/installation/deployment_recommendations.adoc
+++ b/modules/admin_manual/pages/installation/deployment_recommendations.adoc
@@ -178,7 +178,6 @@ These recommendations apply if you meet the following criteria:
 * A cluster of two database servers.
 * Storage on an NFS server.
 * Authentication via an existing LDAP or Active Directory server.
-* A Redis server for file locking
 
 image:installation/deprecs-2.png[Network diagram for a mid-sized enterprise.]
 


tboerger:
Why should that be removed? Redis is a really important recommendation.

settermjd:
The original issue said that it wasn't a default system requirement.

tboerger:
It doesn't say that it's not recommended... It just says it's not part of the images...

settermjd:
True, but I wasn't aware that Redis was a core requirement. Yes, it makes perfect sense if you want to have any form or performance, but it's not mandatory.

tboerger:
It's not mandatory, but it's a really strong recommendation.

/cc @micbar

micbar:
@tboerger @settermjd Instead of arguing, we could just improve the docs around file locking.

Why is DB locking so slow -> not recommended
Why is APCu good for single server or DEV
What are the advantages of Redis?

How too set the redis TTL time. This part is relatively unknown. By default, it uses the value from php max_execution_time
After that time period, Redis cleans up all entries. For long running uploads, this could result in a "LockedException", the upload will fail.
So you have to set e.g.
`‘filelocking.ttl’ => 7200,`

settermjd:
> @tboerger @settermjd Instead of arguing, we could just improve the docs around file locking.
> 
> Why is DB locking so slow -> not recommended
> Why is APCu good for single server or DEV
> What are the advantages of Redis?
> 
> How too set the redis TTL time. This part is relatively unknown. By default, it uses the value from php max_execution_time
> After that time period, Redis cleans up all entries. For long running uploads, this could result in a "LockedException", the upload will fail.
> So you have to set e.g.
> `‘filelocking.ttl’ => 7200,`

Sounds fine. I'll open another PR and close this one.

PR #2426, opened 2020-03-09
Add a NOTE about volumes and about apps management

jnweiger:
There are some catches to the recommended Docker setup
* Admins can easily erase their data, when they kill their docker setup. 
* docker-compose down may change the installed apps

Made the language of the second NOTE less convoluted.

### DIFF
--- a/modules/admin_manual/pages/installation/docker/index.adoc
+++ b/modules/admin_manual/pages/installation/docker/index.adoc
@@ -81,18 +81,24 @@ below:
 
 [source,console,subs="attributes+"]
 ----
-Name                Command                       State             Ports
+Name                              Command                     State   Ports
 __________________________________________________________________________________________
-server_db_1         /usr/bin/entrypoint/bin/s …   Up                {std-port-mysql}/tcp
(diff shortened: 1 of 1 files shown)


tboerger:
The wording is not entirely right... and it doesn't really include best practices. The data is also with volumes part of the regular filetree, just look at /var/lib/docker/volumes.

Instead of doing something like `docker run -v ownclouddockerserver_files:/mnt ubuntu tar cf - -C /mnt . > files.tar` you can simply execute `docker cp`.

And why should somebody execute `docker image ls | grep ownclouddockerserver` to inspect volumes? There is a `docker volume` command...

jnweiger:
@tboerger thanks. Typo fixed. 

 I've modeled the commands while reading https://docs.docker.com/storage/volumes/#backup-restore-or-migrate-data-volumes -- Hoped that this might be good practice.

They don't mention /var/lib/docker -- Not sure why not. Maybe different storage drivers may handle volumes differently? With overlay2fs you are right: there is a nice filesystem tree within each volume under /var/lib/docker .

PR #2427, opened 2020-03-10
compose down with everything?

jnweiger:
Make the command agree with the description. --rmi all is needed for removing images, --volumes is needed for removing volumes.
(Or should we limit the docs to a simpler compose down? Not sure what is meant.)

(sorry, spotted this during first read, but somehow skipped it)


modules/admin_manual/pages/installation/docker/index.adoc
```diff
@@ -125,7 +125,7 @@ stored in `.env` earlier.
 
 Assuming you used docker-compose, as in the previous example, to stop
 the containers use `docker-compose stop`. Alternatively, use
-`docker-compose down` to stop and remove containers, along with the
+`docker-compose down --rmi all --volumes` to stop and remove containers, along with the
```
settermjd: I'm doubting it would, but would it also be worth also including `--remove-orphans`? Otherwise, looks good.

PR #2430, opened 2020-03-11
Small improvement for the post install script

mmattel:
This is a small improvement for the post install script.
Tested by doing, runs fine.

Backporting to 10.4 and 10.3 needed

### DIFF
--- a/modules/admin_manual/examples/installation/post-installation-steps.sh
+++ b/modules/admin_manual/examples/installation/post-installation-steps.sh
@@ -41,10 +41,10 @@ rootuser='root'
 # set the paths above accordingly. This script can link and set the proper rights and permissions
 # depending what you enter when running the script.
 
-# When the instance is setup either post a fresh install or after an upgrade, run this script again but
-# only for securing ".htaccess files". This sets the appropriate ownership and permission for them.
+# When the instance is setup, either post a fresh install or, after an upgrade, run this script again - but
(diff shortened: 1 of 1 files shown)


modules/admin_manual/examples/installation/post-installation-steps.sh
```diff
@@ -37,22 +37,42 @@ rootuser='root'
...
 # depending what you enter when running the script.
-# You have to run this script twice, one time to prepare installation and one time post installation
 
-# In case you upgrade an existing installation, your original directory will be renamed including a timestamp 
+# When the instance is setup, either post a fresh install or, after an upgrade, run this script again - but
+# only for securing ".htaccess files". This sets the appropriate ownership and permissions for them.
 
-# Example input
-# New install using mkdir:     n/n/n/n (create missing directories, setup permissions and ownership)
+# In case you upgrade an existing installation, your original directory will be renamed, including a timestamp.
```
phil-davis: This was existing text, so I will just comment for the future that this sort of use of "In case" sounds unusual. "In case you..." is better and more simply written "If you..."

phil-davis:
@mmattel please make backports

modules/admin_manual/examples/installation/post-installation-steps.sh
mmattel: Thanks, will keep it in mind 👍

PR #2438, opened 2020-03-13
drop old linux platforms, add new.

jnweiger:
Started an update of that page. It got old. I suppose more should be updated here...
It is referenced in installation instructions and README files of customer facing documentation.


jnweiger:
{supported-php-versions} includes 7.0 -- not sure where that comes from.

phil-davis:
Forward port to master is #2440

PR #2440, opened 2020-03-14
drop old linux platforms, add new.

phil-davis:
Forward port of #2438 to master


mmattel:
@jnweiger FYI this is a forward port from 10.4 to master to prevent issues.
Any changes you think of should be made post this PR/merge to master and then backported to eg 10.4 ect (and not the other way round)

phil-davis:
@jnweiger this is simply a forward-port to master of what is already merged to 10.4 branch from your PR.

We need to merge this to get master the same as 10.4

Then you (or someone) can start another PR to make more changes.

PR #2446, opened 2020-03-16
Add known issue with password policy to release notes

pmaier1:


modules/admin_manual/pages/release_notes.adoc
```diff
@@ -116,8 +116,7 @@ Server 10.4 is thoroughly tested against these database versions and proven to w
 
 === Known Issues
 
-Currently, there are no known issues with ownCloud Server 10.4. 
-This section will be updated if problems are discovered.
+- When the **Password Policy** public link expiration policies (_days maximum until link expires if password is set_ and _days maximum until link expires if password is not set_) are enabled, sharing with users/groups does not work any more. A fix for this issue is being worked on. If you have already upgraded to 10.4.0, please disable both options until the fix is available and deployed on your system. https://github.com/owncloud/password_policy/issues/286[#286]
```
phil-davis:
```suggestion
- When the **Password Policy** public link expiration policy (_days maximum until link expires if password is not set_) is enabled, sharing with users/groups does not work any more. A fix for this issue is being worked on. If you have already upgraded to 10.4.0, please disable this option until the fix is available and deployed on your system. https://github.com/owncloud/password_policy/issues/287[#287]
```

See https://github.com/owncloud/password_policy/issues/287#issuecomment-597442075 - it is just that one setting that causes the breakage. The other setting is not causing a problem.

The issue for the breakage is 287.

Issue 286 is about annoying stuff with the UI not being so friendly when creating/adjusting a public link when there are password policy public link expiration policies enabled. IMO that is actually and existing "feature" (I will double-check that against 10.3.2)

pmaier1: Thanks for clarification :+1:

settermjd:
> LGTM.
> Will readers realize that if you do not have the password policy app installed, then this is not an issue at all?

Good point, which is worth drawing attention to, in case users who both aren't aware and don't have the app installed spend time investigating only to find out that they didn't need to be concerned about it.

pmaier1:
Thanks!

pmaier1:
Please backport @settermjd

phil-davis:
Backports to
10.4 #2489
10.3 #2490

PR #2453, opened 2020-03-17
[10.3] Add missing text substitutions and enclose attribute in single quotes

settermjd:
This PR:

- Adds missing text substitutions
- Encloses the `minimum-php-version` attribute value in single quotes, as the value, when
substituted, was being truncated to `7`, and should have been rendered as `7.0`; and
- Renames the attribute `minimum-php-version-shortcode` to `minimum-php-version-short-code`, as that is how it is referenced in the docs.

PR #2454, opened 2020-03-17
[10.4] Rename site attribute

settermjd:
This PR renames the site-wide attribute to reflect how it's used in the docs, fixing a render bug.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -64,7 +64,7 @@ asciidoc:
     php-supported-versions-url: https://secure.php.net/supported-versions.php
     http-status-codes-base-url: https://developer.mozilla.org/en-US/docs/Web/Status
     minimum-php-version: 7.1
-    minimum-php-version-shortcode: 71
+    minimum-php-version-short-code: 71
     recommended-php-version: 7.3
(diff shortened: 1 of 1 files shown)


site.yml
```diff
@@ -64,7 +64,7 @@ asciidoc:
     php-supported-versions-url: https://secure.php.net/supported-versions.php
     http-status-codes-base-url: https://developer.mozilla.org/en-US/docs/Web/Status
     minimum-php-version: 7.1
```
phil-davis: Do you also need to do:
```suggestion
    minimum-php-version: '7.1'
```
similar to #2453

settermjd: This isn't required. It's only when a value, that can be interpreted as a number, not a string, is used and it ends with `0`. I'm sure that I saw an explicit reference to this in the docs, but I'm unable to find it at the moment, in either the AsciiDoc User Manual or the Antora docs.

settermjd: I just asked in the asciidoctor forum, and hope to find out where the reference is (or contribute it if it's not available).

PR #2455, opened 2020-03-17
Rename site attribute to fix render bug

settermjd:
This PR renames the site-wide attribute to reflect how it's used in the docs, fixing a render bug.


site.yml
```diff
@@ -64,7 +64,7 @@ asciidoc:
     php-supported-versions-url: https://secure.php.net/supported-versions.php
     http-status-codes-base-url: https://developer.mozilla.org/en-US/docs/Web/Status
     minimum-php-version: 7.1
```
phil-davis: Do you also need to do:
```suggestion
    minimum-php-version: '7.1'
```
similar to #2453

PR #2459, opened 2020-03-17
Clarify the description of Office Online unlocking

phil-davis:
Maybe writing a more simple sentence will avoid confusion?

I am intending to describe/document more clearly/simply what the required/intended behaviour is. Then of course if the software does not actually do it, then there is a bug to be investigated and fixed.

Will need backport to 10.4 and 10.3

### DIFF
--- a/modules/admin_manual/pages/enterprise/collaboration/msoffice-wopi-integration.adoc
+++ b/modules/admin_manual/pages/enterprise/collaboration/msoffice-wopi-integration.adoc
@@ -59,7 +59,7 @@ The idea is to lock the file so other users, who have access, can't make changes
 
 In other words: the feature ensures that you are the "master editor", your changes will always be the "master state". 
 Other users can make changes, e.g., with the desktop client, but those will create conflict files for them, which can be resolved afterward.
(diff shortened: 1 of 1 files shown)


phil-davis:
@voroyam does this seem clearer?

@DeepDiver1975 IMO this is what happens. Am I correct?

DeepDiver1975:
> @DeepDiver1975 IMO this is what happens. Am I correct?

yes - this is correct - but I don't get the point on documenting office online server's behavior ..... anyway you for sure have a reason to do so ...

phil-davis:
I am just clarifying and simplifying the existing documentation. The way it is currently written is a bit ambiguous.

DeepDiver1975:
> I am just clarifying and simplifying the existing documentation. The way it is currently written is a bit ambiguous.

no worries @phil-davis - I'm just wondering in a general manner ;-)

phil-davis:
@voroyam @settermjd 
Now it has been found that the issue with files getting left locked after closing them, that issue is a known problem with Office Online and Chrome. IMO we do not need to be document stuff like that - but if someone wants to then it can be done in a separate PR.

Please review this PR to see if the change makes the meaning easier to understand.

voroyam:
Big Thanks for making this PR @phil-davis

PR #2470, opened 2020-03-18
minor fixes to ubuntu install doc

mmattel:
These changes and one link fix have whyever slipped thru the last update 🤦‍♂ , fixing it  now.

Backport to 10.3 and 10.4 needed

PR #2479, opened 2020-03-19
minor fixes to Ubuntu install doc

phil-davis:
These changes and one link fix have whyever slipped thru the last update 🤦‍♂ , fixing it now.

Backport to 10.3 and 10.4 needed

Info: successor of #2470 because drone was stuck


phil-davis:
I will make backports...

Issue #2484, opened 2020-03-21
Possible incorrect description for occ files:transfer-ownership


thommierother:
I am struggling  with the transfer of ownership for data from a subdirectory of one user to another in OC 10.4.

The documentation at https://doc.owncloud.com/server/admin_manual/configuration/server/occ_command.html#file-operations says:

`sudo -u www-data php occ files:transfer-ownership --path="folder/to/move" <source-user> <destination-user>
`
And for the --path option:

```
["user_id/files/path"
"user_id/files/mount_name"
"user_id/files/mount_name/path"
```

When testing on 10.4 I get:

```
root@www:/var/www/owncloud# sudo -u www-data php occ files:transfer-ownership --path="thommie4/files/gruenerheiner" thommie4 gruenerheiner
Unknown path provided: thommie4/files/gruenerheiner
root@www:/var/www/owncloud# sudo -u www-data php occ files:transfer-ownership --path=thommie4/files/gruenerheiner thommie4 gruenerheiner
Unknown path provided: thommie4/files/gruenerheiner
root@www:/var/www/owncloud# sudo -u www-data php occ files:transfer-ownership --path="/thommie4/gruenerheiner" thommie4 gruenerheinerUnknown path provided: thommie4/gruenerheiner
```
What finally works is the following:

`sudo -u www-data php occ files:transfer-ownership --path="gruenerheiner" thommie4 gruenerheiner`

Do we have documentation or a code error?

settermjd:
Thanks for raising this, @thommierother. I'll investigate and find out.

settermjd:
I had a read through the documentation and don't see any reference to an array of paths. Can you show me where you saw that please, @thommierother?

thommierother:
The --path Option
https://doc.owncloud.com/server/admin_manual/configuration/server/occ_command.html#file-operations

> When using the --path option, the path must be in one of the following formats:
> 
> "user_id/files/path"
> "user_id/files/mount_name"
> "user_id/files/mount_name/path"
> 
>

settermjd:
Ah, I see. That section isn't meant to suggest that you can use all of them. They are examples of what you could use. I'll have think through how I can remove any ambiguity.

settermjd:
Got it. I'll change the source block to a bullet list instead. The change will be in a PR later today.

PR #2485, opened 2020-03-23
Slight update to the files:transfer-ownership documentation

settermjd:
When reading through the files:transfer-ownership documentation, while
researching #2484, I found that the wording in that section could do
with a small update. This is that update, which has the intent of making
it quicker and easier to read.


phil-davis:
@settermjd I guess this needs backport to both 10.4 and 10.3 branches

PR #2503, opened 2020-03-30
Update all external links to follow the style guide

settermjd:


mmattel:
To which branches do we need backporting ?

settermjd:
> To which branches do we need backporting ?

I'd suggest both 10.4 & 10.3.

phil-davis:
Agree - 10.4 and 10.3. The backport might be different in 10.3 because some stuff might not exist there. But @settermjd will find that out if he gets conflicts during the backport

Issue #2522, opened 2020-04-01
drone CI build messages


phil-davis:
Running drone CI  from the master branch there are warnings like https://drone.owncloud.com/owncloud/docs/7574/1/5
```
asciidoctor: WARNING: ios_settings.adoc: line 78: section title out of sequence: expected level 1, got level 2
asciidoctor: WARNING: skipping reference to missing attribute: session_path
asciidoctor: WARNING: skipping reference to missing attribute: session_path
asciidoctor: WARNING: skipping reference to missing attribute: latest-download-version
asciidoctor: WARNING: skipping reference to missing attribute: request-base-path
asciidoctor: WARNING: skipping reference to missing attribute: request_body
asciidoctor: WARNING: skipping reference to missing attribute: server_uri
asciidoctor: WARNING: skipping reference to missing attribute: username
asciidoctor: WARNING: skipping reference to missing attribute: password
asciidoctor: WARNING: skipping reference to missing attribute: request-base-path
asciidoctor: WARNING: admin_audit.adoc: line 9: section title out of sequence: expected level 1, got level 2
asciidoctor: WARNING: admin_audit.adoc: line 13: section title out of sequence: expected level 1, got level 2
asciidoctor: WARNING: admin_audit.adoc: line 106: section title out of sequence: expected level 1, got level 2
asciidoctor: WARNING: skipping reference to missing attribute: splunk-universal-forwarder-docs-url
asciidoctor: WARNING: skipping reference to missing attribute: php-datetime-url
asciidoctor: WARNING: skipping reference to missing attribute: id
asciidoctor: WARNING: skipping reference to missing attribute: id
asciidoctor: WARNING: skipping reference to missing attribute: id
asciidoctor: WARNING: skipping reference to missing attribute: session_path
asciidoctor: WARNING: skipping reference to missing attribute: session_path
asciidoctor: WARNING: skipping reference to missing attribute: request-base-path
asciidoctor: WARNING: skipping reference to missing attribute: request_body
asciidoctor: WARNING: file_properties.adoc: line 12: id assigned to block already in use: file-property-getcontentlength
asciidoctor: WARNING: file_properties.adoc: line 73: id assigned to block already in use: file-property-size
asciidoctor: WARNING: skipping reference to missing attribute: server_uri
asciidoctor: WARNING: skipping reference to missing attribute: username
asciidoctor: WARNING: skipping reference to missing attribute: password
asciidoctor: WARNING: skipping reference to missing attribute: request-base-path
[...]

PR #2542, opened 2020-04-07
Fix build errors in master

settermjd:
Because the leveloffset attribute hadn't been set, the section levels in the included files were set too high.

### DIFF
--- a/modules/developer_manual/pages/_partials/webdav_api/search/file_properties.adoc
+++ b/modules/developer_manual/pages/_partials/webdav_api/search/file_properties.adoc
@@ -8,11 +8,10 @@
 |Description
 |Namespace
 
-a|[#file-property-getcontentlength]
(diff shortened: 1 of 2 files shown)


settermjd:
> There are still lots of messages like:
> 
> ```
> asciidoctor: ERROR: file_lifecycle.adoc: line 1: level 0 sections can only be used when doctype is book
> ```
> 
> But I guess this PR is just fixing the problems in the files that it is touching.

I don't see that locally. To be honest, it's a slow process to fix these types of issues, as there are so many content sources from which the errors and warnings can originate. I'll see if I can find what I need to do to replicate that error locally.

phil-davis:
The drone log is https://drone.owncloud.com/owncloud/docs/7628/1/5

settermjd:
This relates to #2522.

PR #2543, opened 2020-04-07
Fix build errors in 10.4

settermjd:
This fixes the build errors and a majority of the warnings in the 10.4 branch. 

There are, unfortunately, several that I've not been able to resolve,, despite grepping the entire ownCloud documentation; they are listed below. 

```console
asciidoctor: WARNING: skipping reference to missing attribute: password
asciidoctor: WARNING: skipping reference to missing attribute: request_body
asciidoctor: WARNING: skipping reference to missing attribute: server_uri
asciidoctor: WARNING: skipping reference to missing attribute: session_path
asciidoctor: WARNING: skipping reference to missing attribute: username
```

I've asked in [the Antora users forum](https://gitter.im/antora/users) and am waiting to get feedback so that I can resolve the remaining ones.

PR #2544, opened 2020-04-07
Fix leveloffset errors

settermjd:
This fixes the build errors and a majority of the warnings in the 10.3 branch. It has the same issue with build warnings as #2543.


settermjd:
This relates to #2522.

PR #2547, opened 2020-04-07
Update ubuntu_18_04.adoc

jnweiger:
Rant


jnweiger:
demo for interns only.

PR #2548, opened 2020-04-07
Add ssl-cert

NannaBarz:
hetzner cloud 18.04 apache is installed without ssl-cert


jnweiger:
@settermjd also into master?

PR #2550, opened 2020-04-08
Render the OAuth2 flow Ascii art diagram as an image

settermjd:
While not strictly necessary, it might lend a greater level of professionalism to the documentation if the ascii art is rendered as an image. It's trivial to enable support for this to happen, so it seemed worth doing. Below is an example of what the Ascii art rendered as an image looks like.

Screenshot_2020-04-08 OAuth2


michaelstingl:
@settermjd we already use `plantuml` in some places. Could you also use `plantuml` instead of `ditaa` ?

settermjd:
> @settermjd we already use `plantuml` in some places. Could you also use `plantuml` instead of `ditaa` ?

If the underlying source code block is block is changed to PlantUML, yes, but the source code block contains Ascii art.

settermjd:
> Looks promising. I need to have another look after merge and deploy. 👍

Okay.

mmattel:
@settermjd @michaelstingl 
There have been conflicts, I solved them, CI is now green and the PR can be merged.
Backporting to 10.4 and 10.3

PR #2552, opened 2020-04-08
example was 10.2.1 and changed to 10.4.0

GeraldLeikam:
This is 10.4.0 doumentation branch and there for instructions should use 10.4.0

### DIFF
--- a/modules/admin_manual/pages/installation/ubuntu_18_04.adoc
+++ b/modules/admin_manual/pages/installation/ubuntu_18_04.adoc
@@ -4,6 +4,7 @@
 :webserver-user: www-data
 :webserver-group: www-data
 :install-directory: /var/www/owncloud
+:oc-version: 10.4.0
 
 This is an ultra-short guide to installing ownCloud on a fresh installation of Ubuntu 18.04.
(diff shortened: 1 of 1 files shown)


phil-davis:
IMO this will need forward-port to `master` and a similar change in 10.3 branch.

modules/admin_manual/pages/installation/ubuntu_18_04.adoc
```diff
@@ -4,6 +4,7 @@
 :webserver-user: www-data
 :webserver-group: www-data
 :install-directory: /var/www/owncloud
+:oc-version: 10.4.0
```
settermjd, line 7: There are a number of predefined constants (or site-wide attributes) defined at the bottom of `site.yml`. These include `current-version`, which always tracks the latest stable ownCloud version (but only includes the major and minor version). Can you move `oc-version` to site.yml so that it's easier to keep track of, pls?

phil-davis:
@GeraldLeikam thanks for this. It got implemented in master PR #2763 and "backports" of that to 10.5 10.4 and 10.3 branches.

PR #2554, opened 2020-04-08
replaced variables with real values

iHerrmann77:
ascii doc variables don´t work with shell scripts


settermjd:
> ascii doc variables don´t work with shell scripts

They do, actually. What is missing is `,subs="attributes+"` in the source code block header. I'll create a PR to add it.

settermjd:
With the changes in #2589 (an others), this PR no longer seems required.

settermjd:
Closing in favour of #2589.

Issue #2566, opened 2020-04-16
Script on install document causes Ubuntu system to be unbootable


cgarrovillo:
A part on the owncloud Ubuntu 18.04 installation guide provides a script that breaks ALL permissions throughout the whole system, causing the system to be unbootable

## What Needs to be Documented?

The "Finalise the Installation" part of the [owncloud Ubuntu 18.04 installation](https://doc.owncloud.org/server/10.4/admin_manual/installation/ubuntu_18_04.html) executes a chmod at the very root of the file system and causes the whole system to break

## Why Should This Change Be Made? (Optional)

This is a critical issue. This script needs to be fixed

## What Type Of Content Change Is This? (Optional)
- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ X ] Bug Fix to Existing Content

## Which Manual Does This Relate To? (Optional)
- [ X ] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

ibaranov-cp:
+1 this happens to me as well. Suspect it is the variables at the top of the file finalise-install.sh

https://github.com/owncloud/docs/blob/163d0a4db9135b5c2289e22b497ab0f9cb825462/modules/admin_manual/examples/installation/ubuntu/18.04/finalise-install.sh#L8

Dangerous to leave this kind of stuff in a guide for new folks

cgarrovillo:
I should've known tbh. Even the guide uses an old link to ownCloud 10.2 or something like that. Didn't expect something as destructive as this though.

Looking at the script too, it looks like it's specifically setup to break ALL permissions and set it to ownCloud user. No idea if it is by design but surely destructive as hell.

https://github.com/owncloud/docs/blob/163d0a4db9135b5c2289e22b497ab0f9cb825462/modules/admin_manual/examples/installation/ubuntu/18.04/finalise-install.sh#L22

Imagine deploying this for a small client. Someone should claim responsibility for this

phil-davis:
The variables displayed there are  in the top of https://raw.githubusercontent.com/owncloud/docs/master/modules/admin_manual/pages/installation/ubuntu_18_04.adoc

```
:webserver-user: www-data
:webserver-group: www-data
:install-directory: /var/www/owncloud
```

But they are not getting substituted in the code/script section when the docs are rendered.

[...]

PR #2589, opened 2020-04-20
Update finalise-install script

settermjd:
The intent here is to simplify some of the script, so that it's more repeatable, make the permission settings more consistent, make the script that much more verbose and reliable. This PR is attempting to cover all of the following feedback:

- https://github.com/owncloud/docs/issues/2566#issuecomment-615398150
- https://github.com/owncloud/docs/issues/2566#issuecomment-615501286

Fwiw, I don't know how it would break networking, as mentioned in https://github.com/owncloud/docs/issues/2566#issuecomment-615501286.


phil-davis:
Needs conflicts to be resolved.

settermjd:
> Needs conflicts to be resolved.

done!

settermjd:
It does, yes, but they shouldn't have to be hard-coded.

PR #2602, opened 2020-04-22
Remove orphaned NGINX configuration documentation

settermjd:


phil-davis:
I guess this needs to be applied in branches also?
@settermjd

PR #2619, opened 2020-04-25
Update user_auth_ldap.adoc

NannaBarz:
The point was confusing for me. I missed the hint that you have to create/use a attribute in the active directory not in ownCloud.


voroyam:
ehm - stupid questions - what about users who have openLDAP as a user backend? :D

PR #2621, opened 2020-04-27
[10.4] Update user_auth_ldap.adoc

phil-davis:
Backport #2619 to 10.4 branch


phil-davis:
@micbar I guess nobody is approving the backports to 10.4 and 10.3 these days.
Until/unless something happens about changing the doc versions publishing, the backports need to be done and somebody needs to be the "approval monkey".

Note: approving backports should not need review of the content changes - just:
- check that the changes in the backport match what was already approved and merged to master
- check that the changes are actually applicable to the version they are being backported to

PR #2624, opened 2020-04-28
Improved installation manual.

voroyam:
Updated server version

Removed permissions script which caused confusion and troubles.


voroyam:
> I guess the actual script `installation/ubuntu/18.04/finalise-install.sh` should also be deleted from the repo.

done

voroyam:
thanks for the review @micbar

voroyam:
This has to be backported to the relevant branches.

phil-davis:
Backports to all 4 old branches done.
Needs to be backported that far so that the script really gets removed from everywhere.

PR #2632, opened 2020-05-01
user_external link fix

ho4ho:


phil-davis:
Backports in above linked PRs.
I made all 4 backports because this is an existing mistake that is wrong everywhere and it only takes a few minutes to make the extra backports.

PR #2639, opened 2020-05-04
Add known issue 37324 about user names with only numbers

phil-davis:
IMO this issue only effects sharing on the webUI.

Using other clients to share with a user `123456` should still work in `10.4.0` and `10.4.1` - @HanaGemela somebody could check that.


phil-davis:
Done in #2638

PR #2645, opened 2020-05-12
Incorporate changes from core for config.sample.php

mmattel:
Incorporate the latest changes from core for `config.sample.php`
Triggered by https://github.com/owncloud/core/pull/35969

Note, I was not aware of the removal of `knowledgebaseenabled` key in core. Therefore it is in this PR as it reflects all the changes made from last config-to-docs run.

Needs backporting to 10.4 (maybe 10.3, pls advise)


phil-davis:
Both of these changes are "fixing" doc things that apply the same in 10.3 as well as 10.4 - so please backport to both branches and I can approve.

PR #2649, opened 2020-05-18
Use realistic user names in acceptance tests

phil-davis:
Document the typical user names that we use in acceptance tests.
This was implemented in core PR https://github.com/owncloud/core/pull/37397


phil-davis:
This change will only be noticeable in the core repo from today onwards (i.e. 10.5.0 onwards)
So I guess the "old" 10.3 and 10.4 documentation does not need to mention this - if someone tries to branch off a 10.3 or 10.4 tag in the repo then these "realistic" user names will not quite work.

PR #2661, opened 2020-05-29
Add 10.5 documentation

dschmidt:


phil-davis:
I do not understand this. I thought that `10.5` is a new branch that will stay around for "a long time", just like `10.4` `10.3` ...

So why is there a PR to merge it back to master?

dschmidt:
Yes, it's supposed to be merged to master once (for adding itself to the site.yml etc.).
... or when 10.5 documentation is updated and that should go to master documentation as well.

master is the branch that takes care of building everything - no matter which version branch something was pushed to, so the tooling and the version references need to be updated there.

P.S.: I still need to checkout why it's failing - maybe there's still a minor flaw in this

phil-davis:
> Yes, it's supposed to be merged to master once

I guess that I expected an "ordinary" PR from some "ordinary" named branch that updated stuff in master. Then create 10.5 from that point of master.

When this PR merged, I suspect that the "PR cleanup bot" will automagically delete the branch, which will be a bit sad.

dschmidt:
Ah, makes sense. Then it needs to be repushed - or as you said, an ordinary branch should have been created from the start. 
Those are the nitty gritty details that are really hard to think of. I'll make sure to update the documentation once we have sorted this out completely.

dschmidt:
The issue was that the 10.5 branch referenced itself. Fixed that and updated documentation. Follow up PR:
https://github.com/owncloud/docs/pull/2662

jnweiger:
https://doc.owncloud.com/server/10.5/user_manual/ownCloud_User_Manual.pdf did not arrive yet. @dschmidt

PR #2662, opened 2020-05-29
Add 10.5 to master

dschmidt:
See https://github.com/owncloud/docs/compare/10.5...add-10.5?expand=1 for changes compared to 10.5 branch.

I think this is correct now and tried to adjust the documentation accordingly.
If you have ideas to make it easier, let me know. The problem is that part of the changes are shared between master and 10.x and 10.x needs one individual change, that makes the instructions look a bit confusing ... (compare the link above)

edit:
So the expected outcome is roughly:
commit shared in 10.x and master branches:
- updating versions in `site.yml` and `.drone.star`

only master: 
- adding 10.x branch to `content.sources`

only in 10.x: 
- setting version in `antora.yml`

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -4,7 +4,7 @@ def main(ctx):
     # Version shown as latest in generated documentations
     # It's fine that this is out of date in version branches, usually just needs
     # adjustment in master/deployment_branch when a new version is added to site.yml
-    latest_version = "10.4"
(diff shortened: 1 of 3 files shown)


jnweiger:
https://doc.owncloud.com/server/10.5/user_manual/ownCloud_User_Manual.pdf did not arrive after 3 hours.

PR #2668, opened 2020-06-02
Upload pdfs from base_branch

dschmidt:
My latest changes removed pdf upload from branches, which apparently was not a good idea.

The html documentation can be built centrally from master - pdfs need to be built in and uploaded from their respective branches. This is okay, because they don't cross reference each other between different versions AFAICT and so we don't have constant backporting there.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -148,8 +148,9 @@ def build(ctx, latest_version, deployment_branch, base_branch):
                     "event": [
                         "push",
                     ],
-                    "branch": [
-                        deployment_branch,
+                    "ref": [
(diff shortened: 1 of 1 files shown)


dschmidt:
Different approach would be building everything from master (or rather extend scripting/pipelines  to checkout multiple branches and then use scripts from master to build pdfs from those checkouts. 

That would be way more involved, should we try my current approach and see how it goes for now? If we find that we still do backports all the time, we can still do the more cumbersome approach. (And we'd know if it would be really helpful or not)

Issue #2674, opened 2020-06-11
Add doc of files_external:list --importable-format


phil-davis:
## What Needs to be Documented?
core PR https://github.com/owncloud/core/pull/37513 for core issue https://github.com/owncloud/core/issues/37054 proposes adding a new option:

`occ files_external:list --importable-format`

Add it to existing documentation of the command.

mmattel:
Closing, PR´s got merged

PR #2675, opened 2020-06-11
files_external:list --importable-format documentation

phil-davis:
Fixes issue #2674
core code changes were merged in PR https://github.com/owncloud/core/pull/37513 for 10.5.0

Backport to 10.5 only.

### DIFF
--- a/modules/admin_manual/pages/configuration/server/occ_command/files_external_commands.adoc
+++ b/modules/admin_manual/pages/configuration/server/occ_command/files_external_commands.adoc
@@ -47,12 +47,13 @@ files_external:list [--show-password] [--full] [-a|--all] [-s|--short] [--] [<us
 
 [width="100%",cols="20%,70%",]
 |===
-| `--show-password`   | User to add the mount configurations for, if not set the mount will
-be added as system mount.
-| `--full`            | Don't save the imported mounts, only list the new mounts.
-| `-a, --all`         | Show both system-wide mounts and all personal mounts.
-| `-s, --short`       | Show only a reduced mount info.
-| `--output=[OUTPUT]` | The output format to use (`plain`, `json` or `json_pretty`, default is `plain`).
+| `--show-password`        | Show passwords and secrets
+| `--mount-options         | Show all mount options independent if they are set to their default value or not
+| `--full`                 | Don't truncate long values in table output
+| `-a, --all`              | Show both system-wide mounts and all personal mounts.
+| `-s, --short`            | Show only a reduced mount info.
+| `-i, --importable-format | Provide output values in a format compatible with files_external:import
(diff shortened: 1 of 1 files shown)


mmattel:
From core:
```
A new option to files_external:list has been added - -i or --importable-format - when this
is specified and a JSON output is requested, the items above are corrected in the JSON
to be in a format that can be used for import.
```
This reads for me that `-i` needs a proper output format (`json` or `json_pretty` ?) to be defined - but not plain which is used by default.

phil-davis:
@mmattel any output format is valid. If used with a "normal" list then the text columns for `Storage`  and `Authentication Type` show the "technical" values rather than the "human readable" values.

See https://github.com/owncloud/core/pull/37513#issuecomment-642461919 - @jvillafanez also asked about this. It's good to see that reviewers are looking :)

mmattel:
I have tested it just now.
I propose to add a text below the table like:

...
TIP: When using option `--importable-format`, consider using `--show-password` additionally. Else the user and/or admin has to change all mount passwords post importing manually to make a mount fully functional.
...

This is because if you miss `--show-password`, an import needs post work for either the user and/or the admin !

phil-davis:
> This is because if you miss --show-password, an import needs post work for either the user and/or the admin !

OK, I will look at it. The "problem" here is that the new `--importable-format` option has only been added so that `files_external:export` can switch it on. The availability to someone manually doing `files_external:list --importable-format` is really an accidental side-effect. There is no intention for `files_external:list --importable-format` by itself to produce the output ready for `files_external:import` - for that there is the command `files_external:export` which puts together all the pieces that are needed.

mmattel:
I think that the current implementation has a positive sideffect.
You have to mandatory add another option to make it fully work for importing.
This is a security measure ! :smile:
The only thing is, it must be documented.
Just thinking about, instead of a tip, we can write `NOTE:` and start with: For security measures you have to add...

phil-davis:
@mmattel I wrote a "Note:" section

modules/admin_manual/pages/configuration/server/occ_command/files_external_commands.adoc
```diff
@@ -68,6 +69,12 @@ be added as system mount.
 +----------+------------------+----------+
 ----
 
+NOTE: The `--importable-format` option helps make visible the technical mount settings.
```
mmattel, line 72: I am not a native speaker but this sounds better to me...
`helps make visible the technical mount settings.` --> `helps to make the technical mount settings visible.`

phil-davis: I should read what I write! Fixed.

phil-davis:
Backport to 10.5 is PR #2678

PR #2676, opened 2020-06-11
Document how to replace usernames in acceptance tests

phil-davis:
Documents the implementation of issue https://github.com/owncloud/core/issues/33596

Backport to 10.5 (the possibility is only there from core 10.5 onwards)


phil-davis:
Backport to 10.5 is in #2679

PR #2684, opened 2020-06-18
Move note about supported desktop Linux to desktop section

michaelstingl:
Note is the policy for desktop Linux and was introduced by the desktop team.

No idea about the policy for supported server Linux versions. /cc @pmaier1

### DIFF
--- a/modules/admin_manual/pages/installation/system_requirements.adoc
+++ b/modules/admin_manual/pages/installation/system_requirements.adoc
@@ -62,11 +62,6 @@ For _best performance_, _stability_, _support_, and _full functionality_ we offi
 |* {supported-php-versions}
 |===
 
-[NOTE]
-====
-For Linux distributions, we support, if technically feasible, the latest two versions per platform and the previous LTS Version.
-====
-
 === Hypervisors
 
 * Hyper-V
@@ -87,6 +82,11 @@ The latest stable client supports the platforms listed below:
 ** openSUSE Leap 42.2 & 42.3 & 15.0 & 15.1 & 15.2
 * macOS X 10.10+ (*64-bit only*)
 * Microsoft Windows 7+
+
+[NOTE]
+====
+For Linux distributions, we support, if technically feasible, the latest two versions per platform and the previous LTS Version.
+====
 
 === Web Browser
 


voroyam:
Why move it if you are not sure? Why not copy?

This is the policy for server versions. It's good that it's also applies to the client.

I would not move it. Copy it.

mmattel:
> Why move it if you are not sure? Why not copy?

Client is different to server.
Server says: `Ubuntu 16.04 and 18.04` (where we could add LTS btw)
Client says: `Ubuntu 18.04 & 19.04 & 19.10` (for sure there will be more soon)
Client does not need PHP, MySQL ect. which is by far more restrictive than server

voroyam:
> > Why move it if you are not sure? Why not copy?
> 
> Client is different to server.
> Server says: `Ubuntu 16.04 and 18.04` (where we could add LTS btw)
> Client says: `Ubuntu 18.04 & 19.04 & 19.10` (for sure there will be more soon)
> Client does not need PHP, MySQL ect. which is by far more restrictive than server

The note was written at the server section for a reason. I was in the meeting discussing this. It was written for the server, for the linux distribution support.

**Last 2 minor and 1 LTS.**
LTS for Ubuntu is currently 20.04
Last 2 Minor is Ubuntu 19.10 and ... nothing. At least according to this site:
https://wiki.ubuntu.com/Releases

I disagree with the decision to move = delete this note about our policy from the server section and paste = add it to the client section.

@michaelstingl please provide an argument for moving this instead of copying.

mmattel:
Before we do this change, let us think in gerneral about the Ubuntu releases for desktop and server. It is confusing to say „ the latest two and previous LTS“ and then writing a list of exact releases. ATM, we miss competely 20.04 LTS ?? And, server and desktop have different requirements, as written above. Looking at that, the change in this PR is in terms of clarity correct.
I see the point you are looking at. Maybe we need to rewrite both the sever and desktop part to be future proof.

michaelstingl:
> This is the policy for server versions. It's good that it's also applies to the client.

This policy was originally added for the desktop client. Seems it was later hijacked/adopted for the server (intentionally or accidentally)

> @michaelstingl please provide an argument for moving this instead of copying.

I personally don't have this information from the server, so I'd need confirmation from @pmaier1 that this policy is really valid.

> ATM, we miss competely 20.04 LTS ??

Desktop 2.6.3 builds support 20.04, and updating the list here is part of the release process.

pmaier1:
Thanks @michaelstingl. We can't apply such a general policy to ownCloud Server as we have dependencies on the environment like PHP versions. Therefore we do not support specific operation systems and versions but environments. If an operating system in a certain version provides such an environment, we support it implicitly. The "supported" operating systems can be interpreted as examples that provide such an environment.

jnweiger:
Ubuntu 20.04 is already a supported platform by client 2.6.3, both branded and unbranded.

The client team policy is defined by a total of 3 rules
* We have "Latest two, plus latest LTS, if not amongst these two" as a general rule. Unless overruled by one of the following:
* Removal of a Linux platform requires a deprecation warning with the previous release. E.g. "this is the last release to support XYZ." If the announcement was not made, we keep a platform one more round.
* If technically feasable. We may not add new expected platforms, or may suddenly remove a platform, if we cannot yet build, or can no longer build for that platform.

michaelstingl:
Yeah, impossible to maintain the desktop & mobile system requirements in the server docs. Please check https://github.com/owncloud/docs/pull/2821 instead.

PR #2688, opened 2020-06-22
Replaced Master/Slave  with primary/replica

JKawohl:
I removed inappropriate terminology to get rid of abusive language in the docs.


phil-davis:
You need to start from latest docs master. Git is reporting conflicts.

PR #2689, opened 2020-06-22
Replaced Master-Slave with primary-secondary

JKawohl:
Replaced violent language with an appropriate terminology.


phil-davis:
Backported all the way - see links above.

PR #2695, opened 2020-06-22
[10.0] Replaced Master-Slave with primary-replica

phil-davis:
Backport #2689 

This backports easily enough all the way back to 10.0 docs, so IMO we may as well do it.


phil-davis:
10.0 branch has `.drone.yml` but these days drone expects to see `.drone.star`

I will try adjusting that in drone for a few minutes.

phil-davis:
Passes. @JKawohl please review/approve again. When I force-pushed it automagically dismissed your review.

PR #2699, opened 2020-06-23
update the latest server version

voroyam:
because now the latest one is 10.4.1

### DIFF
--- a/modules/admin_manual/pages/_partials/maintenance/major_release_note.adoc
+++ b/modules/admin_manual/pages/_partials/maintenance/major_release_note.adoc
@@ -11,8 +11,12 @@ Here are some examples:
 [cols=">10%,^25%,65%",options="header",stripes=even]
 |===
 |Version
-|Can Upgrade to 10.3.2?
+|Can Upgrade to 10.4.1?
 |Requirements
+
+|10.3.1
+|Yes
+|
 
 |10.2.1
 |Yes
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/_partials/maintenance/major_release_note.adoc
```diff
@@ -11,9 +11,13 @@ Here are some examples:
 [cols=">10%,^25%,65%",options="header",stripes=even]
 |===
 |Version
-|Can Upgrade to 10.3.2?
+|Can Upgrade to 10.4.1?
 |Requirements
 
+|10.3.1
```
phil-davis, line 17:
```suggestion
|10.3.2
```

voroyam: The structure is :

image

I don't know why, I did not make it, but it's always x.0.1.

If you want to change this structure - PR welcome :)

This PR updates the current latest version, which is 10.4.1 right now.

When 10.5.1 is out - this will be the latest version.

phil-davis: There is `10.0.10` below, so someone in the past has used the latest patch version of each minor version as examples.
But I will make another PR to adjust that.

phil-davis:
This can be done here, then backported to 10.5 and 10.4 branches. Then do another change for 10.5, which goes in master and 10.5 branches.

phil-davis:
Backports are in #2707 and #2708

PR #2706, opened 2020-06-24
Update system_requirements.adoc with currently supported platforms by desktop client

HanaGemela:


phil-davis:
> I'd also argue that the exact list should not be in a document that is versioned by the server version. These two release cycles are not in sync in any way.

That is a difficulty. But it will be a big effort to have a separate set of client documentation that is versioned by client version.

The other question we always have is how far back to backport this updated information. I expect that the latest client will work against 10.5.0 10.4.* 10.3.* and how far back (maybe all the way back to 10.0.*)?

phil-davis:
I backported to 10.4 branch (current server release) and 10.5 branch (upcoming server release).

@HanaGemela Do you want to update in 10.3 or older server documentation?

HanaGemela:
Thanks @phil-davis 
No need for backporting to older versions

Issue #2723, opened 2020-07-15
Email cannnot be provided directly when creating a user


phil-davis:
## What Needs to be Documented?
The "Add User"  section of the Provisioning API docs says that you can provide the email attribute in the POST.
This is not so. See discussion in https://github.com/owncloud/core/issues/37323

Actually the user has to be created with some password (might be a dummy random string), and then "Edit User" done to set the email address. e.g. https://github.com/owncloud/core/blob/master/tests/TestHelpers/UserHelper.php `createUser()` does this sequence during testing.

(and then maybe a password change link etc is generated and emailed to the user... so that the user can set their real password).

Fix the docs.

Also check the docs about setting display name - that needs a similar sequence. But maybe the doc about that is already OK.

## Where Does This Need To Be Documented?
https://github.com/owncloud/docs/blob/master/modules/admin_manual/pages/configuration/user/user_provisioning_api.adoc

## Why Should This Change Be Made? (Optional)
To be correct :)

## What Type Of Content Change Is This? (Optional)
- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## Which Manual Does This Relate To? (Optional)
- [ ] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

phil-davis:
Fixed in #2725

PR #2725, opened 2020-07-15
Fixup doc of Provisioning API Add User

phil-davis:
Fixes issue #2723 

Backport to maintained branches 10.5 and 10.4

Somehow this was "just wrong". I tested with various `curl` commands there is no way to add an email in the "Add User" POST request.

### DIFF
--- a/modules/admin_manual/pages/configuration/user/user_provisioning_api.adoc
+++ b/modules/admin_manual/pages/configuration/user/user_provisioning_api.adoc
@@ -42,21 +42,14 @@ sending a basic HTTP authentication header.
 | Type 
 | Description
 
-| `email` 
-| string 
-a| The email address for the user (_optional_).
-The user will be emailed a link to set their password, _if_ email is configured correctly.
-
-| `groups` 
+| `groups`
 | array 
 | Groups to add the user to (_optional_).
-Any group that does not already exist, will be created.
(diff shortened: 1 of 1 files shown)


mmattel:
Q: you have removed the complete curl request. Does this mean that the curl command overall does not work? In case the curl command would work but without setting the eMail address, it would be good to keep the command but remove the eMail part from curl only.

phil-davis:
> Q: you have removed the complete curl request. Does this mean that the curl command overall does not work? In case the curl command would work but without setting the eMail address, it would be good to keep the command but remove the eMail part from curl only.

I initially removed just the `email` line. Then when I looked at the remaining example, it was identical to the example above it.

mmattel:
I looked again in files changed, for me it looks that the complete curl example is deleted ...

phil-davis:
> I looked again in files changed, for me it looks that the complete curl example is deleted ...

Yes, "the existing example above the deleted example" is exactly the same as "the deleted example without the 'email' line"

So if I had deleted just the 'email' line, then there would have been 2 identical examples.

mmattel:
hahaha, you are right.
the example without email above was not shown in changed files - of course not, as it was not touched 😅 
everything ok - approving

phil-davis:
Backports in #2726 and #2727

PR #2754, opened 2020-08-19
fix more anotra compile warnings

mmattel:
Some more fixes for antora compile warnings.
The warning list is a lot less now :smile:

Backport to 10.4 and 10.5

The older branches 10.3 - 10.1 may get a backport too if someone takes the task.
But this is a lof of work to see if/which changes have been made from 10.1 to 10.4


phil-davis (approved):
LGTM

The `${ENVVAR}` form is not essential as long the the environment variable name does not become concatenated with some other text in the quoted string - that would make it ambiguous. But that is not a problem in these examples. If Antora or whatever does not like it, then removing it is fine.

phil-davis:
Thanks for the great work!

PR #2758, opened 2020-08-24
Document skipOnFedOcV10 test tags

phil-davis:
PR https://github.com/owncloud/core/pull/37837 added the ability to skip acceptance test scenarios when CI is running with a federated server with an old ownCloud version. This allows developers to write new acceptance test scenarios that will pass with a "modern" combination of local and federated server, but are not expected to pass with some old combination.

Document the tags that can be used.

### DIFF
--- a/modules/developer_manual/pages/testing/acceptance-tests.adoc
+++ b/modules/developer_manual/pages/testing/acceptance-tests.adoc
@@ -662,9 +662,19 @@ Use tag formats like the following to skip on a particular major, minor or patch
 [source,gherkin]
 ----
 @skipOnOcV10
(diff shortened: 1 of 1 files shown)


mmattel (approved):

mmattel:
Q: where does someone know which `@skipOnFedOcVxxx` are present to exclude?
Q: Backport to which versions?

phil-davis:
No backport needed - if a developer checks out 10.5.0 or earlier then the acceptance test scripts will not have this ability.

A developer has to choose the old versions of ownCloud which the new feature or fix will not work with - that is the developer's decision based on what they changed and how it effects federated sharing.

PR #2760, opened 2020-08-26
Upload docs from cron jobs.

dschmidt:
This is a regression fix. Drone used to have 'push' as event type for cron jobs and the cron type did not exist.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -107,6 +107,7 @@ def build(ctx, latest_version, deployment_branch, base_branch):
                 "when": {
                     "event": [
                         "push",
+                        "cron",
                     ],
                 },
             },
@@ -126,6 +127,7 @@ def build(ctx, latest_version, deployment_branch, base_branch):
                 "when": {
                     "event": [
                         "push",
+                        "cron",
                     ],
                 },
             },
@@ -147,6 +149,7 @@ def build(ctx, latest_version, deployment_branch, base_branch):
                 "when": {
                     "event": [
                         "push",
+                        "cron",
                     ],
                     "branch": [
                         deployment_branch,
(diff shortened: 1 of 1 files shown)


micbar (approved):

phil-davis (approved):

mmattel:
Just for my info, what is the impact of this change?

phil-davis:
The daily "scheduled task" in https://drone.owncloud.com/owncloud/docs should now do the various pipeline steps to upload the docs. That will make sure that the published docs get synced at least daily.

I think there is a problem that when you do a backport, updating e.g. 10.5 or 10.4 docs branch, the merge of that does not actually trigger an update push to the published docs. It needs a merge of master to happen AFTER the backports are merged. But if there is not a "continuous stream" of changes being merge to master, then the published  10.5 or 10.4 docs get stale.

Now at least the docs will be pushed/published daily.

dschmidt:
> Just for my info, what is the impact of this change?

Basically: there is a daily cronjob which I introduced to collect changes from external repos, e.g. desktop client or ios-app. Otherwise it might happen that changes don't end up in the docs until something else is pushed to this repo's master. 

Basically what @phil-davis said. 
I thought branches should trigger master builds, at least what I intended to do. If it doesn't work, we need to check.

phil-davis:
I might be wrong, but I think that the merges to 10.4 and 10.5 do not fully trigger upload-push-publish of the changes. There was discussion about that some time ago.

In  any case, the daily (nightly) job should now gather up all the docs components and make sure the latest stuff is published.

mmattel:
Thanks for the explanation.
Everything is green, merging ?

PR #2762, opened 2020-08-28
Update config_sample_php_parameters.adoc

pako81:
Out of the box, every Redis instance supports 16 databases so `<dbIndex>` in config.php has to be set between 0 and 15. This Pull Request adds this information.


voroyam (approved):

mmattel (changes requested):
You **CAN NOT** do any changes in this docs file.
You **MUST** do the change in the master file which in config.sample.php in core !!
When that change is merged, follow the readme in repro `config-to-docs` to file a PR in docs.
If you merge here, any changes made locally will be overwritten on the next config-to-docs run !

pako81:
@mmattel got it, thanks.

PR #2779, opened 2020-09-10
Update ubuntu_18_04.adoc

voroyam:
add occ background:cron to set the background job to cron


modules/admin_manual/pages/installation/ubuntu_18_04.adoc
```diff
@@ -157,6 +157,12 @@ occ config:system:set trusted_domains 1 --value="$myip"
 
 === Set Up a Cron Job
 
+Set your backgroud job mode to cron
+
+----
+occ background:cron
```
mmattel, line 163: NO-NO-NO !!
Use:
```
[source,console,subs="attributes+"]
----
{occ-command-example-prefix} background:cron
----
```

mmattel:
Please add to which branches this needs backporting (add to the main PR description block)

modules/admin_manual/pages/installation/ubuntu_18_04.adoc
cdamken: Why dow we need the `{occ-command-example-prefix}` 

Above in line 30, we create an occ that helps with all commands

cdamken (approved):
It looks good for me

mmattel:
Guys, this is totally nonsense. Those who give a thumbsup should be aware that there is a reason for having the occ commands written as commented above. And not fixing it before merging is fully ignorant and short minded :man_facepalming:

@phil-davis @pmaier1

voroyam:
feel free to add a PR that implements your suggestion.

Please note that you are being very negative. There are people who want to make the docs better, and if you always criticise  the attempts to better something with you suggestions, it does not make it easier.

Fact is - your suggestions are just that - suggestions. If support people (me) 
- write the docs and 
- use the docs for testing and 
- give the docs to users and customers, 
- then fix the docs because we find something missing 

- why should you dictate how we write the docs?

mmattel:
- Highlighting via a comment a writing issue in a PR which is against a good and well made docs practice with a good reason for. See all other occ commands made that way
- It can be expected that a knowledged writer or writer with long docs history knows about that fact.
- Not fixing the point after ~45 days by the writer of the PR
- Complaining about being negative after it has been merged without fixing it - against well known practice
- How would you call that?

??
I complain about a behaviour not a person. If you feel yourself attacked...

modules/admin_manual/pages/installation/ubuntu_18_04.adoc
phil-davis: PR #2835 puts in the `[source,console,subs="attributes+"]` format stuff.

The actual `occ` command is OK, because this document has special instructions "Create the occ Helper Script" which tell you to create an `occ` script that "just works". Refactoring that would be a separate discussion...

phil-davis:
Backports:

10.5 #2839 
10.4

voroyam:
I don't quite understand what you want, you write very complicated @mmattel 

If you want a change - PR welcome. Making a comment takes a minute but making a PR takes longer.

I will make changes to the docs that are good for me and my team, not to meet your expectations or some "a good and well made docs practice".

PR #2781, opened 2020-09-14
[10.5] Review supported list of enterprise apps

jnweiger:
Remove enterprise_key app from the list. It no longer exists with 10.5.
Please review the rest....

Suggest to say that "standard" enterprise apps are now included with the normal server.
Suggest to say that docker image owncloud/enterprise:10.5 does not exist. it is obsoleted by owncloud/server:10.5


mmattel:
@jnweiger the target branch is "incorrect", it should be "master", then we (document in this PR) backport to 10.5 only...

pmaier1 (approved):
Please also forward port to master.

phil-davis:
Forward port to master is PR #2811

PR #2789, opened 2020-09-17
Delete ldap_proxy_cache_server_setup.adoc

voroyam:
if this stays up, customers could get the idea that it works in general, and try to apply it to their custom configuration.
this could lead to more issues.


cdamken:
Well, I think it will not often be used and that part can be found on the internet, is not relevant for ownCloud installation.

Agree, remove it

cdamken (approved):
We don't need that anymore

phil-davis:
Backports linked above in #2832 #2833 #2834

Issue #2759, opened 2020-08-25
Document expiration of outgoing remote shares


jvillafanez:
## What Needs to be Documented?

https://github.com/owncloud/core/pull/37548

This is basically the support of expiration for outgoing (not incoming) remote shares. The behaviour is the same as with local shares.
Note that the expiration happens in the origin; the destination doesn't have any control about the expiration

## What Type Of Content Change Is This? (Optional)
- [x] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## Which Manual Does This Relate To? (Optional)
- [ ] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

michaelstingl:
@jvillafanez @micbar please add some bullet points to the top post what needs to be documented. I’m mostly interested how clients can use it. Is there a new capability for example?

jvillafanez:
I don't think there is too much to document because the behaviour is the same as with user and group sharing. I haven't checked in depth the docs, but I expect something like "User, group and remote sharing support automatic share expiration....." or something like that. I don't think we need a specific section for remote sharing.

The capabilities are mostly the same as user and group sharing:
```
{
    "ocs": {
        "meta": {
            "status": "ok",
            "statuscode": 100,
            "message": "OK",
            "totalitems": "",
            "itemsperpage": ""
        },
        "data": {
            "capabilities": {
                "files_sharing": {
                    "api_enabled": true,
                    "public": {
                        "enabled": true,
                        "password": {
                            "enforced_for": {
                                "read_only": false,
                                "read_write": false,
                                "upload_only": false,
                                "read_write_delete": false
                            },
                            "enforced": false
                        },
                        "roles_api": true,
                        "expire_date": {
                            "enabled": false
                        },
                        "send_mail": false,
                        "social_share": true,
                        "upload": true,
                        "multiple": true,
                        "supports_upload_only": true,
                        "defaultPublicLinkShareName": "Public link"
                    },
                    "user": {
                        "send_mail": false,
                        "expire_date": {
                            "enabled": false
                        }
                    },
                    "group": {
                        "expire_date": {
                            "enabled": false
                        }
                    },
                    "remote": {
                        "expire_date": {
                            "enabled": true,
                            "days": "15",
                            "enforced": true
                        }
                    },
[...]

PR #2795, opened 2020-09-21
Update admin file_sharing_configuration with current information

phil-davis:
The section on all the file sharing settings is missing various parts. I read through the document while looking at `index.php/settings/admin?sectionid=sharing` and made the documentation match what is currently in core 10.5 release.

I noticed this when looking at issue #2759 - I can't really do the doc for that until the existing doc is up-to-date.

Needs backport to 10.5

I will think about backport to 10.4 - need to check which things were or were not in 10.4.

### DIFF
--- a/modules/admin_manual/pages/configuration/files/file_sharing_configuration.adoc
+++ b/modules/admin_manual/pages/configuration/files/file_sharing_configuration.adoc
@@ -29,16 +29,22 @@ You have control of a number of user permissions on file shares:
 ** Allow users to send mail notification for shared files
 ** Set the language used for public mail notification for shared files
 ** Allow users to share file via social media
+* Set default expiration date for user shares
+** Set the number of days to expire after
+** Enforce as maximum expiration date
+* Set default expiration date for group shares
+** Set the number of days to expire after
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/configuration/files/file_sharing_configuration.adoc
```diff
@@ -29,16 +29,22 @@ You have control of a number of user permissions on file shares:
...
+** Set the number of days to expire after
+** Enforce as maximum expiration date
+* Set default expiration date for group shares
+** Set the number of days to expire after
+** Enforce as maximum expiration date
 * Automatically accept new incoming local user shares
 * Allow resharing
-* Default user and group share permissions
+* Allow sharing with groups
 ** Restrict users to only share with users in their groups
```
pmaier1, line 41: Not sure why the two restrictions are on level 2 (`**`). Actually they are independent options on the first level.

phil-davis: You are correct - fixed.

pmaier1 (approved):
Nice, thanks :ok_hand:

PR #2796, opened 2020-09-21
Expire date does not just apply to public link shares

phil-davis:
The description of `expireDate` is too restrictive. In 10.5 it applies to public link, user and group shares.

Backport to 10.5

I will also look at backporting to 10.4 - need to check when `expireDate` existed for user and group shares.


pmaier1 (approved):
THX. I can't judge whether this actually is the same for all share types but makes sense because the feature was just recently added for user/group shares.

PR #2797, opened 2020-09-21
Update description of sharing expiration date

phil-davis:
The section "Changing The Share Expiration Date" is out-of-date. In 10.5 the user can specify an expiration date for user, group and public link shares.

Backport to 10.5

I will look at backport to 10.4 also - IMO that had expiration for those 3 share types.


pmaier1 (approved):
Thanks :+1:

PR #2826, opened 2020-10-26
Update index.adoc

voroyam:
Add links to this empty page to give it some meaning.

I added links to our guides on how to install ownCloud with some description. This way a user can easily access the main parts of the installation section.

### DIFF
--- a/modules/admin_manual/pages/installation/index.adoc
+++ b/modules/admin_manual/pages/installation/index.adoc
@@ -1,5 +1,20 @@
 = Installation
 :page-aliases: go/admin-install.adoc
 
-In this section, you will find all the information you need for installing ownCloud. 
+You can install ownCloud in multiple ways, here are our trusted guides:
 
+== xref:installation/manual_installation.adoc[Manual Installation]
+
+This is a thorough guide to installing ownCloud, containing all the information needed for the prerequisites, the dependencies, the actual installation and the configuration afterwards.
+
+== xref:installation/docker/index.adoc[Installing With Docker]
+
+This guide will show you how to install ownCloud with Docker Compose using our YAML file.
+
+== xref:installation/ubuntu_18_04.adoc[Example Installation on Ubuntu]
+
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/installation/index.adoc
```diff
@@ -1,5 +1,20 @@
...
 
+== xref:installation/manual_installation.adoc[Manual Installation]
+
+This is a thorough guide to installing ownCloud, containing all the information needed for the prerequisites, the dependencies, the actual installation and the configuration afterwards.
+
+== xref:installation/docker/index.adoc[Installing With Docker]
+
+This guide will show you how to install ownCloud with Docker Compose using our YAML file.
+
+== xref:installation/ubuntu_18_04.adoc[Example Installation on Ubuntu]
```
cdamken, line 14: Could we change 18.04 to 20.04? I think the commands are the same with exception of the php-smbclient.

cdamken (approved):
I find this a great idea. works for me ;)

modules/admin_manual/pages/installation/index.adoc
voroyam: yes, but then we also need to add a note about the Ondrej repo.

modules/admin_manual/pages/installation/index.adoc
```diff
@@ -1,5 +1,20 @@
 = Installation
 :page-aliases: go/admin-install.adoc
 
-In this section, you will find all the information you need for installing ownCloud. 
+You can install ownCloud in multiple ways, here are our trusted guides:
 
+== xref:installation/manual_installation.adoc[Manual Installation]
```
mmattel, line 6: You can not have a reference in parallel being a section. You must have a fixed text for the section and in a new independent line, you can do the reference with xref.

voroyam: You can not or you should not? In my local doc installation it works.

mmattel:
Fixed paths and names.
Tested with `yarn run antora`

mmattel:
Note: backport can only be done for 10.6 and 10.5

PR #2835, opened 2020-10-28
Adjust command format in ubuntu_18_04.adoc

phil-davis:
1) Add the settings:
`[source,console,subs="attributes+"]`
to the top of the command section.

This was accidentally left out of PR #2789 

Note: the actual `occ background:cron` should be fine, because the document has instructions up the top about how to get yourself an `occ` command that "just works" - see section "Create the occ Helper Script"

2) Fix an include `examplesdir` reference.

### DIFF
--- a/modules/admin_manual/pages/installation/ubuntu_18_04.adoc
+++ b/modules/admin_manual/pages/installation/ubuntu_18_04.adoc
@@ -159,6 +159,7 @@ occ config:system:set trusted_domains 1 --value="$myip"
 
 Set your backgroud job mode to cron
 
+[source,console,subs="attributes+"]
 ----
 occ background:cron
 ----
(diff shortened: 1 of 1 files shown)


mmattel:
@phil-davis you need to change all parts. Only adding the source attribute without adding the macro is half the way... The correct statement is:
```
[source,console,subs="attributes+"]
----
{occ-command-example-prefix} background:cron
----
```

phil-davis:
@mmattel there are already a few `occ` commands in this file. And it has a section "Create the occ Helper Script" that instructs the reader to make a script that is `/usr/local/bin/occ` and will do the `sudo -u ...` stuff.

Because of that, IMO it will be confusing to readers if there is suddenly an `occ` command that reads like `sudo -u www-data php occ background:cron`

mmattel (approved):

PR #2854, opened 2020-11-03
Update index.adoc

jnweiger:
Possible fix for https://github.com/owncloud/core/issues/38074

OWNCLOUD_DOMAIN can have a port number, and should have one, if non-standard.
OWNCLOUD_DOMAIN must not have a protocol, though. 

Not sure if OWNCLOUD_PORT should be implicitly applied to OWNCLOUD_DOMAIN  too, aparently it is not.
That would decide, if this an issue with docu or with code.

### DIFF
--- a/modules/admin_manual/pages/installation/docker/index.adoc
+++ b/modules/admin_manual/pages/installation/docker/index.adoc
@@ -3,19 +3,23 @@
 
 ownCloud can be installed using Docker, using
 https://hub.docker.com/r/owncloud/server/tags[the official ownCloud Docker image].
-This official image works standalone (e.g. for a quick evaluation with `docker run -p{std-port-http}:{std-port-http} owncloud/server`)
-but it is designed to work with a data volume in
-the host filesystem and with separate _MariaDB_ and _Redis_ containers.
+This official image works standalone for a quick evaluation but is designed to be used in a docker-compose setup.
 
+== Quick Evaluation
+
+`docker run -e OWNCLOUD_DOMAIN=localhost:{std-port-http} -p{std-port-http}:{std-port-http} owncloud/server`
+
+== Docker Compose
 
 The configuration:
 
 * exposes ports {std-port-http}, allowing for HTTP connections.
+* uses separate _MariaDB_ and _Redis_ containers.
 * mounts the data and MySQL data directories on the host for persistent storage.
 
-== Installation on a Local Machine
+The following instructions assume you install locally. For remote access, the value of OWNCLOUD_DOMAIN must be adapted.
 
-To use it, first create a new project directory. Then copy and paste the sample
+First create a new project directory. Then copy and paste the sample
 `docker-compose.yml` from this page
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/installation/docker/index.adoc
```diff
@@ -3,19 +3,23 @@
 
 ownCloud can be installed using Docker, using
 https://hub.docker.com/r/owncloud/server/tags[the official ownCloud Docker image].
-This official image works standalone (e.g. for a quick evaluation with `docker run -p{std-port-http}:{std-port-http} owncloud/server`)
-but it is designed to work with a data volume in
-the host filesystem and with separate _MariaDB_ and _Redis_ containers.
+This official image works standalone for a quick evaluation but is designed to be used in a docker-compose setup.
```
DeepDiver1975, line 6: no idea why you are changing this .... let people run a stand alone container. What is the problem?

DeepDiver1975 (changes requested):
From my understanding the usage of the port number is already documented.
Adjustments to the docker compose file look valid.

modules/admin_manual/pages/installation/docker/index.adoc
jnweiger: Problem was without `-e DOCKER_DOMAIN=loclahost:8080` it would not fully work. -- I wanted to add that to make it easier to run the standlone container.

DeepDiver1975: Well - then let's just add this argument and we are fine. No need to change the meaning of the rest of the docs.

DeepDiver1975 (changes requested):

modules/admin_manual/pages/installation/docker/index.adoc
jnweiger: The meaning of the rest is unchanged. I removed duplicated stuff and disentangled the language. I fail to see where this is bad. I can split it in two PRs if needed.

DeepDiver1975: plz split it up. THX

mmattel:
Please run things always against master and then do backports. Not the other way round...
Note in the PR descriptipon to which branches you want to backport to.
Exept fixes to a very particular branch, you must not do differnt to avoid possible divergences.
The mess ceaning up is awful...

DeepDiver1975:
.... and a clean commit history with proper commit message .... well I stopped telling people how to work properly ..... :man_shrugging:

jnweiger:
split into #2858 and #2859 
added backport hints there. (only the obvious, I am no expert)
closing here unmerged. Hope stuff is cleaner then.

Sorry for being so sloppy with the docs, I am too much priority driven these days.

modules/admin_manual/pages/installation/docker/index.adoc
jnweiger: Done. split into #2858 and #2859

PR #2856, opened 2020-11-06
Add note about username and password

Gregliest:
Add a note to the docs that the admin username and password do not change between deploys, but only change if you delete the docker volumes. See https://github.com/owncloud-docker/server/issues/76 for more info.


modules/admin_manual/pages/installation/docker/index.adoc
```diff
@@ -47,6 +47,10 @@ configuration settings. Only a few settings are required, these are:
 | `{std-port-http}`
 |===
 
+Note: `ADMIN_USERNAME` and `ADMIN_PASSWORD` will not change between deploys even if you change the
```
mmattel, line 50:
```suggestion
NOTE: `ADMIN_USERNAME` and `ADMIN_PASSWORD` will not change between deploys even if you change the 
```
Needs correct writing because `NOTE:` is a antora macro creating a proper note.

mmattel:
@DeepDiver1975 mind to comment if the text statement made is correct.
@Gregliest this needs backport to which doc branches (10.5, 10.4 ect)?

Gregliest:
My impression is that this was a design decision, so I would think it applies broadly. However, I'm not familiar with the code base, and when/if this was changed.

mmattel (approved):

mmattel:
backports needed to 10.6, 10.5

PR #2859, opened 2020-11-10
Possible fix for owncloud/core#38074

jnweiger:
OWNCLOUD_DOMAIN can have a port number, and should have one, if non-standard.
OWNCLOUD_DOMAIN must not have a protocol, though.

Not sure if OWNCLOUD_PORT should be implicitly applied to OWNCLOUD_DOMAIN too, aparently it is not.
That would decide, if this an issue with docu or with code.

Obsoletes another part of the unmergeable #2854

Backports needed: 10.5, 10.6 ... maybe more?


DeepDiver1975 (approved):
Not personally tested - trusting the submitter that this will work properly .....

phil-davis:
Backport to 10.5 #2863

phil-davis (approved):

PR #2860, opened 2020-11-13
Add missing part from service config

enbrnz:
The configuration example was previously not copy-pastable because the following section was missing:
```
[Install]
WantedBy=multi-user.target
```
This results in the following error when trying `systemctl enable ...`: `The unit files have no installlation config`


voroyam (approved):
Great catch. please add where this change has to be back ported to. I am guessing 10.5 and 10.4.

phil-davis:
Backports:
10.5 #2861 
10.4 #2862

PR #2864, opened 2020-11-18
removed ldap cache proxy from index

voroyam:
because I forgot to do it when removing the document itself

backports to 10.5, 10.4


phil-davis (approved):

phil-davis:
Backports:
10.5 #2865 
10.4 #2866

PR #2878, opened 2020-11-30
Some text fixes for the external mount point documentation

mmattel:
Some text fixes and added the missing read only option.

Backporting to 10.5 necessary


phil-davis (approved):
LGTM - Read Only got introduced somewhere in a "recent" 10.*

phil-davis:
According to the core changelog the Read Only mount was added in 10.4. So we can backport to 10.5 and 10.4

mmattel:
I can not merge, can you do that form me?
Will do the BP asap post merging

Issue #2850, opened 2020-11-02
Delete the port from the Collabora Online server field 


Akidxx:
After installing and configuring Collabora online to owncloud using appliance:
The port number in the 'Collabora Online server' field should be deleted
Settings > Admin > Additional > Collabora

PR #2883, opened 2020-11-30
Update office.adoc

mschreml:
Added Note to Collabora installation manual [https://github.com/owncloud/docs/issues/2850]


voroyam (changes requested):
Can you test it and then try to formulate the sentence again? I think you don't understand the issue because you did not encounter it :)

The issue is - that as of now - if you install ownCloud + collabora in the appliance, it will fill the server address filed with the default settings of localhost:someport

Now, if you then see it, and decide to change out the localhost part with the real address of the server, you still are met with the challenge of - do I need the port or not?

I want the new "Note" telling the user that if he had the default address in the server field that he should use the real address of his collabora server without the port.

PR #2916, opened 2020-12-08
removed ldap proxy docs

voroyam:


Akidxx (approved):

mschreml (approved):

phil-davis:
why are we doing anything to the docs for 10.0 ?
These are very old. We would have to get the CI and publishing... working with the current drone+docker in order to get the changes to be published.
If we do this, then we need to do it for 10.1, 10.2, 10.3 also.

voroyam:
> why are we doing anything to the docs for 10.0 ?
> These are very old. We would have to get the CI and publishing... working with the current drone+docker in order to get the changes to be published.
> If we do this, then we need to do it for 10.1, 10.2, 10.3 also.

Well, dear Phil, it’s because our great search feature and the wise decision to have 10.0 til 10.5 branches in the online docs.
With both - it makes it difficult to use the search. You get old results like this, and I was trying to remove them.

I totally agree with you that working on 10.0 docs should not be a thing, but as long as they are online - I have to work on them.

phil-davis:
Then "someone" needs to decide what to do about this more generally. For example, there could be a decision to:

1) Keep online "live" docs for all 10.* versions, keep the CI configs up-to-date in all those branches, and therefore be able to make "live" changes to any/all branches when we find "not good advice"/"important mistakes"/... in the old-version docs.
or
2) Generate PDF (or some fixed-format) document for "old" doc version 10.0 10.1... and remove those from the "live" docs. This would mean that when we find "not good advice"/"important mistakes"/... in the PDF, it will/might be a "fait accompli" and more difficult to fix.
or
3) Completely delete "old" docs. Then there is no chance for them to have errors, but users on the old versions will not have easy access any more (they would have to look around in git history...)
or
4) ???

Also, either of (2) or (3) means that people who have saved links direct into `10.0` ... docs will get "404 not found". So someone would need to decide what is "a good thing to do" - for example, provide a link to the current docs, try to auto-redirect into the requested place in the current docs...

voroyam:
Yeah, we had this discussion more than once already. I have been told that there is a process now, let's see.

Issue #2870, opened 2020-11-26
[QA] add 10.6 branch


jnweiger:
Server 10.6.0beta1 and RC1 are built using https://doc.owncloud.com/server/10.5/user_manual/ownCloud_User_Manual.pdf
For the final we need 10.6 branch there.

phil-davis:
And for the next week (or so...) it will be easy to wait before branching. Then we will automagically capture whatever good doc changes we make in master, without having to bother doing backports of every PR.

jnweiger:
TODO;
* [ ] review open PRs, merge e.g. https://github.com/owncloud/docs/pull/2858
* [ ] create a 10.6 branch
* [ ] prepare update of https://gitea.owncloud.services/devops/websites/src/branch/master/ansible/roles/docs/templates/config.j2
* ... anything else?

mmattel:
closing, branch has been created

PR #2919, opened 2020-12-10
branch 10.6 with updated .drone.star and site.yml

jnweiger:
Includes todays latest commits from @voroyam and @mschreml, thanks!

Needed for 10.6.0RC2

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -4,7 +4,7 @@ def main(ctx):
     # Version shown as latest in generated documentations
     # It's fine that this is out of date in version branches, usually just needs
     # adjustment in master/deployment_branch when a new version is added to site.yml
-    latest_version = "10.5"
+    latest_version = "10.6"
 
     # Current version branch (used to determine when changes are supposed to be pushed)
     # pushes to base_branch will trigger a build in deployment_branch but pushing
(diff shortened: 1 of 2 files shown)


micbar (approved):
LGTM

dschmidt (approved):

phil-davis (approved):
Looks OK.

There are plenty of docs issues to finish for 10.6. So if we actually create the branch after merging this PR, then we have to backport every PR to `master`, to `10.6` - but I guess that is life. We have to make the branch some time.

dschmidt:
Well, it can just be merged to master at some point - just the version number changes from this pr need to be reverted on master. Afterwards further merges should be straight forward and don't need to be done for every single commit.

jnweiger:
@phil-davis I've waited 14 days since https://github.com/owncloud/docs/issues/2870#issuecomment-734240745
Sorry to rush that now. I need the PDF be inplace for the preparing the final next Monday.

phil-davis:
> I've waited 14 days since...

Yes, that's fine. The docs work is not making progress at the moment, so we have to move forward with the branching and catch-up the content "later".

dschmidt:
The problem with the missing pdf is most probably that master builds all versioned docs, but it does not build non-master pdfs.

version branches should build and deploy their pdfs, pipelines need to be adjusted for that (and probably backported).

dschmidt:
Fixed in https://github.com/owncloud/docs/pull/2923

as I said: the pdf fix needs backporting for fixing older pdfs upload.

PR #2925, opened 2020-12-10
Do not exclude version branches from running pipelines.

dschmidt:
We need to run pipelines to build pdfs. Steps are properly guarded already.


jnweiger (approved):
Thanks!
PDFs are built by the version branches.

PR #2945, opened 2020-12-20
Missing config sample changes from core 10.6

mmattel:
These are config.sample changes in core for 10.6 but not yet have been transported to docs. This is the result of the config-to-docs run based on branch core/release-10.6.0

Needs backporting to 10.6 only !


phil-davis (approved):

phil-davis:
Backport to 10.6 done - PR #2946

PR #2958, opened 2020-12-22
Improve Manual Installation

mmattel:
This is a consolidation of the manual installation documentation, which is a mess as it currently looks like.
The changes made look already quite well and are now ready for review.

A lot of the content has been reorganized and put together on the right places. Other outdated content has been removed and of course a lot has been updated and corrected. In addition, linked content necessary from the configuration section has been fixed too.

These changes should only be backported when merged to 10.6

@phil-davis I definitely would appreciate to get your feedback. But to do so, you must start a build locally (`yarn run antora`) and see the changes in your browser.

@micbar @pmaier1 your feedback is also welcomed :smile:


mmattel:
@phil-davis the PR is now ready to be reviewed :smile:

micbar:
@cdamken @voroyam Please review.

voroyam:
Seems odd to me to have 2 similar titles 
image

----

Server needs to be capitalised if installation is
image

----


where did you put all the packages and their explanation? I remember Tom was looking for it and asking me when I re organised and put them at the end of the document. I don't see any use in them but maybe others do. Maybe this is a point we should discuss with others.

----


`and how to download and install ownCloud.`
Two "and"s are never good in one sentence.

----


`Even we try to make these steps as easy as possible by offering ready to use commands and scripts, you need`
I feel like there is an "if" here that is missing. "Even if we try..." or something.

----

`This document does not give any proposals regarding protecting your server.`
securing might be more fitting here.

----


`To read about the requirements of your server, read the`
Read is twice here. Should be replaced or formulated differently.

----


Ah, now I found the packages, there are in a separate guide.
Which is not listed in the toc for some strange reason
image

----



`The following section takes care about in detail steps.`
This sentence seem just wrong to me. object is missing I think.

----


You link from the main page to different guides, many different guides. I think this leads to the users having a lot of tabs open and switching back and forth. And none of the guides are listed.

----


`Physical Installation of ownCloud`

why do we need physical in here? Feels like we moving boxes here.

voroyam:
I like that you have cleaned up the toc a big, what I don't like is the 1 article in subfolder practice.

If I have to do an extra click I would like to see more articles in the subfolder and not just one. If there is only one - we can remove the subfolder.

image

mmattel:
> Seems odd to me to have 2 similar titles

Renamed --> "Detailed Installation Guide"
plus made a subsection out of it so it fits to "Quick Installation Guides"

image

> Two "and"s are never good in one sentence.

Yes, there are two “and” but it is ok to have it in this case because of listing items.

> I feel like there is an "if" here

Corrected

> securing might be more fitting here

Corrected

> Read is twice here. Should be replaced or formulated differently

Corrected

> This sentence seem just wrong to me. 

Rephrased

> why do we need physical in here

Physical is correct, because you need to install “binaries” in a working way. Post that task, you have the logical installation.

> your seems wrong. the might be better

I do not get the point, looks ok to me…

> target directory seems superfluous

„target“ removed

> if you adding an example - why not just add a working link

Because you enforce admins to install the latest package and we do not change all the links/texts on every release made.
This part is not a copy/paste section as you have it with the quick install guides.

> …potential future source of many support tickets…

First, this script is working now for about two years flawless. It was there before without any complains from support and I improved it a bit. I also use that script successfully for all my instances, that is a good way of proving it.
Second, this script of course differs from the quick install guides in a couple of ways:
- it collects all manually steps you would have to do by typing or copy/paste
- it takes care having multiple instances to maintain (by having several pre-parameterized starting scripts)
- you can install AND upgrade with the same script – it asks you questions about what you want to do. And you will do upgrades post installation when available.
- for upgrades, it takes care that nothing gets overwritten and you can go back (before running ./occ upgrade)
- it sets the correct ownership and permissions automatically

Please remember, that you have to do these steps anyways for installation AND upgrades again and again. The script automates that process.

> You also advice to use the scripts

Yes, even if you want to do all task manually, at one point you have to set permissions and ownership not only for data and binaries, but also for .htaccess files. Some directories have different requirements, the script takes care of that. The script makes all necessary steps and gives you advices when finished for finalization of an install or upgrade.

> If you like you can put all the commands in to a script and execute it

That´s exactly what it does, in a very comfortable way :smile:
Just give it a try and you will see the big benefit.

> cp -r owncloud /var/www

Corrected  and rephrased

> no idea why we need this

This taregts the “SELinux (Security-Enhanced Linux)” part. It was there before and I kept it as it is.

voroyam:
> Yes, there are two “and” but it is ok to have it in this case because of listing items.
> 
> >

Listing grammar rules are "comma" comma" "and" "period". Last time I checked :) It just looks nicer when you have proper grammar, nice build sentences and appropriate capitalization. Instead of this, that and this and that. But it's your choice in the end.


> Physical is correct, because you need to install “binaries” in a working way. Post that task, you have the logical installation.

"You are technically correct, which is the best kind of correct" (Futurama). But I have never heard nor read a physical installation guide/step. The word just seems more fitting with plugging in cables than downloading something.




> I do not get the point, looks ok to me…

"Your ownCloud package". Does everyone get's his own special package with Signature and Surprise? :) If not, and every one get's the same, then "the" is more fitting I think.

> First, this script is working now for about two years flawless.

and you are basing this on...? 

> It was there before without any complains from support and I improved it a bit.

Are we talking about the permissions script or the installation / upgrade script? Because I don't remember us having an installation script in the docs. About the permission script - someone (tm) could fail to set one variable and the whole server would be in trouble. 



> Yes, even if you want to do all task manually, at one point you have to set permissions and ownership not only for data and binaries, but also for .htaccess files. Some directories have different requirements, the script takes care of that. The script makes all necessary steps and gives you advices when finished for finalization of an install or upgrade.

All hail to the script! It takes care of everything :) I am just saying that the danger in writing complex scripts is that not experienced users won't know what happened or how to debug if they use it and make a mistake. Same thing with the Theming script. In theory you just copy paste the script and it works, but it did not. Because it was huge and split in 2 parts. For the author it's easy peasy because he/she has the context and can quickly see what's wrong. Not so for the average user. That's why I am for the foolproof way of splitting up a large script in to simple commands.

mmattel:
> It just looks nicer when you have...

I made a list out of it, looks better.

> Your ownCloud package...

Got it, fixed

>  I don't remember us having...

Yes it was there, but it is a good sign that you did not have any issues with it - else you would have reported it :smile:

>  That's why I am for the foolproof...

I understand your point. Both ways have benefits and their drawbacks.
Personally I guess that there is no one fit´s all solution. Not for the experienced nor ordinary admins.
Your "Quick Guides" give a nice copy/paste method and you see what you are doing. On the other hand side, the script is a big help too. Someone who does not go into details, would not read the detailed install part, as he wants to quickly do things. I guess that here is the point where we could agree and hook on. I will add a note about that fact - which would help to direct the audience to the proper documents.

As a first step, I added a "warning" sign on the script page.

voroyam (approved):

mmattel:
@voroyam I have added a description when to use which install documents.
Thanks for approving, will do a squash now.

mmattel:
@voroyam or @phil-davis checks have passed. Could you merge as I do not have rights to do so.
I would post merging create the backport to 10.6 Many thanks.

PR #2964, opened 2020-12-28
[10.5] improve manual install

phil-davis:
Backport of PR #2958 

I was surprised - the cherry-pick to 10.5 worked with no conflicts.

@mmattel are you aware of changes here that are specific to 10.6?

If not, then we may as well do this easy backport and have the 10.5 docs be organised the same as 10.6


mmattel:
There are no specific changes restricting to 10.6 only.
I just wanted to avoid long cherry picking and issue solving because the changes were - lets say a lot...
And 10.6 was pretty fresh at all. So lets go 😄

mmattel (approved):

PR #2983, opened 2020-12-30
Update server_prep_18.04 when using multiple PHP installs

mmattel:
This updates the server 18.04 prep document for the situation, when having multiple php versions installed. This happens when using the ondrej ppa and is essential for the successful installation of smbclient via pecl with the chosen php version.

Backporting to 10.6 and 10.5 necessary.

### DIFF
--- a/modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_18.04.adoc
+++ b/modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_18.04.adoc
@@ -10,19 +10,21 @@
 :mcrypt-pecl-url: https://pecl.php.net/package/mcrypt
 :overriding-vendor-settings-url: https://www.freedesktop.org/software/systemd/man/systemd.unit.html
 :transport-huge-pages-url: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/performance_tuning_guide/s-memory-transhuge
-:iscsi_initiator-url: https://help.ubuntu.com/lts/serverguide/iscsi-initiator.html
+:iscsi_initiator-url: https://ubuntu.com/server/docs/service-iscsi
 :ondrej-php-url: https://launchpad.net/~ondrej/+archive/ubuntu/php
-:libsmbclient-php: https://github.com/eduardok/libsmbclient-php
+:libsmbclient-php_url: https://github.com/eduardok/libsmbclient-php
+:update-alternatives_url: https://manpages.ubuntu.com/manpages/precise/man8/update-alternatives.8.html
+:smbclient_url: https://packages.ubuntu.com/search?keywords=smbclient
+:pecl_url: https://pecl.php.net/package/smbclient
 
 == Introduction
 
-The target audience for this document is more sophisticated admins with additional needs and setup scenarios.
-Ubuntu 18.04, by design, only provides PHP 7.2.
-Because PHP 7.2 security support ends in November 2020, you may want to to consider using PHP 7.3.
-If you want to install the PHP 7.3 package and other required PHP extensions for PHP 7.3 for Ubuntu 18.04,
-you need to use the well known, custom, PPA {ondrej-php-url}[ondrej/php].
-Independent of the PHP version used, all other items, such as installing extensions or comments, and notes,
-remain the same.
+The target audience for this document are more sophisticated admins with additional needs and
+setup scenarios. Ubuntu 18.04, by design, only provides PHP 7.2. Because PHP 7.2 security
+support ends in November 2020, you may want to consider using PHP 7.3. If you want to install
+the PHP 7.3 package and other required PHP extensions for PHP 7.3 for Ubuntu 18.04, you need to
+use the well known, custom, PPA {ondrej-php-url}[ondrej/php]. Independent of the PHP version
+used, all other items, such as installing extensions or comments, and notes, remain the same.
 
 == Apache Web Server and PHP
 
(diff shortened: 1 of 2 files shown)


modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_18.04.adoc
```diff
@@ -33,6 +35,113 @@ The following command installs the Apache web server and PHP.
...
+
+[source,console]
+----
+There are 2 choices for the alternative php (providing /usr/bin/php).
+
+  Selection    Path             Priority   Status
+------------------------------------------------------------
+  0            /usr/bin/php7.4   74        auto mode
+* 1            /usr/bin/php7.3   73        manual mode
+  2            /usr/bin/php7.4   74        manual mode
```
phil-davis, line 69: This example does not have PHP 7.2 in the list at all. I have a feeling that Ubuntu 18.04.5 might nowadays already provide something later than 7.2 - but I would need to install a VM to really find out.

phil-davis (commented):
Just a few comments. I am not sure how much detail to have in these documents.

Also, we may as well show PHP 7.4 as the example. Because 7.3 will already be "on the way out", and really people should go straight to PHP 7.4. That will save them having to upgrade again at the end of 2021.

modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_18.04.adoc
mmattel: You are totally right regaring php. But the thing is a bit more complicated.
I am planning a 20.04 document which will contain an upgrade and new install section.
With 7.4+ there is a special case that needs coverage. I have prepared some stuff already but the basis is this document when merged. There will be a reference for php7.4 to 20.04 to avoid double content and 18.04 is already replaced by 20.4. Means, that 18.04 will get a small addon change but it is easier for me to do this in two steps because I can set the cross references properly.

Therefore, if you are fine with the changes made here, please approve and merge. I will do the backports to 10.6 and 10.5 and then I can start working on 20.04 😄

mmattel:
> Also, we may as well show PHP 7.4 as the example. Because 7.3 will already be "on the way out", and really people should go straight to PHP 7.4. That will save them having to upgrade again at the end of 2021.

You are totally right regaring PHP 7.3. But things are a bit more complicated.
I am planning a 20.04 document which will contain an upgrade and new install section.
With 7.4+ there is a special case that needs coverage. I have prepared some stuff already but the basis is this document when merged. There will be a reference for PHP 7.4 to 20.04 to avoid double content and 18.04 is already replaced by 20.4. Means, that 18.04 will get a small addon change but it is easier for me to do this in two steps because I then can set the cross references properly.

Therefore, if you are fine with the changes made here, please approve and merge. I will do the backports to 10.6 and 10.5 and then I can start working on 20.04 😄

modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_18.04.adoc
```diff
@@ -12,17 +12,19 @@
...
 == Introduction
 
-The target audience for this document is more sophisticated admins with additional needs and setup scenarios.
-Ubuntu 18.04, by design, only provides PHP 7.2.
-Because PHP 7.2 security support ends in November 2020, you may want to to consider using PHP 7.3.
-If you want to install the PHP 7.3 package and other required PHP extensions for PHP 7.3 for Ubuntu 18.04,
-you need to use the well known, custom, PPA {ondrej-php-url}[ondrej/php].
-Independent of the PHP version used, all other items, such as installing extensions or comments, and notes,
-remain the same.
+The target audience for this document is more sophisticated admins with additional needs and
```
mmattel, line 22:
```suggestion
The target audience for this document are more sophisticated admins with additional needs and
```

modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_18.04.adoc
phil-davis: Note: I tried a fresh Ubuntu 18.04.5 in a VM. `sudo apt install php` installs PHP 7.4.3 built May 26 2020. For PHP, Ubuntu has a bit of a moving target. At the time of the original Ubuntu 18.04 release it installed PHP 7.2. So some people with an "old existing" Ubuntu 18.04 might have PHP 7.2 or 7.3, depending when they first installed Ubuntu 18.04. The "long term support" (LTS) for Ubuntu is about the kernel and tightly-linked OS stuff. Versions of various programming languages like PHP still move upwards... (I guess that is because in the modern world there is no way any install should stick with the current version of PHP for 5 years)

So it makes writing these sort of docs more difficult.

But actually, for a fresh Ubuntu 18.04, the above is what happens - "auto mode" shows 7.4 and the instructions here end up retro-fitting 7.3

phil-davis (approved):

phil-davis:
I added a few "the" and other very minor grammar things, then rebased and squashed.
I will merge and make backports when CI passes.

IMO we should next update this document for PHP 7.4. ownCloud10 supports PHP 7.4, it is recommended, and nowadays it is the default PHP that gets installed with a fresh Ubuntu 18.04. So there does not seem an longer much point in having instructions for how to retro-fit PHP 7.3.

phil-davis:
Backports:
10.6 #2997 
10.5 #2998

mmattel:
> ... next update this document for PHP 7.4. ...

see #2986 😃 
I am heavily working on it!

PR #2986, opened 2021-01-03
Server Preparation Guide for Ubuntu 20.04

mmattel:
This is a new guide for Ubuntu 20.04

~~I can finalize when https://github.com/owncloud/docs/pull/2983 is merged and backported because I need to create proper references from 18.04 to 20.04 which I will include in this PR.~~

#2983 has been merged in the meanwhile.

Backport to 10.6 and 10.5 only.


mmattel:
@phil-davis @voroyam
The PR is now ready for review. The best thing to start reviewing would be to do a local build and see the changes 😃 The content and the order of items have been tested, including the OS upgrade and the custom samba build. I had gray hairs before, now I am white headed... All things are tested and working.

For U20.04, the focus is on php7.4 delivered via the OS, not the ondrej package as it complicates things. You only need it if you want to have the absolute latest php patches or php8. Everyone can post installing the standard setup, also install the ondrej ppa but the task post installing (like pecls smbclient) are the same - with the notices provided.

Those who want to install php7.4 on U18.04 are pointed to look on the U20.04 guide to see what they will expect and which tasks have to be done, so no double writing needed.

I personally did not expect that much backend work to document it, but with the guides, the things are now relatively smooth to do. Some text to understand the background and more or less copy/paste the commands.

voroyam:
sorry I don't understand why it has to be so complicated. The package list you write is pretty long and has things that are not required for the ownCloud server to run. Why are they being installed?

Why does someone who just wants ownCloud has to compile their own Samba?

I can't give a review to something I don't understand the purpose of

mmattel:
> Why does someone who just wants ownCloud...

that is exactly the point. Due to a change in the underlaying smbclient, you may not longer be able to connect. Which is especially true when NT1 has been used before successfully. For this case, you need to follow the guide to gain reconnectivity again. In a nutshell, no guide, no access...

voroyam:
> In a nutshell, no guide, no access...

no access to ownCloud server without the smbclient?

mmattel:
> no access to ownCloud server without the smbclient?

Correct. No access from the server where ownCloud runs to the smb fileserver where the data is.

JanAckermann:
TBD, moved to server board

micbar:
@voroyam @mmattel Next steps?

mmattel:
approve and merge...
(just rebased)

mschreml (approved):

PR #2989, opened 2021-01-04
Edits to ubuntu installation

phil-davis:
1) various edits to tidy up a few words in both the 18.04 and 20.4 "quick install" docs.

--------

2) I wondered what the dot was in `chown -R www-data. owncloud` - it is an old form of the `chown` command with dot as separator between user and group name. Nowadays colon is the documented way. So `www-data:` is the same as the old `www-data.` And the means to also use the group of `www-data`. In this case you correctly end up with user `www-data` and group `www-data`
- I will put this in a separte PR

--------

### DIFF
--- a/modules/admin_manual/pages/installation/quick_guides/ubuntu_20_04.adoc
+++ b/modules/admin_manual/pages/installation/quick_guides/ubuntu_20_04.adoc
@@ -10,13 +10,13 @@ Run the following commands in your terminal to complete the installation.
 
 == Prerequisites
 
-* A fresh install of Ubuntu 20.04 with SSH enabled.
+* A fresh installation of https://www.ubuntu.com/download/server[Ubuntu 20.04] with SSH enabled.
 * This guide assumes that you are connected as the root user.
 * This guide assumes your ownCloud directory is located in `/var/www/owncloud/`
 
 == Preparation
 
-First, ensure that all of the installed packages are entirely up to date, and that PHP is available in the APT repository.
+First, ensure that all the installed packages are entirely up to date, and that PHP is available in the APT repository.
 To do so, follow the instructions below:
 
 ----
@@ -38,7 +38,7 @@ sudo -u {webserver-user} /usr/bin/php {install-directory}/occ "\$@"
 EOM
 ----
 
-Make helper script executable:
+Make the helper script executable:
 
 ----
 chmod +x /usr/local/bin/occ
@@ -60,12 +60,13 @@ apt install -y \
   wget
 ----
 
-Note : php 7.4 is default version installable with Ubuntu 20.04
+Note : php 7.4 is the default version installable with Ubuntu 20.04
 
 === Install the Recommended Packages
 
-The package php-smbclient was removed from the offical repository, so you have to add ondrej’s ppa:
-You will need the php-smbclient from ondrej's repository.
+The package php-smbclient was removed from the official repository.
+It is available in the ondrej/php repository (ppa).
+Add ondrej’s ppa with these commands:
 
 ----
 sudo add-apt-repository ppa:ondrej/php
@@ -78,7 +79,8 @@ and add the key:
 ----
 apt-key adv --recv-keys --keyserver hkps://keyserver.ubuntu.com:443 4F4EA0AAE5267A6C
 ----
-Now you can install the recommended packages using the following command
+
+Now you can install the recommended packages using the following command:
 
 ----
 apt install -y \
@@ -160,7 +162,7 @@ occ config:system:set trusted_domains 1 --value="$myip"
 
 === Set Up a Cron Job
 
-Set your backgroud job mode to cron
+Set your background job mode to cron
 
 ----
 occ background:cron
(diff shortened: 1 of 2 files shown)


voroyam:
> 2\. I wondered what the dot was in `chown -R www-data. owncloud` - it is an old form of the `chown` command with dot as separator between user and group name.

`www-data.` is short for `www-data:www-data`

voroyam:
Can you explain the benefit in putting scripts in different folders and linking them? To this specific guide. For me it makes it harder to maintain the page, because I have to look for stuff in different locations.

phil-davis:
> www-data. is short for www-data:www-data

That is true. Also `www-data:` is short for `www-data:www-data` and the `:` is the "modern" way that is documented in `man chown`

I made the change because I wondered about the dot at the end of `www-data.` and went looking for the syntax and had to do a bit of searching to find that the dot is an "old" form, e.g. https://serverfault.com/questions/194295/using-dot-as-delimiter-to-specify-group-in-chown

phil-davis:
> Can you explain the benefit in putting scripts in different folders and linking them? To this specific guide. For me it makes it harder to maintain the page, because I have to look for stuff in different locations.

I have no opinion on that. I am just following the current layout of the files. "somebody" who is in charge of "docs engineering" needs to manage those sort of standards and explain why they are useful or not.

voroyam:
> I have no opinion on that. I am just following the current layout of the files. "somebody" who is in charge of "docs engineering" needs to manage those sort of standards and explain why they are useful or not.

in my opinion actions require reasoning. If you think the way it is is good - you do it. If not - you don't. If you have to follow rules - it's a different game, you follow them even if you don't know the reason behind it. I don't see the rule that says "you have to extract scripts in to separate files and link them" in here.

voroyam:
> > www-data. is short for www-data:www-data
> 
> That is true. Also `www-data:` is short for `www-data:www-data` and the `:` is the "modern" way that is documented in `man chown`
> 
> I made the change because I wondered about the dot at the end of `www-data.` and went looking for the syntax and had to do a bit of searching to find that the dot is an "old" form, e.g. https://serverfault.com/questions/194295/using-dot-as-delimiter-to-specify-group-in-chown

okay, so the "old" way still works but there is a new way. got that.

phil-davis:
IMO having a script in a folder `18.04` that is linked to from a 20.04 installation guide has the risk that the script might be edited by somebody to fix it for 18.04, and they may not realise that it is also used in the 20.04 document. So it would automagically have the changed content in both. Maybe the change will be correct for both 18.04 and 20.04, but maybe it needs to be different - who knows?

I just felt that the reference to `18.04` in a `20.04` did not "look right".

I will revert that change. I did not realise that it would be an issue.

voroyam:
> IMO having a script in a folder `18.04` that is linked to from a 20.04 installation guide has the risk that the script might be edited by somebody to fix it for 18.04, and they may not realise that it is also used in the 20.04 document. So it would automagically have the changed content in both. Maybe the change will be correct for both 18.04 and 20.04, but maybe it needs to be different - who knows?
> 
> I just felt that the reference to `18.04` in a `20.04` did not "look right".
> 
> I will revert that change. I did not realise that it would be an issue.

I don't question your fix, I question the whole concept of the script in folder idea :D

phil-davis:
I will also move the `chown` change into a separate PR, so it can be discussed separately.

phil-davis:
OK - ready for review again.

voroyam:
> OK - ready for review again.

I am checking your statement currently, I am not sure if php comes preinstalled on ubuntu 20.04. will give the review after my test.

phil-davis:
> > OK - ready for review again.
> 
> I am checking your statement currently, I am not sure if php comes preinstalled on ubuntu 20.04. will give the review after my test.

That's fine - good to test. I will just say that I did not make any such statement. The existing document already made that statement. I just adjusted the English grammar a little bit.

voroyam:
image

voroyam:
This is what I am testing. There is a difference between installable and installed.

voroyam:
image

modules/admin_manual/pages/installation/quick_guides/ubuntu_20_04.adoc
```diff
@@ -60,12 +60,13 @@ apt install -y \
   wget
 ----
 
-Note : php 7.4 is default version installable with Ubuntu 20.04
+Note : php 7.4 is the default version installed with Ubuntu 20.04
```
voroyam, line 63: my only objection is that "installed" suggests that php 7.4 comes with Ubuntu 20.04 preinstalled.

voroyam (approved):

modules/admin_manual/pages/installation/quick_guides/ubuntu_20_04.adoc
phil-davis: sorry, you are correct. `php` is not pre-installed:
```
$ php
Command 'ph' not found, but can be installed with:

sudo apt install php7.4-cli
```

I will revert back to "installable"

phil-davis:
Backports:
10.6 #2995 
10.5 #2996

PR #3010, opened 2021-01-11
Update ubuntu_18_04.adoc

voroyam:


mmattel:
@voroyam because the file is not linked anymore, pls check if it is orphand and in case delete it too.
You can check this by running, starting following script from the **docs root folder**, `bin/find-orphaned-files.sh`
If your file is listed, delete it and do another commit/push, thanks.

Issue #3014, opened 2021-01-12
Issue related to dep html-entities v2.x 


xoxys:
Build was broken after merging https://github.com/owncloud/docs/commit/653cf01b12b243c6f25d972e9bec1c716ee26c26 two weeks ago... Someone needs to check how to upgrade. Second issue: The related PR was **not** failing for whatever reason maybe that's a caching thing.

The master branch is fixed again, I've downgraded the dependency to v1.x

@phil-davis Can you pin html-entities to v1.x in dependabot?

phil-davis:
Yes, the PR passed the docs-validate docs-build etc steps. So I assumed that everything was fine, and the new major version was good-to-go (that whatever backward-compatibility breaks it had, they did not effect us)

Actually there needs to be someone who knows and understands the workflow and dependencies... of all this, and who can make informed decisions about this stuff. I really don't know who that is.

PR #3019, opened 2021-01-13
add env option to build search index without upload

xoxys:
This PR adds on env variable option to build the search index without uploading (default=true). This option is required to use as much as possible 3rd party dependencies also on PR's to test version upgrades. 

I've also added some basic console logging to make skipped/included parts more transparent.

Fixes partially https://github.com/owncloud/docs/issues/3014

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -68,6 +68,7 @@ def build(ctx, latest_version, deployment_branch, base_branch):
                 "pull": "always",
                 "image": "owncloudci/nodejs:11",
                 "environment": {
+                    "BUILD_SEARCH_INDEX": "true",
                     "UPDATE_SEARCH_INDEX": ctx.build.branch == deployment_branch,
                     "ELASTICSEARCH_HOST": from_secret("elasticsearch_host"),
                     "ELASTICSEARCH_INDEX": from_secret("elasticsearch_index"),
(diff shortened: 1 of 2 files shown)


voroyam:
LGTM 👍

voroyam (approved):

phil-davis:
@xoxys I am not familiar with the build and publish of branches like `10.6`.
Does merge CI of `master` re-generate "the whole docs"?
Or does this sort of code need to also be corrected in `10.6` branch to generate `10.6` docs when those are updated?

mmattel:
With a discussion I had with tboerger a while ago:
* Whenever a PR in master get´s merged, a build is started.
* Whenever a PR in one of the branches get´s merged, this merge does not trigger a build.
You have to wait unitl a master merge was done.

I remember that I filed a issue about a change, without looking. The point was, that there should be a nightly job to see if there are merges in branches post merges in master and start a build if so.

phil-davis:
Thanks @mmattel now I remember about that. 

The annoying thing is that usually a PR is merged to `master`, the merge CI of that publishes the updated`master` docs. Then we immediately do backports and none of those get published until someone does another, usually unrelated, change to `master`.

A "workaround" would be that when a PR is ready to merge in `master`, make the backports, let them finish CI, merge all the backports. Then the last step is to merge the PR to master.

mmattel:
Personally I hate such workarounds.
Either any merge should trigger a build or we use a way like the proposal I made.

PR #3020, opened 2021-01-13
Use antora variable from site.yml in maintenance for the latest download version

mmattel:
This PR replaces the hardcoded latest download version which is when built used in https://doc.owncloud.com/server/admin_manual/maintenance/upgrade.html

References: https://github.com/owncloud/docs/pull/3017 + discussion

Backport to 10.6 only

@phil-davis the variable in site.yml was already present, I just needed to use it.


phil-davis (approved):

phil-davis:
@mmattel you may as well backport this. But it is not essential, because the hard-coded version number text in 10.6 branch happens to be correct anyway. Whatever you like to do.

PR #3024, opened 2021-01-13
Reduce list of 10.x.y versions in example upgrade table

phil-davis:
Thus avoids anyone feeling that they need to add a row to this table every minor version release.


mmattel (approved):
👍

phil-davis:
I will backport to 10.6

PR #3040, opened 2021-01-18
Update manual_upgrade.adoc

JanAckermann:
According to this issue owncloud/core#38297, the documentation does not say anything about the app directories must exist and match the apps_paths section in the config.php while upgrading.
This PR fix this issue

PR #3041, opened 2021-01-18
Add note about apps_paths to developer docs also

phil-davis:
Following on from PR #3040 , there is another place that setting up `apps_paths` is described. This is in the developer docs section, so one would hope that a developer will actually look and create the directory. But we may as well put the note here also.


mmattel (approved):

phil-davis:
Backports:
10.5 #3043 
10.6 #3044

PR #3045, opened 2021-01-18
Create wnd_quick_guide.adoc

voroyam:


modules/admin_manual/pages/enterprise/external_storage/wnd_quick_guide.adoc
```diff
@@ -0,0 +1,155 @@
+= WND Configuration Quick Guide
+:toc: right
+:toclevels: 2
+
+== Installation
+
+NOTE: If you are using Ubuntu 20.04 as your OS, you will need to add this repository:
+
+----
+add-apt-repository ppa:ondrej/php
```
mmattel, line 10: why is the ondrej/php ppa a prerequisite?
someone can have a standard php7.4 installation too
the ondrej pps is particular necessary if you are running U18.04 and want to use php7.3(4)...

voroyam: It's not, that's why its behind a "NOTE:"

mmattel: `If you are using Ubuntu 20.04 as your OS` --> `you will need to add this repository`

This means, as I have written, it is a must, not an option
The usage for NOTE highlights the necessity even more

voroyam: Yeah, if you are using ubuntu 20.04 the default php to be installed is the 7.4 one and that one is missing the smb client. That's why you need to add this library.

mmattel: You already have documents about the prerequisites (quick install guides), you just need to reference. no double writing and maintenance needed. What is needed there, add to install the pecl smbclient. which is then used by any of the php sources. No customer has a understanding by telling, throw away your current php installation and add the ppa just because you else would need to do the pecl install. You should write that the pecl smbclient is a necessary prerequisite and redirect for doing so to the install guide. Just my 2c

voroyam: I can't rely on customers reading anything that is not on the page. Your concern is noted.

modules/admin_manual/pages/enterprise/external_storage/wnd_quick_guide.adoc
```diff
@@ -0,0 +1,155 @@
...
+smbclient -U Username -L //Servername
+----
+
+Take the example of attempting to connect to the share named MyData using occ wnd:listen. Running the following command would work:
+
+----
+sudo -u www-data php occ wnd:listen MyHost MyData svc_owncloud password
+----
+
+The command is case sensitive, and that it must match the information from the mount point configuration.
```
enbrnz, line 155: LGTM, just one more addition at the very end in the trouble shooting section:
- if the output of the `occ process-queue ..` command shows `0 Storages found` this means that there was no corresponding external storage configuration found, for example because the casing between calling the process queue and the web interface does not exactly match.

enbrnz (approved):

PR #3075, opened 2021-01-27
Add Important Note for valid CORS entries

JanAckermann:
Hint for users, to setup valid CORS records

### DIFF
--- a/modules/user_manual/pages/personal_settings/security.adoc
+++ b/modules/user_manual/pages/personal_settings/security.adoc
@@ -17,7 +17,23 @@ The {cors-url}[CORS (Cross-origin resource sharing)] white-listed domains sectio
 
 By default, as in the screenshot above, no domains will be listed.
 If you want to add one, or more, add them, one at a time, in the Domain text field, and click btn:[Add].
-You will then seem list listed, as in the screenshot below. 
+You will then see them listed, as in the screenshot below. 
+
+[IMPORTANT] 
+.Valid records:
+====
+[width="70%",cols="40%,60%",options="header"]
+|===
+| Record Scheme            | Example
+| protocol + domain        | \https://example.com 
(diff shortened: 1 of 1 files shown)


JanAckermann:
Backport needed for 10.6

mmattel:
**Request:**
There is a typo in the line above, yes, not part of your PR but could be easy fixed and added:
`You will then seem list listed, as in the screenshot below.` -->
`You will then see them listed, as in the screenshot below.`

**Question:** 
When reading (one of the example) like `https://example.com *protocol* + *domain*`, I do not get it with regards of the additional elements `*protocol* + *domain*`. How does this finally look like? The sentence below states already that the protocol can only be http(s)

mmattel:
OK, I guess I get what you want to express.
What about if you create a table?
Left with the definition and right with an example?
This would make things much more clear.

modules/user_manual/pages/personal_settings/security.adoc
```diff
@@ -18,6 +18,18 @@ The {cors-url}[CORS (Cross-origin resource sharing)] white-listed domains sectio
 By default, as in the screenshot above, no domains will be listed.
 If you want to add one, or more, add them, one at a time, in the Domain text field, and click btn:[Add].
 You will then seem list listed, as in the screenshot below.
```
phil-davis, line 20:
```suggestion
You will then see them listed, as in the screenshot below. 
```

mmattel:
This is how it looks using a table (pushed):
image

phil-davis (approved):

JanAckermann:
> This is how it looks using a table (pushed):
> image
> @phil-davis if ok to you pls approve and merge. Backport only to 10.6 as written above

Had this already, but needed to switch to another problem, so far 😆 
But yes, the Table is much better to understand 👍

PR #3076, opened 2021-01-27
Update instance.sh

jnweiger:
Suggest to simplify the scripting here.
a) remove the unpenetrable and questionable hack from stackoverflow. (I especially frown the unquoted comparison with /* )
b) suggest to also merge the two scripts. It is much more readable to have variables and code in one script. Variables on top, code below.


mmattel:
ad a)
Thanks for the improvement. I will test it locally

ad b)
The reason for not having one script is intentionally!
It is very benificial if you have more than one instance to maintain.
It is much easier having a common task script and a per instance setup script.

mmattel (approved):
Tested and works.
Backport to 10.6 and 10.5 necessary

phil-davis (approved):

PR #3085, opened 2021-01-28
Adjust Provisioning API

JanAckermann:
The Provisioning API will be located under developer_manual/core/apis
The navigation item 'User Provisioning API' at admin_manual/configuration/user/ will link to this page to section #instruction-set-for-users
A important note that the given user must be an admin or sub-admin will be displayed

### DIFF
--- a/modules/developer_manual/pages/core/apis/provisioning-api.adoc
+++ b/modules/developer_manual/pages/core/apis/provisioning-api.adoc
@@ -1,4 +1,4 @@
-= User Provisioning API
+= Provisioning API
 :toc: right
 :page-aliases: go/admin-provisioning-api.adoc 
 
@@ -20,6 +20,8 @@ https://en.wikipedia.org/wiki/Basic_access_authentication[a Basic Auth header]
 to perform any of the functions listed above. The Provisioning
 API app is enabled by default. The base URL for all calls to the share
 API is *owncloud_base_url/ocs/v1.php/cloud*.
+
+IMPORTANT: The Provisioning API can only be accessed by *Admins* and *Sub-Admins*
 
 == Instruction Set For Users
 
--- a/modules/developer_manual/pages/core/apis/provisioning-api.adoc
+++ b/modules/developer_manual/pages/core/apis/provisioning-api.adoc
@@ -21,14 +21,13 @@ to perform any of the functions listed above. The Provisioning
 API app is enabled by default. The base URL for all calls to the share
 API is *owncloud_base_url/ocs/v1.php/cloud*.
 
-IMPORTANT: The Provisioning API can only be accessed by *Admins* and *Sub-Admins*
-
 == Instruction Set For Users
 
 === Add User
 
-Create a new user on the ownCloud server. Authentication is done by
-sending a basic HTTP authentication header.
+Create a new user on the ownCloud server.
+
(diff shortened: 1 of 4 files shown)


mmattel:
BTW, I in the meeting with the link you provided

JanAckermann:
@mmattel i am in the meeting again

mmattel:
The meeting tells me that it is wayting for you...?

JanAckermann:
@mmattel  can you try this link? https://app.gotomeeting.com/?meetingId=380465909

modules/developer_manual/pages/core/apis/provisioning-api.adoc
```diff
@@ -21,6 +21,8 @@ to perform any of the functions listed above. The Provisioning
 API app is enabled by default. The base URL for all calls to the share
 API is *owncloud_base_url/ocs/v1.php/cloud*.
 
+IMPORTANT: The Provisioning API can only be accessed by *Admins* and *Sub-Admins*
```
phil-davis, line 24: This is not quite true. A normal user can query their own information - https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiProvisioning-v1/getUser.feature#L92
And can change their email address... - https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiProvisioning-v1/editUser.feature#L100
And can reset their own password - https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiProvisioning-v1/resetUserPassword.feature#L67

But they cannot add, change, delete or see anything else.

I am not sure how many words, "caveat", should be written here. Because a simple statement is best anyway.

JanAckermann: I check this

JanAckermann: Oh, I don't like this API design, this is a Provisioning API and should only accessible users, that do the provisioning...
Anyways, there is a Important statement in some sections, which says:
```
The Basic Authorization HTTP header must be used to authenticate this request, using the credentials of a user who has sufficient access rights to make the request.
--
```

This Note is missing on other sections where it should be, so I guess, we should put it in the top generally

phil-davis: There needs to be some API where a user can change their own user attributes, like display name, email address, password... so that clients can call it to do the job when a user wants to reset password etc.

mmattel:
@JanAckermann as discussed yesterday, moving the page (+renaming) is ok, but pls created in the navigation a admin local page like `**** xref:core/apis/index.adoc[User Provisioning API]`
with a bit of text where you further link to the final document#section in the development module.

JanAckermann:
@mmattel right on it :+1:

JanAckermann:
Ready for review, backport to 10.6 Welcome

mmattel:
Just running a build to double check, feedback asap

mmattel (approved):
Made a build and checked the result, all fine 👍

phil-davis (approved):

phil-davis:
Backport to 10.6 and 10.5 - I will do it now, takes less than 2 minutes per backport.

phil-davis:
Backports:
10.6 #3089 
10.5 #3090

PR #3092, opened 2021-02-01
remove old branches; leave only the master branch

voroyam:
We want to leave only the master branch for the admin manual.

This should help to maintain the admin manual since you don't have to back port to 3 other branches.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -9,12 +9,7 @@ content:
     - HEAD
   - url: https://github.com/owncloud/docs.git
     branches:
-    - '10.6'
-    - '10.5'
-    - '10.4'
-    - '10.3'
-    - '10.2'
-    - '10.1'
+    - master
   - url: https://github.com/owncloud/android.git
     branches:
     - master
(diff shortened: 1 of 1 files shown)


mmattel:
NO-NO-NO :bomb: :fire: 

As discussed with you via call in talk, this has very negative impacts and should not be done.

Having only a master branch:
1) does not reflect reality
2) does not take into account that different core versions have different backgrounds and setttings
3) does not take into account that unreleased but merged developments from repros need documentation
4) reduces documentation quality a lot. how do you want to implement version specific differences
5) increases support load, as there is no documentation about the particular releases the customer have. Only master which would then contain everything up to unavailable but merged features/changes.

Everything what is not documented properly will fall back to support as the customer cant see what is relevant for him in which release. Having multiple documentation versions is normal in software and in many other branches and nothing special to ownCloud. I am highly warning about such a step as it will create a negative branding. In times where competitors a next door, a well made documentation is a business card and positive differenciator. 

@pmaier1 @micbar @phil-davis @hodyroff @dragotin

phil-davis (changes requested):
I thought that the policy was to maintain the latest and latest-1 in addition to master.
So leaving 10.6 and 10.5 in this.

Issue #2774, opened 2020-09-07
LDAP: Renaming of groups not supported / propagated


voroyam:
We ned to document that we don't support the renaming of LDAP groups.

Here is what happens:

-  The new name does not appear anywhere in ownCloud - neither in the LDAP config nor in the sharing dialogue 
- If you clear the user / group mapping the users in the renamed group get deleted / disabled and then added back / re enabled

mschreml:
I think I don't have to descripe what happens. Just the "Note:" should be enough, right?

Like:

NOTE: Renaming of groups is not supported. We recommend to use [Custom Groups](https://doc.owncloud.com/server/user_manual/files/webgui/custom_groups.html#creating-custom-groups)

If the text is ok, canI write it in the Troubleshooting Area on [User Auth LDAP](https://doc.owncloud.com/server/admin_manual/configuration/user/user_auth_ldap.html#troubleshooting-tips-and-tricks) or should I write it in [LDAP Integration](https://doc.owncloud.com/server/admin_manual/configuration/server/app_commands/ldap_integration_commands.html)

voroyam:
Just the note about Renaming is okay. Important is the propagated.

For example.

I sync LDAP users and LDAP groups to ownCloud
I then change the name of the LDAP group in LDAP
The change will not be visible in ownCloud
So the change is not propagated from LDAP to ownCloud

So a note or a warning in the User Auth with LDAP in the Group Filter Section is fine with me.

mschreml:
@voroyam done

PR #3110, opened 2021-02-08
add note to User-Auth with LDAP doc

mschreml:
Issue: #2774


modules/admin_manual/pages/configuration/user/user_auth_ldap.adoc
```diff
@@ -316,6 +316,8 @@ y groups found::
   This tells you approximately how many groups will be available in ownCloud. 
   The number updates automatically after any change.
 
+NOTE: Renaming of LDAP-Groups is not supported. Changes in LDAP are not propagated to ownCloud.
```
mmattel, line 319:
```suggestion
NOTE: Renaming of LDAP-Groups on the LDAP Server is not supported. Changes like renaming in LDAP will not be propagated to ownCloud.
```

mmattel:
Has this issue been reported to core or ldap?
@jvillafanez or @butonic, can you take a look on the source side?

phil-davis (commented):
The words are fine, and I believe that this is true! (And I think it has always been true)
Can "someone" confirm that there has always been this restriction?
And what does happen if the LDAP group is renamed? Does the "old" group name in ownCloud keep working but never get any more member updates? Or?

mmattel:
Usually an LDAP Group Name has a GID in the backend. The question is, which one was used and how...

voroyam:
> And what does happen if the LDAP group is renamed? Does the "old" group name in ownCloud keep working but never get any more member updates? Or?

Old group name keeps working.

voroyam (approved):

PR #3124, opened 2021-02-09
Recent changes and additions in config.sample from core

mmattel:
As the name tells, there have been changes and additions in config.sample from core which need to be transported to docs.

References:
https://github.com/owncloud/core/pull/38337
https://github.com/owncloud/core/pull/38379

**NO backport, master only**


phil-davis (approved):
Thanks!

PR #3156, opened 2021-02-16
Reorganised the document

voroyam:
- removed unnecessary paragraphs 
- streamlined the document for installation and configuration

PR #3162, opened 2021-02-17
page rework

voroyam:
clean new PR from https://github.com/owncloud/docs/pull/3156

### DIFF
--- a/modules/admin_manual/pages/configuration/server/virus-scanner-support.adoc
+++ b/modules/admin_manual/pages/configuration/server/virus-scanner-support.adoc
@@ -3,95 +3,32 @@
 :page-aliases: configuration/server/antivirus_configuration.adoc
 :clamav-url: http://www.clamav.net/index.html
 
-== Overview
+== Introduction
 
 {clamav-url}[ClamAV] is the only _officially_ supported virus scanner available for use with ownCloud. 
-It:
+
+Feature List:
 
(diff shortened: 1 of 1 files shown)


enbrnz (approved):
LGTM

modules/admin_manual/pages/configuration/server/virus-scanner-support.adoc
```diff
@@ -144,14 +67,19 @@ For example, to run it every hour at 47 minutes past the hour, add the following
...
 
 image:apps/files_antivirus/antivirus-app.png[image]
 
-TIP: The Anti-Virus app can also be downloaded, installed, and enabled manually.
+--------
+
+To install the App via occ command, execute:
+
+----
+occ market:install files_antivirus
```
mmattel, line 81: Please use `{occ-command-example-prefix}` instead of `occ`

voroyam: discussed with the support team - this adds only the "sudo -u www-data" to the command which is no necessary if the user / customer followed our installation guide and has the wrapper script.

mmattel: If a "user" has more than one instance on the same system, this does not work anymore. It would be better to have a statement in your docs that if the wrapper is used, no preceedings are necessary.
All occ commands shouldbe written according the example provided to be in line with the tons of examples in docs.

PR #3165, opened 2021-02-17
Update ocs-share-api.adoc

nitro2010:
invalid OCS URL to e.g. create share or get informations about shares


mmattel:
Fixed with https://github.com/owncloud/docs/pull/3488 (The share api endpoint does not end in `/pending`)
Closing

PR #3176, opened 2021-02-18
Fail on warnings from asciidoctor

phil-davis:


mmattel:
There are a couple of things to mention.

1. Pls see: https://github.com/asciidoctor/asciidoctor/issues/3078
here you find clearly stated, that setting that flag does not stop building, but the return code will change (only)

2. I am not convinced that a build should break when a warning occurs. It is more, that a contributor should take care about a build warning.

phil-davis:
If CI does not stop on warning, then warnings will get introduced and we will not notice it. So, if we care about warnings, then CI should stop on warning.

A non-zero return code will stop the drone pipeline - that will make the CI result go red. Whatever output has been produced will be thrown away on the drone agent.

phil-davis:
Closing - I don't really know how to achieve this.

PR #3177, opened 2021-02-18
Update full_text_search_commands.adoc

jnweiger:
Remove useless reference to docker-compose.

Appreciate the use of {occ-command-example-prefix} to make everything uniform! But ...

I'd also suggest to simplify {occ-command-example-prefix}  to be just `occ` and add an introductory section about occ, explaining
1) occ *should* be the /usr/bin/occ wrapper script. 
2) in any case occ commands *must* be execute as the same user as the webserver (e.g. apache) is running. The wrapper script takes care of that.


jnweiger:
@EParzefall Huhu :-)

mmattel (approved):

mmattel:
Backporting to 10.5 and 10.6 necessary

phil-davis:
Backports:
10.6 #3183 
10.5 #3184

PR #3201, opened 2021-02-24
Change RestartSec

enbrnz:
Already changed it in the other docs because of some systemd internals: if a service fails more than 5 times within 10 seconds, it won't be restarted.

Backport?


enbrnz:
wrong branch

PR #3216, opened 2021-02-26
Increased TOC Level to 3

voroyam:
...to increase readability. So you can click more stuff :]


xoxys:
Shouldnt we keep it equal to all 3 manuals?

voroyam:
> Shouldnt we keep it equal to all 3 manuals?

valid suggestion. will check and make a new PR.

PR #3217, opened 2021-02-26
Added new metrics documentation.

EParzefall:
I've cleaned up my old branch (hopefully!), here's a new one.
This obsoletes PR #3200.

PR #3200, opened 2021-02-23
Create metrics.adoc

EParzefall:
New doc for Metrics extension.

PR #3218, opened 2021-02-26
Improved metrics documentation

mmattel:
I did some improvements for https://github.com/owncloud/docs/pull/3217:

Nav location
Text
Lists
CamelCase

Regarding curl, the former multiline for the url was not good and I have not found a solution to make this better.

The rendered result looks now much better.

This is the new location for the navigation. The reason for doing so is, what if there is another monitoring app for a different purpose, and it follows the look and feel for the other items.
image

As I have seen that Metrics is not on the marketplace, which means it is not availabe for 10.5/10.6, therefore no backporting necessary.


mmattel:
Based on chats in talk with @jnweiger, Metrics will be announced soon. Dependent on the ownCloud server requirements for Metrics, backports of this and #3200 may be necessary in the sort oder #3200 then #3218

modules/admin_manual/pages/enterprise/monitoring/index.adoc
```diff
@@ -0,0 +1,4 @@
+:section-title: Enterprise Monitoring
+:section-preamble-ender: to monitor the ownCloud Enterprise installation
```
EParzefall, line 2: I like the idea, but should be approved by @pmaier1.

EParzefall (approved):
I'm happy with it and will go ahead with merging. Not likely that @pmaier1 will disagree.

jnweiger:
This changes the metrics url to https://doc.owncloud.com/server/latest/admin_manual/enterprise/monitoring/metrics.html 
a) We fix this url in the metrics appinfo/info.xml via https://github.com/owncloud/metrics/pull/101
b) Marketing will wait for the 10.6 backport so that `/latest/`will work.

modules/admin_manual/pages/enterprise/monitoring/index.adoc
pmaier1: The Metrics app is rather a **reporting** tool than a monitoring tool. Monitoring has a more technical connotation (load, response times, etc.), while the metrics in here rather tells you something about the usage of the installation. Please change this in a follow-up PR.

mmattel: ok, takes only a view minutes

mmattel: PR made, see #3225

PR #3225, opened 2021-03-01
Reporting not Monitoring

mmattel:
References: https://github.com/owncloud/docs/pull/3218#discussion_r584557808

With feedback from @pmaier1, metrics is about reporting and not monitoring.
This PR changes the text and location accordingly.

Backport to 10.5 and 10.6 necessary

@jnweiger pls adopt the appinfo doc path to: https://doc.owncloud.com/server/latest/admin_manual/enterprise/reporting/metrics.html

PR #3230, opened 2021-03-01
[10.5] [PR 3225] Reporting not Monitoring

mmattel:
Backport of PR #3225


jnweiger:
Confirmed. Metrics works fine with 10.6 and 10.5 (but is not compatible with 10.4.x))

Issue #291, opened 2018-05-24
Running drone tests locally


PVince81:
- how to run tests in current directory

    - drone tests

    - download drone https://github.com/drone/drone-cli and have docker available locally

    - look at .drone.yml for possible values or the variables as they appear in the CI drone runs

    - command: PHP_VERSION=7.1 DB_TYPE=postgres TEST_SUITE=phpunit drone exec

    - acceptance tests

    - need local OC setup to have run (occ maintenance:install)

    - cd tests/acceptance && ./run.sh

    - can add feature file name and optionally line number: ex ./run.sh features/sharing-v1.feature:123

    - from drone use TEST_SUITE=api-acceptance-tests

    - smashbox tests

    - https://github.com/owncloud-docker/smashbox

    - docker run -e SMASHBOX_URL=localhost:9680 -e SMASHBOX_USERNAME=admin -e SMASHBOX_ACCOUNT_PASSWORD=admin owncloud/smashbox lib/test_basicSync.py <= will not work, docker, does not understand localhost, and needs admin password

    - docker run -e SMASHBOX_URL=172.42.16.124:80/octest -e SMASHBOX_USERNAME=admin -e SMASHBOX_ACCOUNT_PASSWORD=admin -e SMASHBOX_PASSWORD=admin owncloud/smashbox lib/test_basicSync.py 

    - docker run -e SMASHBOX_URL=172.42.16.124:80/octest -e SMASHBOX_USERNAME=admin -e SMASHBOX_ACCOUNT_PASSWORD=admin -e SMASHBOX_PASSWORD=admin owncloud/smashbox lib/test_basicSync.py -t 0 <= specific test 

    - in the future will be unified in a single command

Something to put in the developer docs maybe.
We don't have QA docs yet, right ?

@patrickjahns

voroyam:
@phil-davis any progress on this?

PR #3272, opened 2021-03-08
drone testing

voroyam:
fixes https://github.com/owncloud/docs/issues/291

### DIFF
--- a/modules/developer_manual/pages/testing/drone.adoc
+++ b/modules/developer_manual/pages/testing/drone.adoc
@@ -1,0 +1,85 @@
+= Drone Testing
+:toc: right
+
+:drone-cli-url: https://github.com/drone/drone-cli
+:smashbox-url: https://github.com/owncloud-docker/smashbox
+
+== Introduction
+
+This document will show you how to run drone tests locally.
+
+== Setup Drone
+
+- Download the {drone-cli-url}[command line client] for the Drone Continuous Integration Server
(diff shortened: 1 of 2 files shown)


modules/developer_manual/pages/testing/drone.adoc
```diff
@@ -0,0 +1,85 @@
...
+== Introduction
+
+This document will show you how to run drone tests locally.
+
+== Setup Drone
+
+- Download the {drone-cli-url}[command line client] for the Drone Continuous Integration Server
+- Have a local docker available
+- Look at file `.drone.yml` for possible values or the variables as they appear in the CI drone runs
+- Command: `PHP_VERSION=7.1 DB_TYPE=postgres TEST_SUITE=phpunit drone exec`
```
enbrnz, line 16: [PHP 7.2 is already no longer supported](https://www.php.net/eol.php), if in any way possible I would try to use a supported PHP version in a new document.

enbrnz (commented):
Otherwise, unfortunately I don't understand what is happening here and I also am unable to test this. Perhaps it would be better to ask somebody who is using this to check for the factual correctness of this article?

modules/developer_manual/pages/testing/drone.adoc
voroyam: changed to php 7.4

mmattel (approved):

Issue #2294, opened 2020-02-07
Clarify doc about sharing


oliviercailloux:
I recently [tried to use file sharing](https://central.owncloud.org/t/how-does-shared-folder-work/23490) and found two points quite unclear in the user manual.

First, I did not realize, reading the user manual, that the [Guest sharing](https://doc.owncloud.org/server/10.4/user_manual/files/webgui/sharing.html#sharing-files-with-guest-users) feature enables my guest to use the ownCloud client to sync. I thought it would only provide access to the server web interface.

Second, I think that the user manual never mentions that it is hopeless to try to use [Federated sharing](https://doc.owncloud.org/server/10.4/user_manual/files/federated_cloud_sharing.html) when the administrator of the server has not explicitly trusted the other server before. 

It seems it is necessary to read the Administrator manual to find [this information](https://doc.owncloud.org/server/10.4/admin_manual/configuration/files/federated_cloud_sharing_configuration.html) (and even if you reach there, it is hard to find as it is to be found under the `Configuration` / `Files` section whereas it does not appear under the `Files` section in the User manual).

 As a user (and not an administrator) of the server, I didn’t think about reading the Administrator manual in detail (I only had taken a cursory look at it).

Not sure how to correct it but I thought I’d mention my experience in case someone sees how to improve this aspect.

settermjd:
Hi @oliviercailloux. Thanks for taking the time to detail your experiences in this issue. In collaboration with the developers, I'll review this section of the documentation and make all the required changes.

voroyam:
Hello @oliviercailloux ,

so from what I understood you want the documentation to mention:

- Guests can use the ownCloud client to sync.

The place for this information is probably [here](https://doc.owncloud.org/server/user_manual/files/webgui/sharing.html#sharing-files-with-guest-users)

Do you agree?

----

Then you want to have the documentation mention that federated sharing has to be established before creating a new federated share. Correct?

probably a hint somewhere in [here](https://doc.owncloud.org/server/user_manual/files/federated_cloud_sharing.html#how-federated-sharing-works)

Tell me if my assumptions are correct, and I get to work right away.

oliviercailloux:
Yes, and yes! And ideally, let the reader know how she can check whether her adminsitrator has established federated sharing.

voroyam:
Other than asking - I don't know if there is a check a user can do.

PRs are open, thanks for noticing and notifying us.

PR #3280, opened 2021-03-09
added note about sync client and guest users

voroyam:
fixes https://github.com/owncloud/docs/issues/2294

added note about the guest users being able to use the sync client

### DIFF
--- a/modules/user_manual/pages/files/webgui/sharing.adoc
+++ b/modules/user_manual/pages/files/webgui/sharing.adoc
@@ -86,6 +86,8 @@ permissions applied, except for the ability to share with other users.
 
 image:guest-users/content-shared-with-guest-user.png[Content shared with a guest user]
 
+NOTE: Guest users can also use the **Desktop/IOS/Andriod Sync Clients** to access synced files locally.
+
 == Updating Shares
 
 To change any of the properties of a share, again, you first need to


modules/user_manual/pages/files/webgui/sharing.adoc
```diff
@@ -86,6 +86,8 @@ permissions applied, except for the ability to share with other users.
 
 image:guest-users/content-shared-with-guest-user.png[Content shared with a guest user]
 
+NOTE: Guest users can also use the **desktop sync client** to keep their files up to date.
```
mmattel, line 89:
```suggestion
NOTE: Guest users can also use the **Desktop/IOS/Andriod Sync Clients** to access and/or sync files locally.
```

mmattel:
Backport to 10.5 and 10.6 necessary

modules/user_manual/pages/files/webgui/sharing.adoc
voroyam: I don't think you can access your files without synching them. By connecting your account to the server your are synching the server state to your device. I think we can leave out the and/or. 

What do you think?

mmattel: good point, I have updated the suggestion. thanks!

PR #3283, opened 2021-03-09
Update examples of Assert in acceptance tests

phil-davis:
Later releases of PHPunit Assert library have a different namespace. Update the examples of it use in acceptance tests.

I noticed this while looking at the remaining ToDos for issue #156 

No need to backport - these developer docs are really intended to be used along with the current core master.


mmattel (approved):

mmattel:
> these developer docs are really intended to be used along with the current core master.

just for the record, not knowing if, but is there a statement about this to have it clear?

PR #3284, opened 2021-03-09
Develop with latest core master

phil-davis:
Add some words to the developer docs intro pages so that readers will understand that the developer docs are intended for use with current core master. (There might be all sorts of different stuff if someone checks out an old release branch and tries to develop or run tests from those)

### DIFF
--- a/modules/developer_manual/pages/core/index.adoc
+++ b/modules/developer_manual/pages/core/index.adoc
@@ -1,3 +1,5 @@
 = Core Development
 
 In this section you will find all the details you need to develop ownCloud’s core.
+
+You should develop starting with the latest code in the core master branch. The core development documentation is written assuming that you are doing that.
--- a/modules/developer_manual/pages/core/introduction.adoc
+++ b/modules/developer_manual/pages/core/introduction.adoc
@@ -1,4 +1,4 @@
 = Introduction
 
-Please make sure you have set up a xref:general/devenv.adoc[Development Environment].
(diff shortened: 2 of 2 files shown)


modules/developer_manual/pages/core/index.adoc
```diff
@@ -1,3 +1,5 @@
 = Core Development
 
 In this section you will find all the details you need to develop ownCloud’s core.
+
+You should develop starting with the latest code in the core master branch. The core development documentation is written assuming that you are doing that.
```
mmattel, line 5:
```suggestion
IMPORTANT: You should develop starting with the latest code in the core master branch. The core development documentation is written assuming that you are doing that.
```
Not a must highlighting with important, but creates good visibility. Reject if not ok

mmattel (approved):

modules/developer_manual/pages/core/index.adoc
phil-davis: IMO this is just a note/recommendation/information aimed at developers. So not really necessary to write "IMPORTANT".

mmattel:
Post thoughts regarding backports. Would´nt it be good to at least BP to 10.6 (as 10.5 will be phased out soon due to upcoming 10.7) and someone who is reading "accidentially" 10.6 should get optical notified?

phil-davis:
Yes, it only takes a minute! Backports:
10.6 #3285 
10.5 #3286

PR #3337, opened 2021-03-12
Fix to dynamically create the attribute-to-argument list for asciidoctor-pdf

mmattel:
The following fix makes rendering of attributes in pdf documents working!

**This issue has not been identified since we started creating the PDF docs with Antora**

**Background**
We use _site wide attributes_ defined in site.yml, which are intensively used in all modules. These attributes harmonize the look and feel and ease maintenance a lot. Example: definition: `minimum-php-version: 7.2` usage: `{minimum-php-version}` rendering: `7.2` Means that any change we do in the definition will immediately affect all pages using the attribute `{minimum-php-version}`. This works great when building the docs for html.
But when creating the pdf´s, a different script is used which does not and cannot use site.yml. Nobody has taken care about these attributes and when rendered, you do not get `7.2` but `{minimum-php-version}` which looks awful.

**Solution**
One possible way to go was to hard code the attribute list (key/value pair) into the argument list of the creation call. This has the drawback, that there would have been two places to maintain changes. This path would have been relatively easy to build, but was not chosen by obvious reasons.

The chosen way was to extract on each build the attributes from site.yml and create a dynamically, script usable argument list for the build. The transformation script is based on bash and therefore easy implementable. (If someone has a better idea by including a program via node manager, I am fine with, but should be a different task)

**Result**
After testing, all attributes used are rendered correctly !
As a bonus, I added a small test script which just prints the attributes as arguments in a human readable form, so we can easily compare in case of issues. The adopted main script also contains some removable remarks for intense testing having all arguments not just the ones coming from the attribute transformation.  

**Post Note**
I have contact with the developer of Antora who also provides the pdf creation script and was told, that there is work in progress for an updated version. This version should be able to use the definitions from site.yml. But this is something for the future.

Kudos to @voroyam as he indirectly made me aware of this issue

Backporting to 10.5 and 10.6 necessary

@dragotin fyi (as discussed)

### DIFF
--- a/bin/cli
+++ b/bin/cli
@@ -17,11 +17,21 @@ STYLE="owncloud"
 STYLES_DIRECTORY="resources/themes"
 LANGUAGE=
 
+DIR="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
+# define path and file of .yml to parse
+YMLFILE="$DIR/../site.yml"
+
 ERR_UNSUPPORTED_LANGUAGE=20
 ERR_UNSUPPORTED_MANUAL=21
 ERR_UNSUPPORTED_ACTION=22
 ERR_LINTER_NOT_AVAILABLE=23
 ERR_NO_LANGUAGE_SPECIFIED=24
+
+# yamlparse.sh is a script to get a list of key/value pairs from site.yml which
+# will then be used as additional dynamically created argument set for asciidoctor-pdf.
+# Else the attributes used can´t be resolved and are printed like {key}
+
+source $DIR/yamlparse.sh
 
 if [[ -z "${VERSION:-}" ]]; then
     if [[ -n "${DRONE_TAG:-}" ]]; then
@@ -157,6 +167,11 @@ function build_pdf_manual()
     local book_file="books/ownCloud_${manual_infix}_Manual.adoc" 
     local nav_file="modules/${manual}_manual/nav.adoc"
 
+    # Get the dynamic list of attributes from site.yml
+    # The output after sed is a string like -a key=value -a key=value ...
+    # For testing, also use the yamltest.sh script
+    local attributes=("$(parse_yaml $YMLFILE | grep asciidoc_attributes | sed 's/asciidoc_attributes_/-a /g' | sed 's/\")/\"/' | sed 's/=(/=/' | sed "s/\"'/\'/" | sed "s/'\"/\'/")")
+
     if [[ "$DRY_RUN" == true ]]; then
         echo "Manual Generation - **DRY RUN**"
         echo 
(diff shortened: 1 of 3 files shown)


voroyam:
testing it now. 

> I have contact with the developer of Antora who also provides the pdf creation script and was told, that there is work in progress for an updated version. This version should be able to use the definitions from site.yml. But this is something for the future.

Yeah, @xoxys noticed that in a new version of Antora there are some improvements, and we should test run the alpha version of it.

voroyam:
```
dmitry@MBP ~/g/docs (fix_pdf_attributes)> git pull origin fix_pdf_attributes
From github.com:owncloud/docs
 * branch                fix_pdf_attributes -> FETCH_HEAD
Already up to date.
dmitry@MBP ~/g/docs (fix_pdf_attributes)> bin/cli -m
readlink: illegal option -- f
usage: readlink [-n] [file ...]
bin/cli: line 34: ./yamlparse.sh: No such file or directory
```

voroyam:
strange, on my test server it just works. 

does this

```
DIR="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
```

require something?

xoxys:
Not tested it. Personally I think such bash "magic" is sometimes hard to debug and has some potentials for issues related to untested corner cases. Especially for the yaml parsing part I would suggest something like https://mikefarah.gitbook.io/yq/ and give the antora 3 alpha release a try, but thats just my two cents.

Besides that I dont think these change affects the docker or CI stack.

voroyam:
fully agree with @xoxys. if something goes wrong and you have to debug - it's hard to understand where the error is coming from when you have complex scripts that you did not write.

but, if it solves the issue with the variables not being rendered in the PDF version - okay. 
I personally could live without the variables but since this is not my decision I have to go with it.

voroyam:
tested in the html rendered doc on my local server and had a glance at the user, admin and developer pdfs.

voroyam:
I think a maybe not intended side effect is that you can no longer test the PDF versions locally without installing a bunch of stuff.

mmattel:
@voroyam 
On Ubuntu
```
bin/cli -m
Generating version 'master' of the admin manual, dated: March 12, 2021
Generating version 'master' of the developer manual, dated: March 12, 2021
Generating version 'master' of the user manual, dated: March 12, 2021
```

> does this ... require something?

No, it resolves the path the running script is located independent if it is a symbolic link or not.
But I see that you are using a Macbook Pro. OSX is known having only a limited set of options and arguments for commands which are elswhere fully available. You may need to update commands using homebrew.

voroyam:
Ah, okay. MacOS got me again :D

voroyam (approved):

xoxys (approved):

mmattel:
Additional Note: you need to make the two scripts executable with `sudo chmod +x scriptname.sh` ... This can only be done locally.

EParzefall (approved):
Great to get this fixed.

Issue #3278, opened 2021-03-08
New docs branch needed: 10.7


jnweiger:
The release process for 10.7.0 has started. Entering beta phase now.

When we enter RC phase I'd like to have https://doc.owncloud.com/server/10.7 operational.

phil-davis:
Agree, best to leave it until "close to release" to minimize the amount of constant backports that have to be done.

jnweiger:
The beta tarballs currently contain the 10.6 user manual pdf. 
This needs to become a 10.7 manuals when "close to release" -- challenge accepted :-)

mmattel:
FYI, I would like to get #3337 merged first, as it enables proper pdf creation, also for 10.5.
When this is done, we can go for 10.7 as with that, 10.5 will be phased out (keeping only two releases active)

jnweiger:
10.7.0 reached Codefreeze. Buidling RC1. 
Please prepare the 10.7 branch do be online by the end of this week. In particular an shiny new pdf is needed when building the final 10.7.0
https://doc.owncloud.com/server/10.7/user_manual/ownCloud_User_Manual.pdf

mmattel:
Closing, fixed by #3367

PR #3367, opened 2021-03-17
Necessary Branch Changes for 10.7

mmattel:
This PR changes the workflow for branches to include 10.7 and drop 10.5.
This means, that 10.5 is frozen but still accessible if you enter the link on the web manually.
Any new backports, if necessary, will now go to 10.6 and 10.7

There is already a 10.7 branch made based on origin/master containing the proper setup for 10.7 only, which I pushed just right before.

When this PR is merged, the situation is set.

Post merging, we need to unprotect 10.5 (see the repo settings) and rename it to `x_archived_10.5`

Fixes: https://github.com/owncloud/docs/issues/3278

FYI: @micbar @pmaier1 @EParzefall @mschreml @voroyam

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -4,7 +4,7 @@ def main(ctx):
     # Version shown as latest in generated documentations
     # It's fine that this is out of date in version branches, usually just needs
     # adjustment in master/deployment_branch when a new version is added to site.yml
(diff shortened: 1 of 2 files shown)


phil-davis (approved):
Looks reasonable

PR #3374, opened 2021-03-18
Update ubuntu_20_04.adoc

jnweiger:
use sudo -E, so that `occ us:res demo --password-from-env` would work.


jnweiger:
In branch 10.6 the install of 10.7 is described.

image

Is this an intentional sneaky hint that people should update? :-)

EParzefall (approved):
LGTM

EParzefall:
Backport to 10.7 needed, but not to 10.6. Will create a PR for 10.6 with the corrected installation URL.

PR #3378, opened 2021-03-18
Update migrating.adoc

jnweiger:
Cleanup of rsync options: This guide asks for non-standard rsync options without explaining them.
-A preserve ACLs
-x do not cross filesystem boundaries
-t preserve timestamps

For transfering the mysql .bak file:

  * ACLs, filesystem borders, and timestamps are irrelevant.
  * Suggest to use -v instead, to see progress, in case the file is large.

For transfering the folders:

  * ACLs should also be irrelevant, as we instruct to chown www-data users;
  * -x is potentially harmful - we should *not* stop at filesystem borders, if any.
  * -t and -v are nice to have; -a is correct.

### DIFF
--- a/modules/admin_manual/pages/maintenance/migrating.adoc
+++ b/modules/admin_manual/pages/maintenance/migrating.adoc
@@ -163,7 +163,7 @@ Then, export the database to the new server.
 
 [source,console]
 ----
-rsync -Aaxt owncloud-dbbackup.bak root@new_server_address:/var/www/owncloud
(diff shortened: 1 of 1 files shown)


mmattel (approved):

mmattel:
Backporting to 10.6 and 10.7 necessary
@mschreml

mschreml:
backported to:

10.7 - #3381
10.6 - #3382

PR #3402, opened 2021-03-26
don't change /usr/lib/systemd/system/httpd.service

grischdian:
It's not a good idea to change the service-file which was installed by the package.
This file may needs new configuration options after an upgrade but the packagemanager will not 
change the file if you changed it manually.

There are 3 diffrent ways to do it right - systemd will detect the new files automatically but you need to run "systemctl daemon-reload" after way 1 and 2. 

- copy /usr/lib/systemd/system/httpd.service to /etc/systemd/system and do the changes here
  -> not a good idea, same reason as above
- manually create /etc/systemd/system/httpd.service.d/ and a file in this folder e.g. override.conf 
  where you are able to override only the parameters you wanna change
- systemctl edit httpd will do the same but it will create the httpd.service.d and the override.conf for you

I would prefer the easiest way and that is systemctl edit httpd


mmattel:
will a review asap

mmattel (approved):

mmattel:
I improved the changes and the text a bit and checked with a local antora build

Needs forward porting to master and 10.7

phil-davis (approved):

PR #3426, opened 2021-03-31
Switch Desktop branch to 2.8

TheOneRing:


mmattel:
@TheOneRing 
Q1: we have shown in the past, 2 last versions of client. With this PR you are now only having one. Is this intended?
Q2: is 2.8 based on 2.7 so any new upcoming changes in 2.8 can rely on past changes made (and I guess there will be)?

TheOneRing:
> 
> 
> @TheOneRing
> Q1: we have shown in the past, 2 last versions of client. With this PR you are now only having one. Is this intended?

As development on 2.6 stopped and only support one version at a time, I don't think we need to run ci on the last two versions.
> Q2: is 2.8 based on 2.7 so any new upcoming changes in 2.8 can rely on past changes made (and I guess there will be)?

Not sure I get your point, but yes everything form 2.7 is in 2.8 and 2.9, changes are always populated forward.

mmattel:
> yes everything form 2.7 is in 2.8 and 2.9

fine, this is the info I needed 👍 

> 2.6 stopped and only support one version at a time

I just want to get sure, if this is the way to go, we are fine with 👍

mmattel (approved):

dschmidt:
Why backports? 

The menus are only generated from master.

mmattel:
I know, that is just to keep things consistant. It has no technical impact.

dschmidt:
tbh to me it seems like a waste of time :-/

PR #3438, opened 2021-04-02
Fix Collabora Secure View file naming

mmattel:
This fixes the file naming to that what the page is about, keeping the original link and redirecting to the new page. The document is about Collabora Secure View and not about Collabora Integration like the Microsoft Office Integration document. This is now corrected.

Backport to 10.6 and 10.7 necessary


mschreml (approved):

mschreml:
backported to:

10.7 - #3439
10.6 - #3440

PR #3442, opened 2021-04-06
document improvement

voroyam:
Changes:

- I changed  `owncloud-old-version-number` to `backup_owncloud` for convenience. If I go to the ownCloud folder and use tab completion I always get ownCloud-something because I update my server. If the back up copy of ownCloud is in the backup_ownCloud folder - I won't get that problem. I assume that I am not the only one with this issue.
- I removed the enterprise part of the upgrade docs because now we have only one single server package for enterprise and community.
- I restructured the re-enabling part of the document because it was weird **tripple** bullet points with a warning at the end. Now it looks unified.

this is how it looks now
image


mmattel (approved):

mmattel:
Backport needed to 10.7 and 10.6

PR #3446, opened 2021-04-06
Adding the occ commands for the Metrics app

mmattel:
Adding the necessary occ commands for the Metrics app

Relocated all files for the occ commands so they can be found much easier - No change in SEO!

Tested by making a build

Backport to 10.7 only


EParzefall (approved):
Looks good.

PR #3449, opened 2021-04-07
Do not register ondrey repo twice

jnweiger:
the add-apt-repostory command effectively does the same as the echo|tee command. 

also, remove sudo, as we say in the beginning, that we assume the user is logged in as root.


voroyam (approved):

voroyam:
backport to 10.6 and 10.7 needed

jnweiger:
Thanks!

PR #3488, opened 2021-04-19
The share api endpoint does not end in `/pending`

butonic:
`pending` is just relevant for listing the pending shares ... currently the page uses it for all sharing API calls...

I think the pending section might now need to append `/pending` somewhere...


mmattel (approved):

mmattel:
Backport to 10.6 and 10.7

PR #3511, opened 2021-05-03
User manual cleanup because of files on the wrong location

mmattel:
The `admin_audit` documentation was double located not only in the admin manual but also in the user manual and totally outdated. These file(s) got removed. In addition, some files in the user doc were saved in the wrong subfolder. Those got relocated and proper relinked in the navigation.

Needs backporting to 10.6 and 10.7


phil-davis (approved):
Deleted doc is debugged doc :)

PR #3583, opened 2021-05-20
Do not suggest cron every minute

jnweiger:
Also: e.g. files_lifecycle needs a daily cron job, which is not done by occ system:cron -- it needs an exlicit cronjob entry.
Do we want to give an example for this  here or in the files_lifecycle section?

### DIFF
--- a/modules/admin_manual/pages/configuration/server/background_jobs_configuration.adoc
+++ b/modules/admin_manual/pages/configuration/server/background_jobs_configuration.adoc
@@ -47,14 +47,14 @@ Using the operating system Cron feature is the preferred method for executing re
 This method enables the execution of scheduled jobs without the inherent limitations which the
 web server might have.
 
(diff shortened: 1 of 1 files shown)


mmattel (approved):

mmattel:
I guess backport to 10.6 and 10.7 (pls do not forget to mantion)

> files_lifecycle needs a daily cron job, which is not done by occ system:cron -- it needs an exlicit cronjob entry.

:thumbsup: I recommend to add this in 
https://doc.owncloud.com/server/admin_manual/enterprise/file_management/files_lifecycle.html

pls see https://github.com/owncloud/docs/issues/3584 - all occ commands for `lifecycle:` are missing

jnweiger:
Please backport to 10.6 and 10.7  -- files_lifecycle min version is 10.5

Issue #3494, opened 2021-04-20
Metrics App is not compatible with IE 11


pmaier1:
## WHAT Needs to be Documented?
A hint in the Metrics app documentation that the integrated admin dashboard is not compatible with IE 11 as new technologies have been used that are not supported by IE 11.

## WHERE Does This Need To Be Documented (Link)?
https://doc.owncloud.com/server/admin_manual/enterprise/reporting/metrics.html

## WHY Should This Change Be Made?
To make users aware of the limitations.

## (Optional) What Type Of Content Change Is This? 
- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [x] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

@mmattel

mmattel:
References: https://github.com/owncloud/metrics/issues/109

mmattel:
Closing with #3618

PR #3618, opened 2021-06-04
Note that IE 11 is not compatible with Metrics

mmattel:
References: https://github.com/owncloud/docs/issues/3494 (Metrics App is not compatible with IE 11)

Adda a note about that fact.

Backport to 10.7 only

### DIFF
--- a/modules/admin_manual/pages/enterprise/reporting/metrics.adoc
+++ b/modules/admin_manual/pages/enterprise/reporting/metrics.adoc
@@ -30,7 +30,7 @@ special setup in order to not put the production database under huge load when g
 this, please replicate your installation (application server + read-only database) and install/use the
 Metrics App on the replica.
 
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/enterprise/reporting/metrics.adoc
```diff
@@ -30,6 +30,8 @@ special setup in order to not put the production database under huge load when g
 this, please replicate your installation (application server + read-only database) and install/use the
 Metrics App on the replica.
 
+NOTE: Consider that IE 11 is not compatible with Metrics, because new web technologies have been used that are not supported by IE 11.
```
phil-davis, line 33:
```suggestion
NOTE: IE 11 is not compatible with Metrics, because new web technologies have been used that are not supported by IE 11.
```
IMO less words reads better.

EParzefall: Agreed.

EParzefall (approved):
Otherwise looks good.

PR #3638, opened 2021-06-08
Fix include locations for two occ commands

mmattel:
Fix the include location and files for two occ commands.
Overseen during a backport, 10.6 has the "old" directory structure, 10.7 onwards the new one.
No Backport, merges to 10.6 only!


EParzefall (approved):
LGTM

PR #3639, opened 2021-06-08
Improved script-guided installation.

EParzefall:
Clarified some information after testing the script-guided installation.

Backport to 10.7

### DIFF
--- a/modules/admin_manual/pages/installation/manual_installation/script_guided_install.adoc
+++ b/modules/admin_manual/pages/installation/manual_installation/script_guided_install.adoc
@@ -1,103 +1,99 @@
-= Scripts for the Script Guided Installation
+= Scripts for the Script-Guided Installation
 :toc: right
 :toclevels: 1
 :apache_docs: https://httpd.apache.org/docs/2.4/
 
 == Introduction
 
-WARNING: Use the scripts on your own responsibility. It may not work as expected if not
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/installation/manual_installation/script_guided_install.adoc
```diff
@@ -127,14 +123,13 @@ TIP: If you do not install or upgrade, the script sets ownership and permissions
   Used when installing or upgrading only. Enter the full path/filename to the tar source
   file, downloaded from owncloud.
 
-== Downloading
+== Creating the scripts
 
-Download both scripts to the same location and rename the script `instance.sh` according your
-needs.
+Copy & paste both scripts into an editor, adjust the instance.sh script and save them in the same directory. Name the script which we're calling `instance.sh` according to your needs so you can easily identify to which ownCloud installation it applies. However, the script owncloud_prep.sh is called with exactly that name by the instance script, therefore you must save it with exactly that name.
```
mmattel, line 128:
```suggestion
Copy & paste both scripts into an editor, adjust the instance.sh script and save them in the same directory. Name the script which we're calling `instance.sh` according to your needs so you can easily identify to which ownCloud installation it applies. However, the script `owncloud_prep.sh` is called with exactly that name by the instance script, therefore you must save it with exactly that name.
```

mmattel (approved):
LGTM

PR #3664, opened 2021-06-10
Correct X-XSS-Protection

mmattel:
References: https://github.com/owncloud/docs/issues/3660 (oC 10.7 X-XSS-Protection)

The X-XSS-Protection setting needed to be corrected to be inline with 10.7 core.

Backport to 10.7 only

Issue #3660, opened 2021-06-09
oC 10.7 X-XSS-Protection


voroyam:
## WHAT Needs to be Documented?
How you can configure apache to protect agains X-XXS

## WHERE Does This Need To Be Documented (Link)?

https://doc.owncloud.com/server/admin_manual/configuration/server/harden_server.html

## WHY Should This Change Be Made?
Because since 10.7 you get an error message documented in https://central.owncloud.org/t/oc-10-7-x-xss-protection/32051

mmattel:
Closing via #3664

mmattel:
Same issue: https://github.com/owncloud/docs/issues/3037

Issue #3037, opened 2021-01-16
Change X-XSS-Protection "1; block" -> "0"


ho4ho:
## WHAT Needs to be Documented?

- Change X-XSS-Protection "1; block" to "0"
- Why now "0" using

Reasons why:

https://github.com/OWASP/CheatSheetSeries/issues/376#issuecomment-602663932:

> I've been studying this issue for a while and it's *dangerous* to set this header. It makes sites that are not vulnerable to XSS, vulnerable.
> ...
> Developers need to either remove the header or set it to zero, I think the debate is over. There is NO good reason to set this header to blocking mode.

and:

https://github.com/helmetjs/helmet/issues/230#issuecomment-611904252
https://github.com/helmetjs/helmet/issues/230#issuecomment-611706715
https://github.com/helmetjs/helmet/issues/230#issuecomment-614106165

## WHERE Does This Need To Be Documented (Link)?

https://doc.owncloud.com/server/10.6/admin_manual/configuration/server/harden_server.html#serve-security-related-headers-by-the-web-server

## WHY Should This Change Be Made?

Explain why changed X-XSS-Protection "1; block" to "0"
https://github.com/owncloud/core/pull/38305
https://github.com/owncloud/core/issues/38236

## (Optional) What Type Of Content Change Is This? 
- [x] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

mmattel:
Closed via #3664

PR #3673, opened 2021-06-13
Kick block; from X-XSS-Protection

ho4ho:
Change to #3664 (also see #3037, #3660). If X-XSS-Protection disable not having `; mode=block` should be included.


mmattel (approved):

mmattel:
Backport to 10.7 only

PR #3688, opened 2021-06-15
Update server_prep_ubuntu_20.04.adoc

jnweiger:
The documentation should not refer to https://www.php.net/supported-versions.php
as "Currently supported PHP Versions". This is easily misunderstood as "currently supported by ownCloud".

They list 8.0, but we don't support 8.0

Backports needed, of course.

### DIFF
--- a/modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_20.04.adoc
+++ b/modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_20.04.adoc
@@ -22,7 +22,8 @@ version lower than 7.4, your PHP version is either already EOL or, for PHP 7.3, 
 in 2021. Consider upgrading to PHP 7.4. For more information see
 {php_versions_url}[Currently Supported PHP Versions]. This document considers two basic scenarios.
 You have/plan to upgrade from 18.04 LTS to 20.04 LTS or you have a fresh Ubuntu 20.04 LTS installation.
-Php 8.0 is not supported by ownCloud 10.7 server.
+
+IMPORTANT: PHP 8.0 is not supported by the ownCloud 10.x server series.
 
 NOTE: The commands and links provided in this description should give you proper hints but are
 without any responsibility.
--- a/modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_20.04.adoc
+++ b/modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_20.04.adoc
@@ -22,6 +22,7 @@ version lower than 7.4, your PHP version is either already EOL or, for PHP 7.3, 
 in 2021. Consider upgrading to PHP 7.4. For more information see
 {php_versions_url}[Currently Supported PHP Versions]. This document considers two basic scenarios.
 You have/plan to upgrade from 18.04 LTS to 20.04 LTS or you have a fresh Ubuntu 20.04 LTS installation.
+Php 8.0 is not supported by ownCloud 10.7 server.
 
 NOTE: The commands and links provided in this description should give you proper hints but are
 without any responsibility.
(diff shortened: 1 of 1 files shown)


modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_20.04.adoc
```diff
@@ -22,6 +22,7 @@ version lower than 7.4, your PHP version is either already EOL or, for PHP 7.3,
 in 2021. Consider upgrading to PHP 7.4. For more information see
 {php_versions_url}[Currently Supported PHP Versions]. This document considers two basic scenarios.
 You have/plan to upgrade from 18.04 LTS to 20.04 LTS or you have a fresh Ubuntu 20.04 LTS installation.
+Php 8.0 is not supported by ownCloud 10.7 server.
```
mmattel, line 25:
```suggestion

IMPORTANT: PHP 8.0 is not supported by ownCloud 10.7 server.
```

mmattel:
Backport to at least 10.7, needs checking if content exists for 10.6

modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_20.04.adoc
EParzefall: I assume we won't support php 8 in the next releases either.

EParzefall (approved):
Great catch.

modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_20.04.adoc
phil-davis: We don't support PHP 8 in ownCloud core 10.0 ... 10.7 and won't support it in the upcoming 10.8.

Support might come in the future - that is yet to happen, and when that happens we will need to look ffor these sort of statements and work out what words to rewrite.

modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_20.04.adoc
```diff
@@ -23,6 +23,8 @@ in 2021. Consider upgrading to PHP 7.4. For more information see
 {php_versions_url}[Currently Supported PHP Versions]. This document considers two basic scenarios.
 You have/plan to upgrade from 18.04 LTS to 20.04 LTS or you have a fresh Ubuntu 20.04 LTS installation.
 
+IMPORTANT: PHP 8.0 is not supported by the ownCloud 10.x server series.
```
phil-davis, line 26: To me, the words also imply (although they do not say directly) that PHP 8.0 will never be supported by the 10.x server series. I am not sure that the implication in my head is true. We might release `10.future` with PHP 8 support, or we might decide to number it 11.0 or? I just want to make sure that we do not mislead readers into thinking that PHP 8.0 support will never happen, or something.

Maybe a product owner/manager could comment about the words to use here? @pmaier1

jnweiger: I did not mean the entire owncloud 10 server series, when I said 10.7 :-)
I agree with Phil, we should never say never, (php8 support may come faster than expected). I cannot tell if it will be a server 11.x or 10.x when php8 support gets added.

EParzefall: We'll definitely not have it in the upcoming 10.8, correct? That's why I wasn't happy about only mentioning the lack of support for 10.7. How about a simple: PHP 8.0 is currently not supported by the ownCloud server.

phil-davis: That sounds good - "currently" is the key word!

EParzefall: @phil-davis Cool, but uhm, how do we fix it after my pre-mature commit here? Edit locally and make a new push?

phil-davis, line 26:
```suggestion
IMPORTANT: PHP 8.0 is not currently supported by the ownCloud server.
```

phil-davis: The easy way to adjust stuff is to do an online review, making a suggestion(s) and then GitHub gives you a chance to commit the suggestions. I did that just now.

phil-davis (approved):

modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_20.04.adoc
EParzefall: Thanks!

PR #3690, opened 2021-06-15
Update screenshots

hurradieweltgehtunter:
See https://github.com/owncloud/core/pull/38821


mmattel:
If Ci gets green, we have to check if the image sizes are the same because of png -> jpg via a local build. Just for security that nothing looks different (beside the content).

mmattel:
Note, no backporting as this is for 10.8 only

mmattel:
@hurradieweltgehtunter 
I ran a local build and I am not happy with the images - they are blurry...
This is well seeable and does not look good.
image
Beside that, I will take the chance to resize the images in the wizard a bit, to the same size we have in the user section (75%) after there are non-blurry images uploaded.

hurradieweltgehtunter:
I noted the very small resolution of the original images but did not change anything. 
I can reupload with bigger sizes

mmattel:
> I can reupload with bigger sizes

@hurradieweltgehtunter please do so, build resizing is not a problem. Just let me know when you are done and I will check 😄

hurradieweltgehtunter:
> > I can reupload with bigger sizes
> 
> @hurradieweltgehtunter please do so, image resizing for the build is not a problem. Just let me know when you are done and I will check 😄

Already done

mmattel (approved):

PR #3720, opened 2021-06-21
Add docs for 'tries to' in acceptance tests 

### LINKED ISSUE #3128: Document use of "tries to" in acceptance test steps
From discussion in issue https://github.com/owncloud/web/issues/4680

We use "tries to" in `When` steps of acceptance tests to:
- make it clear to the reader that the step is not expected to have a "normal" result
- to allow the underlying test code to be a bit different (for example, to remember that the action was not expected to succeed, or to not bother throwing an exception for the "error" that it will get...)

Also document what we do for scenarios where we want to "interrupt" the normal workflow of the UI, in order to take an unusual path. See issue comment https://github.com/owncloud/web/issues/4680#issuecomment-776445447

phil-davis:
Fixes #3128 

1) adjust some commas, minor crud
2) update some out-of-date information about skeleton files in tests
3) Fix test example for DEBUG_ACCEPTANCE_API_CALLS
4) Add docs for 'tries to' in acceptance tests

Backport not needed. This is intended to describe "real time" development against the latest branches of repos that have acceptance tests.

### DIFF
--- a/modules/developer_manual/pages/testing/acceptance-tests.adoc
+++ b/modules/developer_manual/pages/testing/acceptance-tests.adoc
@@ -387,9 +387,9 @@ Then the steps to execute for that scenario are listed.
 
 There are 3 types of test steps:
 
-- `Given` steps that get the system into the desired state to start the test (e.g. create users and groups, share some files)
-- `When` steps that perform the action under test (e.g. upload a file to a share)
-- `Then` steps that verify that the action was successful (e.g. check the HTTP status code, check that other users can access the uploaded file)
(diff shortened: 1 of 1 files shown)


mmattel (approved):

mmattel:
@phil-davis is there any backport needed? in case pls tell where to
(as a rule of thumb, if no bp is needed, just say no backport needed in the main comment 😄 )

phil-davis:
> @phil-davis is there any backport needed? in case pls tell where to
> (as a rule of thumb, if no bp is needed, just say no backport needed in the main comment smile )

I updated the main comment - no backport needed.

PR #3741, opened 2021-06-25
Prerequisites before deploying 10.8

mmattel:
This is a preperation for the upcoming 10.8 release (branch)

The landing page and some site-wide variables in site.yml need a preperation which can only be done close before a new server release is deployed. Some variables in site.yml are new and must only go into master and 10.7. When 10.8 docs is deployed the next days and with that 10.6 is frozen, some old variables can be deleted as they are not longer in use.

The docs internal related release notes has been removed as they are not of use for readers.
Do not get confused, the removed file is NOT the server release notes...

The 4 year old and unmaintained IOS doc has also been removed from the reference as the current minimum version is IOSv11+ which got available in September 2017. Post deploying 10.8, we can safely remove it from site.yml and the repo be archived.

The list of old server releases has been updated.

**STRICTLY** to master and backporting to 10.7 only,
where in the 10.7 site.yml some keys need to be reset to the former 10.7 state. An extra note will be made in the 10.7 backport so that we know that it is correctly implemented.

This PR can savely be merged, a build for master/10.7/10.6 runs error-free.


site.yml
```diff
@@ -48,11 +50,18 @@ asciidoc:
     idprefix: ''
     idseparator: '-'
     experimental: ''
+    # to be deleted after the 10.8 branch has been deployed
     latest-version: 10.7
     latest-download-version: 10.7.0
     previous-version: 10.6
     current-version: 10.7
-    page-version: 10.7
+    # to be deleted
```
phil-davis, line 58: Lots of references have been adjusted to use `latest-server-version` etc.
So why does this comment say "to be deleted"?

phil-davis (approved):

EParzefall (approved):
LGTM

site.yml
mmattel: Atm, in site.yml we have variables for the server release but the name itself does not reflect it. When looking to desktop which is built atm via this repo, we do now also have not only master but branch releases. To have the right branch used during the build for the desktop part, we need new variables. I took the chance to add new once for server which reflect the name so we know what it is good for. Imagine the future where the build base and server are in seperate repos. When 10.8 is created which is a clone of master, we have 2 identical branches  and 10.7 which is slightly different, 10.6 is archived and out. At that moment we can savely delete the „old“ variables as they are no longer be used in any branch, not in the definition (site.yml) nor in any documentation. 10.6 has too many dependencies on the old variables and is in addition too different for an easy backport. Therefore I have decided not to go that path. When this PR is merged, the proper BP to 10.7 is made and the new 10.8 branch is created, 10.6 is archived, we can safely remove the old varaibles from the 3 branches which I tried to mark 😎

phil-davis: Now I think I understand - this comment is the end of the "to be deleted" section.
I thought it indicated that the next line is "to be deleted"

PR #3742, opened 2021-06-28
Add Secure View sharing note

mrow4a:
includes note on limitations of secure view when sharing with multiple groups and users


mrow4a:
@pmaier1 is this limitation accepted or there should be fix for it implemented?

mmattel:
Q: does this needs backporting and if to which releases?

pmaier1:
> @pmaier1 is this limitation accepted or there should be fix/feature for it implemented?

As discussed, let's keep the limitation for now and document it.

mrow4a:
@mmattel just had a discussion with @pmaier1 and limitation is accepted until further request. Thus it would go into 10.X documentation until resolved in some release.

Seems we will not have such limitation in new roles/ocis architecture. What was agreed to maybe try to have share corner case resolved to enable sharing without validation failing in multi-group share scenario.

mmattel:
@phil-davis @EParzefall mind to do a language check?

phil-davis (approved):

PR #3744, opened 2021-06-29
Delete single quotes in docker run command

ChrisEdS:
Single quotes are not needed


mmattel (approved):
@ChrisEdS thanks for the PR.
Needs backporting to 10.7 and 10.6

PR #3753, opened 2021-06-30
Fix share expiration response description

phil-davis:
The response actually contains a  string like "expiration":"2021-06-30 00:00:00"

Note: the share expires at the end of the specified date, even though the response has a 00:00:000 timestamp in it. That is mentioned elsewhere in documentation.


mmattel (approved):

mmattel:
Does this need any backporting?

phil-davis:
> Does this need any backporting?

I will backport to 10.7 and 10.6 (I guess this has been wrong for a long time)

Issue #3767, opened 2021-07-05
Update PHP min version to 7.2.5


phil-davis:
## WHAT Needs to be Documented?
PR https://github.com/owncloud/core/pull/38934 and https://github.com/owncloud/core/pull/38891 changed the minimum PHP version for ownCloud 10.8 from 7.2.0 to 7.2.5. So the supported PHP  versions for 10.8 are 7.2.5 or greater up to 7.4 (latest patch release is always recommended!)

## (Optional) What Type Of Content Change Is This? 
- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [ ] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

PR #3768, opened 2021-07-05
Update PHP min version to 7.2.5

mmattel:
Fixes: https://github.com/owncloud/docs/issues/3767 (Update PHP min version to 7.2.5)

Core changed the minimum PHP version for ownCloud 10.8 from 7.2.0 to 7.2.5.

Backport to 10.8 necessary

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -75,6 +75,7 @@ asciidoc:
     php-supported-versions-url: https://secure.php.net/supported-versions.php
     http-status-codes-base-url: https://developer.mozilla.org/en-US/docs/Web/Status
     minimum-php-version: 7.2
+    minimum-php-printed: 7.2.5
     minimum-php-version-short-code: 72
     recommended-php-version: 7.4
     recommended-php-version-short-code: 74
@@ -82,7 +83,7 @@ asciidoc:
     std-port-memcache: 11211
     std-port-mysql: 3306
     std-port-redis: 6379
-    supported-php-versions: '7.2, 7.3, and 7.4'
+    supported-php-versions: '7.2.5+, 7.3, and 7.4'
   extensions:
     - ./lib/extensions/tabs.js
     - ./node_modules/asciidoctor-kroki/src/asciidoctor-kroki.js
--- a/modules/admin_manual/pages/configuration/server/caching_configuration.adoc
+++ b/modules/admin_manual/pages/configuration/server/caching_configuration.adoc
@@ -157,7 +157,7 @@ NOTE: The installer will automatically start `memcached` and configure it to lau
 [cols=",",options="header",]
 |===
 | PHP Version | Filename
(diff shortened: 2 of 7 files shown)


site.yml
```diff
@@ -74,15 +74,15 @@ asciidoc:
...
+    minimum-php-version: 7.2.5
+    minimum-php-version-short-code: 725
     recommended-php-version: 7.4
     recommended-php-version-short-code: 74
     std-port-http: 8080
     std-port-memcache: 11211
     std-port-mysql: 3306
     std-port-redis: 6379
-    supported-php-versions: '7.2, 7.3, and 7.4'
+    supported-php-versions: '7.2.5, 7.3, and 7.4'
```
phil-davis, line 85:
```suggestion
    supported-php-versions: '7.2.5+, 7.3, and 7.4'
```
How should we refer to this?
Any version 7.2.5 7.2.6...7.2.x should work, and we actually prefer that sites use the latest patch release.

site.yml
```diff
@@ -74,15 +74,15 @@ asciidoc:
     owncloud-changelog-url: https://owncloud.com/changelog/server/
     php-supported-versions-url: https://secure.php.net/supported-versions.php
     http-status-codes-base-url: https://developer.mozilla.org/en-US/docs/Web/Status
-    minimum-php-version: 7.2
-    minimum-php-version-short-code: 72
+    minimum-php-version: 7.2.5
```
phil-davis, line 77: This is used in some places where it generates a path to a file. For example: https://github.com/owncloud/docs/blob/master/modules/admin_manual/pages/configuration/server/caching_configuration.adoc

```
| PHP Version | Filename
| {minimum-php-version} | `/etc/php/{minimum-php-version}/mods-available/memcached.ini`
```
In the first column we might want to say `7.2.5` (or `7.2.5+` or?). In the 2nd column it needs to say only `7.2`

There might need to be more entries in `site.yml` that have different ways of expressing the version?

mmattel: Thanks for catching 👍 

There are two scenarious where `{minimum-php-version}` is used but in different ways.
1. as a print notice about the minimum level
2. as a variable for a path component

Having now a minimum patch release for PHP, my favorite is to keep the `{minimum-php-version}` variable and revert it to 7.2 and introduce a `{minimum-php-print}` variable having 7.2.5

Will take a closer look

phil-davis: Yes, `minimum-php-version` is really `minimum-php-minor-version` - we could insert `minor` in all those sort of names, but then the var name is so long that it looks like a German word :)

phil-davis (approved):

site.yml
mmattel: fixed, pls review again

phil-davis (approved):

PR #3793, opened 2021-07-07
add code snippet rendering

jnweiger:


jnweiger:
backport to 10.8 only.

mmattel:
This is the wrong location for the change !

You need to change this in `config.sample.php` in core which then gets transported with a `config-to-docs` run.

mmattel:
Closing, correct fix with the referenced PR´s

PR #3796, opened 2021-07-08
Replaced outdated Firebug URL

EParzefall:
Firebug link is more or less defunct, so we better point readers to the Mozilla and Chrome/Chromium dev tools.
Backports to 10.7 and 10.8 needed.


xoxys (approved):

TheRealBro (commented):
:+1:

PR #3829, opened 2021-07-12

jnweiger:
Struggling with te differet ways of doing federation.

In any case, I've never seen "(remote)", I only see "(guest)" or "(federated)" after an email address. I believe the latter is meant here.

image

In my screenshot I have all the normal sharing options, not only "(*) can edit" as suggested in the text... 
Please double check, if that needs an update (or if I am simply doing something else...)

### DIFF
--- a/modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
+++ b/modules/admin_manual/pages/configuration/files/federated_cloud_sharing_configuration.adoc
@@ -76,7 +76,7 @@ This requires no action by the user on the remote server; all it takes is a few 
 . Enable the Federation app.
 
 . Then, create a federated share by entering username@serveraddress in the sharing dialog (for example `freda@https://example.com/owncloud`). When ownCloud verifies the link,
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: there is historical confusion between the terms "remote share" and "federated share". Some user-visible places still use "remote", most use "federated". I raised issue https://github.com/owncloud/core/issues/38871 a few weeks ago.

core PR https://github.com/owncloud/core/pull/38877 is my first rough attempt at making core more consistent.

The API, capabilities stuff uses "remote" to advertise "federated shares". That is a client-visible thing that can't change. But that is not end-user-visible, so it is not such a big deal.

Internally the code uses variables names, constants etc with the string "remote". That is back-end code, and I guess it will stay like that.

The end-user-visible things in the UI (and maybe some help/explanations in `occ` commands?) need to be tidied up to be consistent.

phil-davis (approved):

phil-davis:
Backports:
10.8 #3834 
10.7 #3835

PR #3837, opened 2021-07-14
Update the generate-site.js script

mmattel:
After an intense discussion with @JanAckermann regarding ElasticSearch and a lot of testing, we identified a first improvement of the generator script used. We now more how to improve ES and have set a first change for. This PR streamlines the process a bit by removing an unecessary build variable in .drone.star and compacting the query if a search index has to be build. Nothing changes for the build-to-web process, but locally no index preperation will be done anymore improving build times for testing.

If a local environment is created and the necessary variables are set, it also will work locally and an index can be built. This eases the research how to improve the index creation a lot.

The change is not that big as it may look. The removal of the `if` statement needed to correct the indent of the code lines below.

Note, local build times are reduced by +1min !

Note, this change is also necessary for the docs-client-desktop repo (but is independent)

Backporting to 10.7 and 10.8

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -103,7 +103,6 @@ def build(ctx, latest_version, deployment_branch, base_branch, pdf_branch):
                 "image": "owncloudci/nodejs:14",
                 "environment": {
                     "BUILD_SEARCH_INDEX": ctx.build.branch == deployment_branch,
-                    "UPDATE_SEARCH_INDEX": ctx.build.branch == deployment_branch,
                     "ELASTICSEARCH_HOST": from_secret("elasticsearch_host"),
                     "ELASTICSEARCH_INDEX": from_secret("elasticsearch_index"),
                     "ELASTICSEARCH_PORT": from_secret("elasticsearch_port"),
--- a/generator/generate-site.js
+++ b/generator/generate-site.js
@@ -36,12 +36,13 @@ async function generateSite (args, env) {
 }
 
 async function generateIndex (playbook, pages) {
-  if ((process.env.BUILD_SEARCH_INDEX || 'true') !== 'true') {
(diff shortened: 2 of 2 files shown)


xoxys:
I don't understand the change related to the index build and update process... It was intended to have a way to build but **not** update the elastic index e.g. on a CI pull request run and build + update the elastic index on every master CI run. This process seems to be broken now, and that's where your build time "improvment" comes from. From what I can see index builds are not tested anymore.

grafik

Right now, I don't understand what is the real improvement of this change? What does it fix/improve/solve? Local testing was already possible before, but maybe I miss something.

xoxys:
As a comparison an old PR
grafik

JanAckermann:
@xoxys 
The code which was intended to 'generate search index' did nothing but parsing some data, without doing anything with it, so this slowed down the process - for no reason IHMO.

xoxys:
The generation steps do parsing to loop over all pages and collect and prepare data to build the index, right? https://github.com/owncloud/docs/blob/master/generator/generate-site.js#L44-L86 

Yes from my understanding it is useful to do it on every PR to avoid breaking parsing and index data generation due to content changes. No, I don't remember failures related to this, but I don't work on a daily basis. I don't understand the slowdown part:

- on localhost testings, you can set both variables to `false` to skip these steps
- on the CI please compare the screenshot, the PR CI finished in 5min45sec instead of 6min13sec thats an optimization of < 30sec for a pipeline that is bleeding fast already (you know the problems of e.g. the core pipelines)

If you can live with the current behavior, fine for me as well, I just wanted to mention that you may end up with failures on master branch commits/merges due to an untested part in the generator script that could have been verified in a PR just to save less than 30sec in testing.

mmattel:
It was not intended that CI does not trigger a no-index... will check.

PR #3844, opened 2021-07-16
Update test-pilots.adoc

jnweiger:
as owncloud.com and ownclud.org no longer differ, I'd suggest to only use one form (owncloud.com) for all references.


mmattel (approved):

mmattel:
Backport to 10.8 and 10.7

PR #3915, opened 2021-08-02
Remove the need for perl from the backport script

mmattel:
After a lot of testing (kudos to @EParzefall), we identified how to proper remove the need for perl in the backport script and replace it with bash commands. Not knowing why, but the use of perl caused issues - this is now eliminated.

Backport to 10.8 and 10.7


EParzefall (approved):
Works fine! :-)

PR #3922, opened 2021-08-04
apache 2.4: updated redirect rule

thommierother:


mmattel:
@thommierother thanks for the PR. Mind to add a description what the change is about?

thommierother:
mod_rewrite and the "rewrite" rules operate on the Server-side. mod_alias and "redirect" simply asks the client to refetch with the new URI.  mod_rewrite is recommended as "last resort" but not needed for such a simple case. See https://httpd.apache.org/docs/trunk/rewrite/avoid.html and https://httpd.apache.org/docs/trunk/mod/mod_alias.html#redirect for reference

mmattel:
I cant make a suggestable commit, so you need to do, but on line 111 you also need to fix the text from `RewriteRule` -- `Redirect`.

mmattel:
Huuu...
I was just reading the whole document especially the content below the changes, which reads the following. The keywords are `not to use redirect rules...violate the OpenID Connect specification`:

Here is the complete statement:

**TIP:** `Depending on the respective infrastructure setup there can be other ways to solve this. In any case, please make sure not to use redirect rules as this will violate the OpenID Connect specification.`

If your change is correct, please prove not violating the secs.

thommierother:
Interesting. I can not find any info that the OIDC spec **forbids** usage of "redirect 301" instead of a rewrite rule on the web-server side and in the context of the clients service discovery information (./well-known ... URI) . We use "redirect 301" in the OIDC configs at our customers infrastructure too. Without issues, but with another client software (java-based).

mmattel:
> I can not find any info that the OIDC spec forbids usage of "redirect 301" instead of a rewrite rule on the web-server side and in the context of the clients service discovery information (./well-known ... URI).

Will have a look asap, pls be patient...

thommierother:
I know that the redirect is not recommended in cases where we already have a security context with personal data/token data. But in this case it is only the first service discovery - this shouldnt be a problem ... Maybe @DeepDiver1975 knows more ...

DeepDiver1975:
Our clients are not accepting a 301 due to the mentioned security impact. Maybe not mandatory by the specs - but by us ;-)

thommierother:
thanks for clarification, in this case my PR is obsolete

thommierother:
closed, as the change would violate OC specs

PR #3959, opened 2021-08-16
Add warning around removal of external storage configuration

mrow4a:
Add a note to enter proper storage config as config changes create a new storage ID making existing database entries orphans increasing overhead. Manual deletion is necessary.

Backport to 10.8 and 10.7

### DIFF
--- a/modules/admin_manual/pages/configuration/files/external_storage/configuration.adoc
+++ b/modules/admin_manual/pages/configuration/files/external_storage/configuration.adoc
@@ -24,6 +24,8 @@ image:configuration/files/external_storage/enable-app.png[Enabling External Stor
 
 == Storage Configuration
 
+IMPORTANT: Before adding a storage in production environment make sure its configuration is correct. Removal of the external storage or change of its configuration does not remove metadata entries from database belonging to previous storage configuration.
+
 To create a new external storage mount, select an available backend from
 the dropdown *Add storage*. Each backend has different required options,
 which are configured in the configuration fields.
@@ -127,3 +129,7 @@ Alternatively, replace `–all` with the user name to trigger a rescan of the us
 for example every 15 minutes, which includes the mounted external storage.
 
 TIP: See xref:configuration/server/occ_command.adoc#the-filesscan-command[the occ’s file operations] for more information.
+
+== Known limitations
+
+- Removal of the external storage or change of its configuration does not remove metadata entries belonging to previous storage configuration. This may impact performance of the installation as previous configuration metadata entries get orphaned. Removal of orphaned entries requires manual deletion of orhpaned storage cache by its storage id. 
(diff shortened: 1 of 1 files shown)


mrow4a:
@phil-davis do you know who could help pushing forward this documentation?

modules/admin_manual/pages/configuration/files/external_storage/configuration.adoc
```diff
@@ -127,3 +129,7 @@ Alternatively, replace `–all` with the user name to trigger a rescan of the us
 for example every 15 minutes, which includes the mounted external storage.
 
 TIP: See xref:configuration/server/occ_command.adoc#the-filesscan-command[the occ’s file operations] for more information.
+
+== Known limitations
+
+- Removal of the external storage or change of its configuration does not remove metadata entries belonging to previous storage configuration. This may impact performance of the installation as previous configuration metadata entries get orphaned. Removal of orphaned entries requires manual deletion of orhpaned storage cache by its storage id.
```
phil-davis, line 135:
```suggestion
- Removal of the external storage or change of its configuration does not remove metadata entries belonging to the previous storage configuration. This may impact performance of the installation as previous configuration metadata entries get orphaned. Removal of orphaned entries requires manual deletion of orphaned storage cache by its storage id. 
```

modules/admin_manual/pages/configuration/files/external_storage/configuration.adoc
```diff
@@ -24,6 +24,8 @@ image:configuration/files/external_storage/enable-app.png[Enabling External Stor
 
 == Storage Configuration
 
+IMPORTANT: Before adding a storage in production environment make sure its configuration is correct. Removal of the external storage or change of its configuration does not remove metadata entries from database belonging to previous storage configuration.
```
phil-davis, line 27:
```suggestion
IMPORTANT: Before adding a storage in a production environment make sure its configuration is correct. Removal of the external storage or change of its configuration does not remove metadata entries from the database belonging to the previous storage configuration.
```

phil-davis (commented):
I am not really familiar with the UI for this. s it easy to fill in all the needed settings and then "OK" right at the end after checking?

mrow4a:
@phil-davis I think the problem here is you do not need to click any OK for it :) Thus before adding some config to prod it should be very thoroughly architected and validated.

phil-davis:
> @phil-davis I think the problem here is you do not need to click any OK for it :) Thus before adding some config to prod it should be very thoroughly architected and validated.

That makes it difficult for admins, if the data entry fields keep auto-saving as they fill in the data.

mrow4a:
@phil-davis I think you can configure it also outside of UI (and probably what one should do when initializing production setup)

https://doc.owncloud.org/server/admin_manual/configuration/files/external_storage/configuration.html#introduction
https://doc.owncloud.org/server/admin_manual/configuration/server/occ_command.html#files-external

mmattel (approved):

PR #4002, opened 2021-09-07
fix: add favicons as static files

xoxys:
Favicons were manually copied to the web server. This way, this way, Antorra will include these files in the final build as well. 

Needs backport


EParzefall (approved):
LGTM

PR #4037, opened 2021-09-15
backtick removed from wnd examples.

jnweiger:
Cannot work with backticks this was illegal shell syntax.
As a general rule, backticks should never appear anywhere in shell scripts. When command interpolation is actually intended, `$()` should be used instead of backticks.

Backports optional.


mmattel:
Thanks for catching !

Backporting for 10.8 mandatory, needs to have a closer look regarding 10.7

mmattel (approved):

Issue #4032, opened 2021-09-13
Assign defined pages a noindex attribute


mmattel:
## WHY Should This Change Be Made?

When creating a page, this page can be a frame only which includes several sub pages to be built as complete page. As example see the occ commands page. Antora builds all pages. It builds not only the main page with all the includes, but also the individual pages. These individual pages can be accessed manually if you know the url for - but this is senseless. 

Crawlers like Google also index these sub pages and makes them availabe for searchers - which should be prevented.

## Targeted Solution
Particular include pages should have the ability to be tagged with a noindex attribute directly in the .adoc file preventing to be indexed by crawlers.

@xoxys fyi

mmattel:
I already have found possible ways to address this issue but need to do some testing and have a in depth dive into.

See: https://docs.antora.org/antora/2.3/page/attributes/#page-attributes
which could be used via our docs-ui template with a conditional rendering

xoxys:
Agree, feel free to implement it.

PR #4038, opened 2021-09-15
Make all included occ files not being indexed by crawlers

mmattel:
This is part of:
https://github.com/owncloud/docs-ui/pull/332 (Make a conditional noindex meta directive)
and fixes: https://github.com/owncloud/docs/issues/4032 (Assign defined pages a noindex attribute)

**Background**
In Antora, you can create custom page attributes, see:
https://docs.antora.org/antora/2.3/page/page-attributes/#custom-attribute

These page attributes can be set in `site.yml` (global), in `antora.yml` (for a particular version) or on a `page` level. Compared to other attributes, these are available to the UI which can be used there to be either printed or for conditional html code.

The latter is being used to add a conditional meta tag `noindex,nofollow` for robots.

When in any of our documentations according the description above the attribute `page-noindex` is set (it just needs to be not empty), the page rendered will get the meta tag set.

This is necessary eg for the occ commands which are generated by including single files. Antora not only renders the final page, but also all including pages, even they are not linked anywhere. Crawlers find and provide them to searchers. With this mechanism, we instruct crawlers not to do so for marked pages returning only real content.

**Relevant Pages Changed**
This PR adds the `page-noindex` attribute to each document which is included by a include directive which is not referencing a file from the `_partials` directory. This affects all occ command parts in the admin guide and some files in the developer guide. 

Beside that, I fixed a double included file in the occ command set...

Backporting to 10.8 and 10.7

@xoxys fyi

### DIFF
--- a/modules/admin_manual/pages/configuration/server/occ_commands/app_commands/2fa_commands.adoc
+++ b/modules/admin_manual/pages/configuration/server/occ_commands/app_commands/2fa_commands.adoc
@@ -1,4 +1,5 @@
 = Two-factor Authentication
+:page-noindex: yes
 
 Marketplace URL: {oc-marketplace-url}/apps/twofactor_totp[2-Factor Authentication]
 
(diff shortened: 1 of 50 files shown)


EParzefall (approved):
LGTM

PR #4061, opened 2021-09-22
Update of the makepdf script and configuration to be inline with the other repos

mmattel:
Just a small update, it handles the case that the first letter is not capital (see iOS)

Backport to 10.7 and 10.8


phil-davis (approved):

EParzefall (approved):
LGTM

PR #4075, opened 2021-09-24
Specify minor versions of MariaDb that are supported

phil-davis:
Related to issue https://github.com/owncloud/core/issues/39283

MariaDb 10.6 stable was recently released. There seem to be some different settings that are needed (see the issue).

For now, we could adjust the docs so that the actual supported MariaDb minor versions are mentioned.

https://downloads.mariadb.org/mariadb/+releases/ - patch releases are coming out regularly for 10.2 10.3 10.4 10.5 - so that is why I put those in the list. And full CI passes with each of those versions.

### DIFF
--- a/modules/admin_manual/pages/installation/system_requirements.adoc
+++ b/modules/admin_manual/pages/installation/system_requirements.adoc
@@ -2,7 +2,7 @@
 :toc: right
 :toclevels: 1
 :php-intl-ext-url: http://php.net/manual/en/intro.intl.php
-:ppa-guide-url: https://itsfoss.com/ppa-guide/ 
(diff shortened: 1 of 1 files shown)


mmattel (approved):

mmattel:
Are there backports needed?

phil-davis:
Yes, this applies to all oC10 versions. So backports are needed to 10.8 and 10.7 docs.

I  would like to get an opinion from @pmaier1 as PM agreeing to what we write here. It's important that we correctly document versions that work, and that we support.

IljaN (approved):

pmaier1 (approved):
Thanks for taking care!

Issue #4078, opened 2021-09-25
Syntax error in example config for adding preview providers.


dephekt:
## WHAT Needs to be Documented?
A syntax error needs corrected in the example configuration for adding additional preview providers to ownCloud.

## WHERE Does This Need To Be Documented (Link)?
https://doc.owncloud.com/server/10.8/admin_manual/configuration/files/previews_configuration.html#adding-a-preview-provider

It would need backported from master to 10.8 and 10.7.

## WHY Should This Change Be Made?
If someone directly copy and pastes the example into their ownCloud config.php file, it will crash the application due to the syntax error preventing parsing of the configuration. A comma is missing at the end of the MarkDown provider.

## (Optional) What Type Of Content Change Is This? 
- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [x] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

dephekt:
If used as is, the error in the crash log is this:

```json
{
    "date": "2021-09-25T14:00:01+00:00",
    "parentId": null,
    "id": "689c12b3c7f196fdb431e344d98fc4e7614f2b61559fd4.88038859",
    "class": "ParseError",
    "message": "syntax error, unexpected ''OC\\Preview\\MP3'' (T_CONSTANT_ENCAPSED_STRING), expecting ']'",
    "stacktrace": [
        {
            "file": "\/var\/www\/owncloud\/lib\/private\/Config.php",
            "line": 64,
            "function": "readData",
            "class": "OC\\Config"
        },
        {
            "file": "\/var\/www\/owncloud\/lib\/base.php",
            "line": 135,
            "function": "__construct",
            "class": "OC\\Config"
        },
        {
            "file": "\/var\/www\/owncloud\/lib\/base.php",
            "line": 530,
            "function": "initPaths",
            "class": "OC"
        },
        {
            "file": "\/var\/www\/owncloud\/lib\/base.php",
            "line": 1070,
            "function": "init",
            "class": "OC"
        },
        {
            "file": "\/var\/www\/owncloud\/console.php",
            "line": 74,
            "function": "require_once"
        },
        {
            "file": "\/var\/www\/owncloud\/occ",
            "line": 11,
[...]

PR #4079, opened 2021-09-25
Fix syntax error in preview provider example.

dephekt:
Reference: #4078 

Fixes a small issue with a missing comma which, if the example is copy and pasted, will cause ownCloud to crash until the issue is tracked down and fixed.

Needs backported to 10.8 and 10.7.


phil-davis (approved):

phil-davis:
@dephekt thanks!

Can you make backports?

dephekt:
> @dephekt thanks!
> 
> Can you make backports?

Sure, added backports. For some reason the backporting script was messing up the pull title and showing just `ul` in the commit message. I locally amended the bad messages and force pushed them so everything has a consistent title from the original MR/commit.

phil-davis:
@dephekt thanks!

(I make backports with "regular" `git` commands, so I am not familiar with that script)

PR #4083, opened 2021-09-27
Fix client config location for OIDC

mmattel:
References: https://github.com/owncloud/docs/issues/3498 ([OIDC] Add Client ID, Client Secret, Redirect URIs to general OIDC docs)

This PR changes the following:

* Transform the Client info into a table for better readability
* Fix the Desktop content according the referenced issue
* Relocate this info into the main OIDC document
and set a link to this new section from the Kopano example page
* Added a General Example section
* Fixing some small things

@michaelstingl

Backport to 10.8 only

PR #4092, opened 2021-09-29
add documentation for oauth2 trusted clients

C0rby:
Some people were rightly asking for documentation about the OAuth2 app's trusted clients feature. So here it is.


modules/admin_manual/pages/configuration/server/security/oauth2.adoc
```diff
@@ -96,6 +96,19 @@ To install the application, place the content of the OAuth2 app inside your inst
 
 To enable token-only based app or client logins in `config/config.php` set `token_auth_enforced` to `true`.
 
+==== Trusting Clients
+
+Since version 0.5.0 of the OAuth2 app, you can mark clients as trusted. This will have the effect that the consent step in the authentication process will be skipped for this client.
+
+image:configuration/server/security/oauth2_configuration.png[image]
+
+If you want to mark an existing client as trusted, you have to copy the `Client Identifier (ID)` and the `Client Secret`, then delete the existing entry and finally add it again via the occ comand. Follow this link regarding xref:configuration/user/oidc/oidc.adoc[ownCloud client config details].
```
mmattel, line 105: Questions:
* **How** did these entries find their way into that screen?
* Copy some content as you can see in the screenshot - **ok**
* **Delete content** - from where, in the gui, per occ, what is the process in details, where to click or which command
*  Readding as trusted per command - **ok**

modules/admin_manual/pages/configuration/server/security/oauth2.adoc
```diff
@@ -96,6 +96,19 @@ To install the application, place the content of the OAuth2 app inside your inst
...
+
+[source,console,subs="attributes+"]
+----
+{occ-command-example-prefix} oauth2:add-client <name> <client-id> <client-secret> <redirect-url> [<allow-sub-domains> [<trusted>]
+----
```
mmattel, line 110: I fixed how an occ command is added in docs **BUT**:

We do not have any entry in our occ documentation about an oauth2 command set ?!

mmattel:
@C0rby could you redo the screenshot but without the `owncloudweb` item.
See the referenced pr #4083 where in the given link the cofiguration of `owncloudweb` is with respect to the redirect uri different to the client ones not static. Having this, an image with static and admin created redirect uri examples will confuse readers - they will for sure try the given owncloudweb link of the screenshot.

modules/admin_manual/pages/configuration/server/security/oauth2.adoc
C0rby: > How did these entries find their way into that screen?

Desktop, Android and iOS are there per default, the owncloudweb entry was added by the admin. Is that relevant for this section of the documentation?

> Delete content - from where, in the gui, per occ, what is the process in details, where to click or which command

Good point. I just noticed a bug in the UI because there should be trashbin buttons on the right side but they are outside of the viewport. You can see them when you resize the webui to 90%.

mmattel:
Will do the review tomorrow

PR #4138, opened 2021-10-13

voroyam:
fixed wrong occ system:cron command from 

occ system cron
to
occ system:cron


voroyam:
Thanks for the back ports

Issue #4127, opened 2021-10-11
Run `config-to-docs` after PR #has been merged


JanAckermann:
## WHAT Needs to be Documented?

PR https://github.com/owncloud/core/pull/38583 introduces a new config setting `dav.propfind.depth_infinity` so we might want to run config-to-docs after this PR has been merged

PR #4141, opened 2021-10-14
config-to-docs run

mmattel:
Fixes: https://github.com/owncloud/docs/issues/4127 (Run config-to-docs after PR #has been merged)

CTD run because of a new confing in core: https://github.com/owncloud/core/pull/38583 ([full-ci] Enable streaming for propfind)

NO Backport, master only

@michaelstingl should there be an info be made in the docs-client repos?

@JanAckermann @DeepDiver1975 @JammingBen  FYI


michaelstingl:
> @michaelstingl should there be an info be made in the docs-client repos?

I don't think so.

PR #4187, opened 2021-11-01
Fix accidentially added occ richdocuments at 10.7

mmattel:
Richdocuments was accidentially added in the occ commands at 10.7 with the last backport.
This is now fixed. In addition, the list of page-aliases is now not one single big line but line by line for better readability. This is based on the input
https://antora.zulipchat.com/#narrow/stream/282400-users/topic/Feature.20Request.20.3Apage-aliases.3A

Will do this for occ for master and 10.8 too.

No backport, 10.7 only


EParzefall (approved):
LGTM

PR #4188, opened 2021-11-01
Make page-aliases in occ line by line

mmattel:
This PR does not change the result when building, but makes the readbility of the page-aliases in the document much better having them line-by-line. This is based on the input
https://antora.zulipchat.com/#narrow/stream/282400-users/topic/Feature.20Request.20.3Apage-aliases.3A

Backport to 10.8 only
Docs PR https://github.com/owncloud/docs/pull/4187 holds the 10.7 change for the same + a fix.


EParzefall (approved):
LGTM

PR #4189, opened 2021-11-01
Make page-aliases better readable when having multiple elements

mmattel:
No content change, just for better readbility and to avoid errors/typos.
Make page-aliases better readable when having multiple elements to be aliased.

Backport to 10.8 and 10.7


EParzefall (approved):
LGTM

PR #4247, opened 2021-11-09
Fixing external links returning 404

mmattel:
This PR fixes (hopefully all) external links returning a 404. This also includes example domains which are now printed but not presented as link. Depending on the checker, there maybe more to be found.

Note that dead links which do not have a replacement URL have been removed.
Some Microsoft URL´s which have been redirected are now corrected.
Texts have been adopted if needed in case a URL has changed or been removed.

Backport to 10.8 and 10.7


EParzefall (approved):
LGTM

PR #4274, opened 2021-11-15
s3_compatible_object_storage_as_primary.adoc use marketplace with /apps/

jnweiger:
Seems the {oc-marketplace-url} does not have the trailing /apps/ path component. 
Without that, we get 404 errors.
@mmattel should the /apps/ be inside {oc-marketplace-url} or appended, as per this PR?

See https://doc.owncloud.com/server/next/admin_manual/configuration/files/external_storage/s3_compatible_object_storage_as_primary.html

backports to 10.9  please (at least) :-)


mmattel (approved):

mmattel:
> should the /apps/ be inside {oc-marketplace-url} or appended, as per this PR

Per PR, we keep the plain URL in the attribute and anything else will be added when used

PR #4288, opened 2021-11-17
consistent use of semicolon

jnweiger:
ceph and scality example already had a trailing semicolon, 
the amazon aws example and the ntroductory snippet dont.

I've tested the use of comma: PHP assignements require a semicolon at the end. A comma is a syntax error.
When using the snippets in a separate file `s3.config.php` the trailing comma leads to a blank page, and no error logged in `owncloud.log`.
With a trailing semicolon, everythnig works fine.

Other examples with trailing semicolon:  https://github.com/owncloud/files_primary_s3/tree/master/tests/drone


phil-davis (approved):
This is a correct fix to the syntax.

Note: most users "should not notice the problem" (famous last words) because they will already have a `config.php` that starts with:
`$CONFIG = [`
and ends with:
`];`

Or even in the format (like mine):
`$CONFIG = array (`
ending
`);`

and they are pasting the 'objectstore' block of code inside the existing config.

phil-davis:
@mmattel please make whatever backports are needed, and I can approve.

PR #4305, opened 2021-11-18
Update s3_compatible_object_storage_as_primary.adoc

jnweiger:
We mention encryption in two of the examples. But we also declare primary_s3 incompatible with encryption.
Is this a different kind of encryption, that is actually supported? If so, please clarify the language and disregard this PR.

### DIFF
--- a/modules/admin_manual/pages/configuration/files/external_storage/s3_compatible_object_storage_as_primary.adoc
+++ b/modules/admin_manual/pages/configuration/files/external_storage/s3_compatible_object_storage_as_primary.adoc
@@ -77,8 +77,6 @@ $CONFIG = [
         'arguments' => [
             // replace with your bucket
             'bucket' => 'owncloud',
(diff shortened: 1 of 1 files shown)


mmattel:
> Is this a different kind of encryption, that is actually supported?

I `guess` that this is a parameter that is handed over to the s3 storage. I can´t remember that we have such a parameter but I may be wrong. If this is correct, we should improve the note about the s3 incompatibility of encryption inside owncloud. Do you know someone we can ask?

mmattel:
Got an answer - S3 has the ability to use built-in encryption defined in the settings. Beside "...incompatible with any app that uses direct file I/O...", it makesno sense to double encrypt - use the bucket built in encryption. I will highlight this topic a bit in a new PR making this more clear.

Closing if you agree

mmattel:
Closing in favour of #4330

PR #4309, opened 2021-11-20
Changes necessary for 10.9 branch

mmattel:
This are the changes necessary to finalize the creation of the 10.9 branch.

The 10.9 branch is already pushed and prepared and is included in the branch protection rules.

When 10.9 (core) is finally out, the 10.7 branch can be archived,
see step 3 in  https://github.com/owncloud/docs/blob/master/docs/new-version-branch.md

Note, that the 10.9 branch in this repo is already created, but the `latest` pointer on the web will be set to it automatically when the tag in core set. This means, that in the docs homepage, `latest` will point to 10.8 until the tag in core is set accordingly. When merging this PR, 10.7 will be dropped from the web but is available via pdf as usual.

Note, this PR must be merged before the 10.9 tag in core is set to avoid a 404 for `latest`.

Before merging this PR, we should take care that 10.7 has all changes necessary merged as post merging the 10.7 pdf is fixed.

@pmaier1 @jnweiger fyi

@EParzefall @phil-davis post merging this, we need to backport all relevant changes to 10.9


EParzefall (approved):
LGTM

phil-davis (approved):
LGTM.
Before "dropping" 10.7 docs, it will be good to get all the links and... working perfectly, so that the 10.7 PDF will have good working structure and links.

mmattel:
> Before "dropping" 10.7 docs, it will be good to get all the links and... working perfectly, so that the 10.7 PDF will have good working structure and links.

That has been done already, but let us know if you have something in your mind that needs to be fixed. 10.7 will exist in parallel to 10.9 (after this merge) until 10.9 (core) gets published - means we have about 2 weeks to fix open stuff.

PR #4381, opened 2021-11-28
[10.8] Fix Admin Username / Password Typo

NoiSek:
Fixes a typo in the documentation that would prevent the user from ever being able to set a default admin username / password.


mmattel (approved):

mmattel:
Thanks for catching 👍

Issue #4392, opened 2021-11-29
Activity rename- and move actions


JammingBen:
## WHAT Needs to be Documented?

The next activity app version (2.7.0) will support activities for rename- and move actions. However, this feature is disabled by default and has to be enabled via config: `occ config:app:set activity enable_move_and_rename_activities --value yes`

see https://github.com/owncloud/activity/pull/1018 & https://github.com/owncloud/activity/pull/1031

## WHERE Does This Need To Be Documented (Link)?

https://doc.owncloud.com/server/next/admin_manual/configuration/server/activity_configuration.html

## (Optional) What Type Of Content Change Is This? 
- [x] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

PR #4404, opened 2021-11-30
New occ Activity commands

mmattel:
Fixes: https://github.com/owncloud/docs/issues/4392 (Activity rename- and move actions)

New occ commands for the activity app.
Because the change is app version and not core dependent, we can backport.

Backport to 10.9, 10.8 and 10.7


EParzefall (approved):
LGTM

PR #4420, opened 2021-11-30
add complete tar ball

voroyam:
since we have the complete community tar bar we should use it. then you just need to enable the apps instead of installing them.


mmattel (approved):
In general, I am not happy using a fixed version in the documentation which gets outdated with the next release. This is due to the naming of the tar versions made (yes, I am not happy about it, did I say that already? 😅 ). But this is ok for the quick guide.

Needs Backporting to 10.9, 10.8 and 10.7

voroyam:
We would need to update this link with every new release, but it's a win for users and customers because they already will have all the apps available to them (in my view that's a win).

PR #4445, opened 2021-12-02
remove hint about 2FA in notifications API dev doc

pmaier1:
Not sure why it says that you'd need 2FA for the API. This is nonsense. Removing it.


IljaN (approved):

modules/developer_manual/pages/core/apis/ocs-notification-endpoint-v1.adoc
```diff
@@ -4,10 +4,6 @@
 :request-base-path: ocs/v2.php
 :2fa-app-url: https://github.com/owncloud/twofactor_totp
```
phil-davis, line 5: IMO this line can also be removed.

phil-davis:
@mmattel this could be backported to all maintained doc versions.

PR #4502, opened 2021-12-14
[10.8] [PR 4500] Move branch specific attributes from global site.yml to branch relevant antora.yml

mmattel:
Backport of PR #4500

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -8,3 +8,9 @@ nav:
 
 asciidoc:
   attributes:
+    minimum-php-version: 7.3
+    minimum-php-printed: 7.3.0
+    minimum-php-version-short-code: 73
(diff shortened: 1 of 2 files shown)


antora.yml
```diff
@@ -8,3 +8,9 @@ nav:
 
 asciidoc:
   attributes:
+    minimum-php-version: 7.3
```
phil-davis, line 11: 10.8 supports PHP 7.2 - I guess you will do a separate PR to update the specific values?

phil-davis (approved):

antora.yml
mmattel: yes, this PR should just move the attribues as they are.
the next PR is 10.8 (resp 10.7 only) and contains the correct values

PR #4515, opened 2021-12-16
Add PHP extensions to various docs

### LINKED ISSUE #2513: Lists of PHP extensions are different in 2 docs
The lists of PHP extensions are different in:
https://github.com/owncloud/docs/blob/master/modules/admin_manual/pages/installation/ubuntu_18_04.adoc
https://github.com/owncloud/docs/blob/master/modules/admin_manual/pages/installation/server_prep_ubuntu_18.04.adoc

Only in server_prep_ubuntu_18.04 list:
php-bcmath
php-bz2
php-gettext
php-gmp
php-igbinary
php-seclib
(6 items)

Only in ubuntu_18_04 list:
php-common
php-json
php-xml
(3 items)

In both:
php-apcu
php-curl
php-imagick
php-gd
php-imap
php-intl
php-mbstring
php-mysql
php-redis
php-ssh2
php-zip
(11 items)

Optional in ubuntu_18_04 list:
php-smbclient
php-ldap

Sort out why, and get the lists correct and consistent.

phil-davis:
Fixes #2513 

There are Ubuntu install docs in `manual_installation` and `quick_guides`. They have different subsets of PHP extensions listed. (I was hoping that the quick guide would just be a subset of those in the "full" installation)

I have added ones that I think are obviously missing from one or other, to get the lists more in sync. Maybe some are auto-installed anyway, and that is why this has not been a problem? But it seems good to have both lists as similar as reasonable.

The differences are now:

Extra in the quick guides: `php-apcu php-libsmbclient php-smbclient` (the quick guides have separate instructions for the smb stuff)

Extra in the manual installation guides: `php-igbinary php-gmp php-bcmath php-bz2` (I am not sure why these are not needed/useful to have for "quick install".

### DIFF
--- a/modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_18.04.adoc
+++ b/modules/admin_manual/pages/installation/manual_installation/server_prep_ubuntu_18.04.adoc
@@ -160,7 +160,7 @@ sudo apt install rsync
 # PHP extensions needed to use ownCloud
 sudo apt install php-mysql php-mbstring php-gettext php-intl php-redis \
      php-imagick php-igbinary php-gmp php-bcmath php-curl php-gd php-zip \
-     php-imap php-ldap php-bz2 php-ssh2 php-phpseclib
+     php-imap php-ldap php-bz2 php-ssh2 php-phpseclib php-common php-json php-xml
 ----
 
 == Useful Commands For Managing PHP Extensions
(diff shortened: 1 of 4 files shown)


mmattel (approved):

phil-davis:
I will backport to 10.9 and 10.8

mmattel:
php-apcu is a bit faster than redis, but you need redis for transactional file locking and for custered environments. The performance difference is not that big especially when using sockets on single servers and for professional environments redis the product of choice anyways. Therefore I did not add php-apcu to the full install guide and stayed focused on redis.
the math stuff is also good the have in a full install guide because when not using a hsm, the gain is measurable because of the libraries used. if someone uses a quick install, he is for sure not in professional environments but will recieve a working miminal standard enironment for using ownCloud. that is the old tension between quick vs full...

PR #4526, opened 2021-12-21
Use the current complete tarball to install 10.9.0

phil-davis:
some guides have the current complete timestamp. Should this auto-update via variable?

IMO this is what @jnweiger is intending to do in PR #39608

Backport to 10.9


mmattel (approved):

mmattel:
This is a good fix.
For a long term, I will add an attribute in antora.yml (branch specific) which will help maintaining this more easily.

PR #4541, opened 2021-12-27
Fix a pdf build issue

mmattel:
By chance looking into the detailed CI result regarding the pdf build (green!), I encountered an issue:

```
asciidoctor: ERROR: failed to parse formatted text: ​<a href="https://github.com/issues?q=is%3Aissue+user%3Aowncloud+is%3Aopen+label%3A"Needs+info"+sort%3Acreated-asc">Bugs which need info</a>
```
This is because the url contains double quotes which cant be rendered (but used...)

The fix is to utf8-encode the double quotes and manually replace them with %22.

This PR further adds a description about this process and fixes/adds some texts in the repo documentation.

Backport to 10.9 and 10.8

### DIFF
--- a/docs/build-the-docs.md
+++ b/docs/build-the-docs.md
@@ -393,5 +393,5 @@ In case CI needs to be restarted, which can happen in the rare case it was not t
 
 ```
 git commit --allow-empty -m "restart ci"
-gip push -f
+git push
 ```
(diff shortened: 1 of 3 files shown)


docs/build-the-docs.md
```diff
@@ -388,3 +386,12 @@ to delete the `cache` directory with the command below (use the error path print
...
 rm -r /var/owncloud/docs/cache
 ```
+
+### Manually Restarting CI
+
+In case CI needs to be restarted, which can happen in the rare case it was not triggered post pushing automatically, you need to manually (re)start the CI. This can be done by creating an empty commit and pushing it. To do so, change to the branch in question and follow the git commands below:
+
+```
+git commit --allow-empty -m "restart ci"
+gip push -f
```
phil-davis, line 396:
```suggestion
git push
```

If just adding a commit, like this, then `git push` should be enough. The local branch will be cleanly 1 commit ahead of the upstream at GitHub.

tbsbdr (approved):
not tested but looks legit

PR #4554, opened 2021-12-29
Global Attribute Cleanup

mmattel:
Changes in global attributes used/new attributes created.

* Some url specific attributes used in site.yml did not have a unified naming convention.
* Introduction of a new global attribute for the ownCloud complete bundle.
This makes it easier to update a new oC version name at places where used.

Backport to 10.9 and 10.8

Note for 10.8, this needs a manual fix to exclude 10.9 stuff (eg. at the release notes ect)

Note, there will be build warnings (but not errors) which will go away after backporting. Therefore backporting must be done immediately after merging!


EParzefall (approved):
LGTM

Issue #4427, opened 2021-11-30
Names of Secure view document not consistent over the releases.


voroyam:
## WHAT Needs to be Documented?
Secure View should be the name of the document describing secure view or Collabora / Secure view but it should be consistent. 

## WHERE Does This Need To Be Documented (Link)?
https://doc.owncloud.com/server/10.7/admin_manual/enterprise/collaboration/collabora_secure_view.html
https://doc.owncloud.com/server/10.8/admin_manual/enterprise/collaboration/collabora_secure_view.html
https://doc.owncloud.com/server/next/admin_manual/enterprise/collaboration/collabora_secure_view.html

## WHY Should This Change Be Made?
in 10.7 it's Secure View
in 10.8 and master it's secure view

## (Optional) What Type Of Content Change Is This? 
- [ ] New Content Addition
- [ ] Old Content Deprecation
- [ ] Existing Content Simplification
- [ ] Bug Fix to Existing Content

## (Optional) Which Manual Does This Relate To?
- [x] Admin Manual
- [ ] Developer Manual
- [ ] User Manual
- [ ] Android
- [ ] iOS
- [ ] Branded Clients
- [ ] Desktop Client
- [ ] Other

PR #4571, opened 2021-12-31
Lower case in secure view

EParzefall:
This fixes issue #4427 and more. Secure view is now only capitalized in headlines since it's a feature not a product.
Backports to 10.9 and 10.8.


mmattel:
I always thought that Secure View is a brand name we have and therefore written with capital letters at the beginning,
but I may be wrong with that ...

mmattel (approved):

phil-davis (approved):

voroyam:
> I always thought that Secure View is a brand name we have and therefore written with capital letters at the beginning, but I may be wrong with that ...

I thought so as well. It's the name of a feature => names are written in capital letters.

The problem I had with the document name (title) was "fixed" with 10.9 because 10.7 was removed from the list. in 10.7 the title of the document was something different.

PR #4580, opened 2022-01-03
fix branded clients access link

mmattel:
Missed to fix the access ink to the branded client documentation.
pdf works (was fixed before).

Backport to 10.9 and 10.8


EParzefall (approved):
LGTM

PR #4681, opened 2022-01-26
Update kopano-setup.adoc

jnweiger:
openidconfig is json syntax.
Please double check the keyword.

Cosmetic, no backports required.


jnweiger:
Ah, nonsense, that is indeed php. I bet it was somewhere. Cannot find it anymore...

PR #4711, opened 2022-03-05
Add user guides

mmattel:
This PR adds the `docs-webui` repo which contains the `classic server ui` and the `owncloud web ui`. Note that atm the owncloud web structure is present without content in the repo and therefore not activated. Activation of owncloud web can be done in docs-webui easily and no changes are needed here in docs to do so.

Post merging this, the deactivation and removal of the user documentation in the docs-server repo needs to be made including a check of possible broken links.

With adding the new docs-webui, a new navigation was necessary which will also enable easily the integration of oCIS.

Note that oCIS and webui release/release-note links are named in the navigation but commented out. If this gets activated, we need to uncomment and add the files accordingly.

Note that we have to temporary disable pdf access for 10.9 and 10.8 which we can reenable with Antora 3 and a changed pdf creation process. An upgrade to Antora 3 should be made before the next server version is released.


EParzefall (approved):
LGTM

PR #4749, opened 2022-06-02
Prepare for ocis beta 3

mmattel:
Updates the attribute for ocis beta 3

Merging when we get the ok from @micbar

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -93,7 +93,7 @@ asciidoc:
 #   ocis
     latest-ocis-version: 'next'
     previous-ocis-version: 'next'
-    ocis-version: '2.0.0-beta3'
+    ocis-version: '2.0.0-beta.3'
     ocis-downloadpage-url: 'https://download.owncloud.com/ocis/ocis/testing/'
     ocis-ext-includes: 'https://raw.githubusercontent.com/owncloud/ocis/docs/extensions/_includes/'
 #   webui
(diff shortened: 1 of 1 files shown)


site.yml
```diff
@@ -93,7 +93,7 @@ asciidoc:
 #   ocis
     latest-ocis-version: 'next'
     previous-ocis-version: 'next'
-    ocis-version: '2.0.0-beta2'
+    ocis-version: '2.0.0-beta3'
```
phil-davis, line 96: Note: there has recently been some discussion about version strings and semver for the part of the string after 2.0.0

The semver standard would designate this `2.0.0-beta.3`

@micbar or anyone? Comments please. CC: @xoxys

mmattel: I have seen this too, but as long there is no official statement/change, I stick to the old way of writing.

Note that a naming change impacts how folders are named on `download.owncloud.com` - we are using this attribute to create the correct url, see the ocis quick guide https://doc.owncloud.com/ocis/next/quickguide/quickguide.html#setup

kulmann: Decision was made that the next ocis beta will be named `2.0.0-beta.3`. While semver doesn't strictly require the `.`, it seems to be common practice, which we'll follow as well from now on.

kulmann, line 96:
```suggestion
    ocis-version: '2.0.0-beta.3'
```

mmattel: From chat: folders are generated automatically by the build process according the version set.

phil-davis (approved):

PR #4755, opened 2022-06-24
Removed php from occ calls

d7oc:
The occ command doesn't need to be called with php as it contains a shebang nowadays. So removing php in the corresponding calls.


phil-davis:
```
phil@phil-Inspiron-5468:~/git/owncloud/core$ occ app:list --enabled

Command 'occ' not found, but there are 21 similar ones.
```

On Ubuntu 20.04 in my git clone development repo.

Is there some magic that I need to do?

d7oc:
Is occ in your PATH? If not you need to call `/var/www/owncloud/occ`.

mmattel (approved):

phil-davis:
True - in my case just in an ordinary directory where I cloned the git repo, I need to put `./` in front:
```
./occ app:list --enabled
```
That works.

PR #4792, opened 2022-11-26
Rename Classic Server (no reference changes)

mmattel:
Renaming the headlines `Classic Server` to `ownCloud Server` on request of @dragotin.

Note no changes to references are made, optical change only.

Only merge on approval of @dragotin


EParzefall (approved):

jnweiger:
That makes naming less logical and probably adds some confusion.
We now have **ownCloud** server which is not part of **ownCloud** Infinite Scale.
And we have **Infinite Scale** server, which is part of ownCloud **Infinite Scale**.

PR #4871, opened 2023-08-18
IE11 deprecation note

pako81:
Added planned IE11 deprecation for next core version in the 10.13 Release Notes.

### DIFF
--- a/modules/ROOT/pages/server_release_notes.adoc
+++ b/modules/ROOT/pages/server_release_notes.adoc
@@ -46,6 +46,10 @@
 
 Dear ownCloud administrator, find below the changes and known issues in ownCloud Server 10.13.0 that need your attention. You can also read the {oc-changelog-url}[full ownCloud Server changelog] for further details on what has changed.
 
+=== Internet Explorer 11 Deprecation Note
+
+Internet Explorer 11 support will be dropped in the next server release.
+
 === Update Symfony from 4.4.* to 5.4.*
 
 Symfony framework has been updated to LTS major version 5.4. This also affected a number of apps which required a new release and have been now included in the complete bundle.


mmattel:
Question just to be precise:
* It is already dropped because of dependency upgrades in 10.13
* or it will be dropped with the next release AFTER 10.13 because of upcoming dependency upgrades?

phil-davis (approved):
LGTM. This means that we still support IE11 for running the classic web UI with an ownCloud 10.13.0 server.

But 10.14.0 will not support IE11 (some things might still work, but no guarantee)

pako81:
We have seen that the JS dependencies updates done in 10.13 already broke things in core in combination with IE11 (and possible in other apps where they have been applied as well?). For this reason, as of 10.14.0/11 IE11 support will be discontinued.

mmattel:
For me this sounds quite strange:
`We have seen that the JS dependencies updates done in 10.13 already broke things in core in combination with IE11 (and possible in other apps where they have been applied as well?)`
Which means we support a browser where stuff is already broken for this browser.

Imho, an additional release note would be good that tells that already some IE11 things might not be fully functional due to necessary dependeny upgrades made in 10.13

phil-davis:
https://github.com/owncloud/core/pull/40934 fixed an IE11 problem in core.
files_mediaviewer is a "newer" app and has always been known to not be supported on IE11.

IMO there are no more known issues in core 10.13 with IE11. So we can continue to support it for this release.
Is anyone else aware of another IE11 problem?

hodyroff:
The notice is forward looking, always was. Yes, we fixed an issue for IE11 in 10.13 now, therefor we decided to do the depreciation warning for the next version. If we find 10.14 works, we still have an out and can continue to support.

mmattel (approved):

PR #4891, opened 2023-10-09
Add release notes for ocis 4.0.2

mmattel:
References: https://github.com/owncloud/docs-ocis/pull/631 (Prepare for ocis 4.0.2)

Add release notes for recently released ocis 4.0.2

The source of the issues fixed is: https://github.com/owncloud/ocis/issues/7370 with the referenced PR's.

### DIFF
--- a/modules/ROOT/pages/ocis_release_notes.adoc
+++ b/modules/ROOT/pages/ocis_release_notes.adoc
@@ -13,6 +13,30 @@
 
 toc::[]
 
+== Infinite Scale 4.0.2
+
+[discrete]
+=== General
+
+This is a patch release only, please update as soon as possible.
+
+[discrete]
+=== Issues Fixed
+
+* Fixed a problem where the states of received shares were reset to PENDING in the `ocis migrate rebuild-jsoncs3-indexes` command. https://github.com/owncloud/ocis/pull/7336[#7336]
+
+* Fixed an issue that allowed to create two schools with the same school number. https://github.com/owncloud/ocis/pull/7351[#7351]
+
+* Actually pass `PROXY_OIDC_SKIP_USER_INFO` option to oidc client middleware. https://github.com/owncloud/ocis/pull/7220[#7220]
+
+* Reintroducing the `USERS_LDAP_USER_SCHEMA_ID` variable which was accidently removed from the users service
+with the 4.0.0 release. https://github.com/owncloud/ocis/pull/7321[#7321]
+
+* Always pass adjusted default nats options. https://github.com/cs3org/reva/pull/4214[cs3org/reva#4214]
+
+* In the yaml example for Keycloak, set `GRAPH_USERNAME_MATCH` to `none`, to accept any username that is
+also valid for keycloak. https://github.com/owncloud/ocis/pull/7230[#7230]
+
 == Infinite Scale 4.0.1
 
 [discrete]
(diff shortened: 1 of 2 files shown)


EParzefall (approved):

modules/ROOT/pages/ocis_release_notes.adoc
```diff
@@ -13,6 +13,30 @@
...
+* Fixed a problem where the states of received shares were reset to PENDING in the `ocis migrate rebuild-jsoncs3-indexes` command. https://github.com/owncloud/ocis/pull/7336[#7336]
+
+* Fixed an issue that allowed to create two schools with the same school number. https://github.com/owncloud/ocis/pull/7351[#7351]
+
+* Actually pass `PROXY_OIDC_SKIP_USER_INFO` option to oidc client middleware. https://github.com/owncloud/ocis/pull/7220[#7220]
+
+* Reintroducing the `USERS_LDAP_USER_SCHEMA_ID` variable which was accidently removed from the users service
+with the 4.0.0 release. https://github.com/owncloud/ocis/pull/7321[#7321]
+
+* Always pass adjusted default nats options. https://github.com/cs3org/reva/pull/4214[cs3org/reva#4214]
```
micbar, line 35: This is a complete list. Not needed IMO

micbar: We always mention one important one and link to the full changelog.

mmattel: Disagree on the "not needed" statement.
* We provide this info for server patch releases
* We provide this info for ocis patch releases (see 4.0.1)
* All external code we use eg for building docs provide a changelog for patch releases.
And as far I can tell, this is true for ocis, core and _all_ other repos.

Therfore we stick to this well known and accepted behaviour and I see no reason we should break this.

mmattel: Ok, this sounds a bit different now, I get the point.

On a list of 62 changes as we had in 4.0.0, we picked the 12 most important ones and referenced the rest to the full blown changelog.
On 10.13.1 we described 4 out of 9.
On 10.13.2 we will describe 7out of 7

For 4.0.2 we describe 6 changes out of 6 which sounds totally fine and inline from the quantity pov. If you would like to have that different, I need to know which ones you would like to drop and where to reference for the full list.

For server we have a good changelog including patch releases at: https://owncloud.com/changelog/server/ but for ocis, we are lacking the patch releases part in: https://owncloud.com/changelog/infinite-scale/  --> which is not imho professional and should be fixed. But we have a changelog in https://github.com/owncloud/ocis/releases/tag/v4.0.2 ...

I have not searched for it, but I found it and as you can believe, was not happy about 😵 🤪

micbar (approved):

modules/ROOT/pages/ocis_release_notes.adoc
```diff
@@ -13,6 +13,30 @@
...
+=== General
+
+This is a patch release only, please update as soon as possible.
+
+[discrete]
+=== Issues Fixed
+
+* Fixed a problem where the states of received shares were reset to PENDING in the `ocis migrate rebuild-jsoncs3-indexes` command. https://github.com/owncloud/ocis/pull/7336[#7336]
+
+* Fixed an issue that allowed to create two schools with the same school number. https://github.com/owncloud/ocis/pull/7351[#7351]
```
phil-davis, line 28:
```suggestion
* Fixed an issue that allowed two schools to be created with the same school number. https://github.com/owncloud/ocis/pull/7351[#7351]
```

phil-davis (approved):

PR #4897, opened 2023-10-17
Add 41014 as known issue to 10.13.2

jnweiger:
Do we have a proper issue for that?


pako81:
Should we specifically mention federated shares? AFAICT we were not able to nail the performance issues down to federation only.

pako81:
> Do we have a proper issue for that?

This one only: https://github.com/owncloud/enterprise/issues/6098

pako81 (approved):

PR #4921, opened 2024-01-02
Remove modules directory

mmattel:
The modules directory containing main page content is now handled in the [docs-main](https://github.com/owncloud/docs-main) repo. This makes it possible to do locally a quick build of the man page which only takes a view seconds compared to 7 minutes before. The main corresponding PR was https://github.com/owncloud/docs-main/pull/1.

Having that, there is now only the build environment necessary for all repos containing content.

**IMPORTANT:**
It also means, that - for the time being - we need to update site.yml on version changes on both the **docs** (content:sources + attributes) repo and **docs-main** (attributes only) repo.

A local build runs fine.


EParzefall (approved):
LGTM

PR #4922, opened 2024-01-02
Remove some scripts that are here no longer necessary

mmattel:
References: #4921 (Remove modules directory)

These scripts are now no longer necessary in this repo as it contains no doc content (no modules directory).


kobergj (approved):
:heart:

PR #4977, opened 2024-07-11
Enable ocis rolling releases

mmattel:
This PR enables / prepares for ocis rolling releases.

The tab text in a service in the ocis admin docs for the `next` release named `master` is then printed
as `master + Rolling 6.1.0`. A PR in the docs-ocis repo will follow asap.

This does not affect branched releases. They stay with their defined number in the tab text.

Additionally I found an improvement in the site structure docs which I added to the fix.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -113,9 +113,11 @@ asciidoc:
     previous-ocis-version: '4.0'
     # These versions are just for printing like in docs-main release info, but not used in docs-ocis.
     # Versions in the ocis docs need to be defined in the branch specific docs-ocis/antora.yaml file.
-    # To do so, change the values in docs-ocis/antora.yml like compose_tab_1_tab_text.
-    ocis-actual-version: '5.0.4'
+    # To do so, change the values in docs-ocis/antora.yml like service_tab_1_tab_text.
+    ocis-actual-version: '5.0.5'
     ocis-former-version: '4.0.7'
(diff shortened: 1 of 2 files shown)


site.yml
```diff
@@ -113,9 +113,11 @@ asciidoc:
     previous-ocis-version: '4.0'
     # These versions are just for printing like in docs-main release info, but not used in docs-ocis.
     # Versions in the ocis docs need to be defined in the branch specific docs-ocis/antora.yaml file.
-    # To do so, change the values in docs-ocis/antora.yml like compose_tab_1_tab_text.
-    ocis-actual-version: '5.0.4'
+    # To do so, change the values in docs-ocis/antora.yml like service_tab_1_tab_text.
+    ocis-actual-version: '5.0.5'
     ocis-former-version: '4.0.7'
```
dragonchaser, line 118: is this correct? former-version should be 5.0.4 or variable should IMHO be called ocis-former-production-version or ocis-former-major-version.

dragonchaser (changes requested):

site.yml
mmattel: yes it is all correct.

nitpicking about the name 😅 
no, it is about major releases and last former major was 4

next prod will be lets say 7.0.0, we then have actual 7.0.0 and former 5.x.y
note you can look into the files changed and you see in the comments above what it is used for... 😄

dragonchaser (approved):

PR #4996, opened 2024-12-12
Update gitignore (add .idea)

mmattel:


phil-davis (approved):
Note: users can configure git on their local computer to ignore whatever temporary/settings files their development environment creates, which depends on which development system/IDE they use. That avoids having to list `.vscode` `.idea` and... for every development environment that exists, in the `.gitignore` of every repo.

But we already have `.vscode` listed, so may as well have `.idea`

PR #5001, opened 2024-12-29
Create a global attributes file

mmattel:
This PR is a preperation to handle a global attributes file for all repos.

With a subsequent PR, the relevant code and other changes will be added.
This will also be done in other doc repos!

Currently, this change only adds a file with attributes copied from site.yml
That file is the basis for all other repos and the future single point of maintenance.
With the next PR, the attributes will get removed from site.yml and code will be added that loads the attributes dynamically.

### DIFF
--- a/global-attributes.yml
+++ b/global-attributes.yml
@@ -1,0 +1,76 @@
+#   branch-specific-variables: 'see antora.yml'
+#   build relevant
+    # asciidoctor prefix for section IDs
+    idprefix: ''
+    idseparator: '-'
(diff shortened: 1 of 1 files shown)


kobergj (approved):

global-attributes.yml
```diff
@@ -0,0 +1,76 @@
+#   branch-specific-variables: 'see antora.yml'
+#   build relevant
```
mmattel, line 2:
```suggestion
#   global attributes loaded via extension
```

PR #5082, opened 2026-04-13
feat: add /latest/ URL redirect to current stable version

DeepDiver1975:
## Summary

- Adds `urls.latest_version_segment: latest` and `urls.latest_version_segment_strategy: redirect:from` to `site.yml`
- Antora will now generate static HTML redirect files at `/component/latest/**` pointing to the current stable version (e.g. `/desktop/latest/` → `/desktop/6.0/`, `/server/latest/` → `/server/10.16/`, etc.)
- Works for all components and all sub-pages, not just the index
- Uses the already-configured `--redirect-facility static` flag — no build script changes needed

## Notes

- `redirect:from` keeps the versioned URL as canonical; `/latest/` acts as an alias
- Prerequisite: `master` branches in component repos should have `prerelease: true` in their `antora.yml` so Antora correctly identifies the highest stable version as "latest" — please verify before merging

## Test plan

- [ ] Build the site locally with `npm run antora` and confirm `public/desktop/latest/` redirect files are generated
- [ ] Verify `public/desktop/latest/index.html` contains a redirect to `/desktop/6.0/`
- [ ] Check a sub-page, e.g. `public/desktop/latest/faq/` also redirects correctly
- [ ] Confirm component repos' `master` branches have `prerelease: true` in `antora.yml`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,10 @@
 site-dev.yml
 yarn-error.log
-cache/
 node_modules/
 public/
 pdf_web/
 tmp/
+.cache/
 .DS_Store
 .vscode/
 .idea/
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,13 +1,9 @@
 {
-  "name": "owncloud-docs",
-  "version": "4.0.0",
+  "name": "docs",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
-      "name": "owncloud-docs",
-      "version": "4.0.0",
-      "license": "AGPL-3.0",
       "dependencies": {
         "@elastic/elasticsearch": "^7.17.14",
         "antora": "^3.1.14",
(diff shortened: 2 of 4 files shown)


mmattel:
Claude is more guessing than knowing and having a tested PR is anyways out of scope when AI creates PRs...

With commit: [make it working and add related updates](https://github.com/owncloud/docs/pull/5082/commits/3aa3fc2dd7007e6615a6ac4367e43894423ea205), the following fixes have been incorporated and fully tested:

* `latest_version_segment` and `latest_version_segment_strategy` is now proven to work **as added**
Additional and required comments have been incorporated, which help to understand their usage with GH Pages
* The change is now fully tested and does the following **for multi versioned components**:
   * The highest version will get `latest` in the URL instead of the version number
   This has the positive side effect, that if one references via link, it will always be resolvable.
   * The real version number is still shown on the page and via the drop-down selector
   * Additionally, one can use the version number in the URL instead
   * Existing hard coded links from the outer world that used `latest` in the URL will be resolved again
* Relevant build keys have been moved from `package.json` to `site.yml` settings
This makes the parameters much easier to handle and if CLI overwriting is necessary better readble
* The directory of the Antora cache has been updated
* `package.json` has been cleaned up and updated

**IMPORTANT I**
To make this work as described, all content repos that have versions MUST have `prerelease: true` in the `antora.yml` file that is not a stable version !

**IMPORTANT II**
As fallout of this change, an additional PR must be created that fixes sitemap creation.

**IMPORTANT III**
All versioned content repos need an update how to create a version branch. This will be done independently so as not to block the build task.

mmattel (approved):

PR #5119, opened 2026-07-02
fix: remove the asciidoctor dependency

mmattel:
This PR removes the asciidoctor dependency.

This dep is a legacy dating back long time ago when a special mechanism was used to create PDFs.
This mechanism has been removed years ago but we "forgot" to remove the dependency.

Antora can now create PDFs in a new and very professional way using [Antora Assembler](https://docs.antora.org/assembler/latest/) where all dependencies are part of the assembler setup.

Note that the UI is independent and needs the assembler dependency with the version fixed to 3.x which is for building the test page when developing the UI.

The removal has been tested locally running a full docs build and no issues occurred.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,6 @@
     "": {
       "dependencies": {
         "antora": "^3.1.15",
-        "asciidoctor": "^3.0.4",
         "asciidoctor-kroki": "^0.18.1",
         "js-yaml": "^5.2.0"
       },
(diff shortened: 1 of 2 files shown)


DeepDiver1975 (approved):
Verified: removes orphaned `asciidoctor` npm dep (+lockfile prune); no remaining consumer (extensions use Antora-bundled Asciidoctor via the `Opal`/`Asciidoctor` globals, not the standalone package; `asciidoctor-kroki` retained). Full Antora build CI green.

PR #5137, opened 2026-07-24
fix(redirects): honor legacy go.php doc links on the static site

DeepDiver1975:
## What

ownCloud core builds documentation deep links via `OC_Defaults::buildDocLinkToKey()` as `https://doc.owncloud.com/server/<version>/go.php?to=<key>`. On the old Sphinx site `go.php` was a server-side PHP script that 302-redirected each key to the real page. The docs now build as a **static Antora site on GitHub Pages**, which runs no PHP and ignores the query string, so **every `go.php?to=` URL 404s**. Every already-deployed ownCloud server keeps emitting these links, so the fix must live on the **site**.

This ports [`owncloud-docs-monorepo` PR #35](https://github.com/DeepDiver1975/owncloud-docs-monorepo/pull/35) to the production `owncloud/docs` playbook. It ships a client-side redirector as Antora **supplemental files** layered on the remote `docs-ui` bundle — nothing in `owncloud/docs-ui` changes.

## How

- **`supplemental/js/go-redirect.js`** — maps each `go.php` key to its page under the version root, remaps unpublished version segments (the concrete current stable like `10.16`, or old releases) to `/server/latest/`, and falls back to the version-root entrypoint for unknown/missing keys. It runs in `<head>` on every page and is a no-op unless the path ends in `go.php`, so it redirects **before the 404 body paints** (no visible flash).
- **`supplemental/partials/head-scripts.hbs`** — overrides the bundle partial to load the redirector. Overriding a stock partial wholesale drops its contents, so this reproduces the bundle's Matomo block and the `uiRootPath` global, then adds the redirector `<script>`.
- **`site.yml`** — enables the supplemental overlay (`supplemental_files: supplemental`).
- **`test/go-redirect.test.js` + `npm test` + CI step** — unit-tests the mapping logic and, on the built site, guards against version drift (`PUBLISHED_VERSIONS` must equal `public/server/*`) and page moves (every mapped target must exist under `public/server/latest`).

## Adaptations from PR #35

- Domain-root paths (`/server/latest/…`), not the monorepo's repo path prefix — the redirector logic is prefix-agnostic and needed no change.
- `PUBLISHED_VERSIONS = ['10.15', 'next', 'latest']` — the segments this playbook actually emits (`10.16` is served at `/server/latest/` via `latest_version_segment_strategy: replace`; `master` at `/server/next/`).
- Test infra added (this repo had none): `node --test` wired into CI after the build step so the guards run against `public/`.

## Testing

- `npm test` — 15 tests pass (the two build-dependent guards `t.skip` when `public/` is absent).
- `npm run antora && npm test` — full build; guards confirm `public/server/*` = `10.15`, `latest`, `next` and every mapped target exists under `public/server/latest`.
- Local smoke: `/server/latest/go.php?to=admin-sharing` → `…/admin_manual/configuration/files/file_sharing_configuration.html`; unknown key → `…/index.html`; `/server/10.16/go.php?to=user-webdav` → `/server/latest/classic_ui/files/access_webdav.html`.

## Out of scope

- The `owncloud/core` `buildDocLinkToKey()` change (Part B).
- Any change to `owncloud/docs-ui`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,9 @@ jobs:
           rm -f pagefind.log
           echo 'doc.owncloud.com' > public/CNAME
 
+      - name: Test (go.php redirector map, incl. target existence)
+        run: npm test
+
       - name: Save cache
         if: github.event_name == 'push'
         uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "antora-dev-local": "antora --stacktrace --url http://localhost:8080 site-dev.yml",
     "antora-dev-bundle": "antora --stacktrace --ui-bundle-url ../docs-ui/build/ui-bundle.zip --url http://localhost:8080 site-dev.yml",
     "serve": "http-server public/ -d -i",
-    "pagefind": "pagefind --site public"
+    "pagefind": "pagefind --site public",
+    "test": "node --test test/*.test.js"
   },
   "dependencies": {
     "antora": "^3.1.15",
--- a/site.yml
+++ b/site.yml
@@ -63,7 +63,7 @@ output:
   dir: public
 
 ui:
-  # supplemental_files: supplemental
+  supplemental_files: supplemental
   output_dir: assets
   bundle:
     snapshot: true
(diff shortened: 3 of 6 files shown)


phil-davis (approved):

phil-davis:
Fixed https://github.com/owncloud/firstrunwizard/issues/202

